PR #12, opened 2014-01-23
Early loop detector: dependency loop remover from cache

xaionaro:
Related to: https://bugs.gentoo.org/show_bug.cgi?id=391945

This detector just calculates loops, solves weak loops (with at least one use or after dependency), and print information about all found loops (including unsolvable).

Solving is just breaking the loop with removing one or few use/after dependencies.

This detector cannot solve loops fully based on "need" dependencies. To prevent hang up in this case, it's need to use the later loop detector (https://github.com/OpenRC/openrc/pull/13) or solve it manually (sysadmin will be notified about any loop).

Presentation in PDF: https://github.com/xaionaro/documentation/raw/master/openrc/earlyloopdetector/early-loop-detection.pdf


williamh:
@xaionaro:
Can you meet with myself or @dwfreed sometime on irc to chat again about this code? I'm not sure where we left it. We both agree that there is a problem we need to solve, so let's get together again and chat about it. :-)

heroxbd:
@xaionaro, Hey Dimitry, long time no see.  MIssed you a lot.  Glad to tell that we have OpenRC 0.18.3 in Debian.  Your updates has resolved all the build issues on ppc64el and kFreeBSD.

@williamh, William, sorry for being idle for long.  The point of breaking a loop is not only for Debian, but to make OpenRC more robust against configuration mistakes.  Such a mistake as dependency loop can be fetal (unable to boot).  Dimitry has made the point that the loop solver breaks the weakest dependence in the loop, which makes sense.

The only concern for me is the code complexity, which I think, yes, it should be addressed cooperatively.

xaionaro:
Hello. Sorry that I still didn't fix the coding style of the patch as @williamh requested long time ago. I just have no time. It's quite bad years in my university right now (partly due to geopolitical problems), so I have to work a lot and more. :(

stalkerg:
I suppose we should do something with this pull request.

williamh:
I'm against trying to remove loops behind the sceens; we have no way of knowing for sure what the relationships of the services should be. I would rather just warn about them and allow users to fix them.

stalkerg:
> I would rather just warn about them and allow users to fix them.

For me it's ok but I worry about "rc_parallel=no" by default. It's not normal for modern init system. 
OpenRC must push users to write strict and right rules.

xaionaro:
I can add an option to not break loops automatically, today if required.

On the on hand this patched was written for Debian and it works there well enough, so as for me mission accomplished. On the another hand, I'd like to be able to say people that OpenRC can safely boot in parallel mode (and do not hang), so I'd like to see this patched be merged to the upstream code.

stalkerg:
> I can add an option to not break loops automatically, today if required.

At least it's good functional. But I don't know what thinking @williamh about it. 

>we have no way of knowing for sure what the relationships of the services should be.

If init system don't know about relationships then who knows? My main goal enables parallel by default. 
What should we do for that?

williamh:
@xaionaro I don't want to break loops at all, so please just have this patch print error messages if loops are detected; I think loops indicate bugs which should be fixed by the init script authors. Also, please rebase the patch on current master. I'm definitely interested in this, so I'll take a look once it is updated.

@stalkerg What I meant above is, if we get into a deadlock situation, there is not really a smart way to resolve it automatically. We don't really know what services do, so we can't really decide which service to start first to break the deadlock. I think it is a better choice to warn that there is a loop when we are building the deptree but let the script author take care of it.

stalkerg:
Maybe it's destiny but I have catch deadlock with parallel=on after upgrade to 0.24.

williamh:
I am uncomfortable with this pr because we should warn admins about loops and make them resolve the loops themselves rather than randomly drop dependencies. If you want to do that, feel free to update this and re-open it. Also, please avoid breaking the API. Thanks much.

PR #46, opened 2015-03-29
Fix rc_verbose processing

williamh:
We were processing the rc_verbose setting before we sourced the
configuration file where it was set; this fixes the issue.


Whissi:
For the logs: This PR restores the fix from 6965d85f which was accidental reverted by commit de60ffeebe. So this is a regression.

PR #49, opened 2015-04-01
Fix tmpfiles processing

williamh:
Tmpfiles.d processing had /run overriding /usr/lib and /etc, but this is
not correct. The correct order, from lowest to highest, for tmpfiles
processing is:

* /usr/lib/tmpfiles.d/*.conf
* /run/tmpfiles.d/*.conf
* /etc/tmpfiles.d

This means /run/tmpfiles.d/*.conf can override /etc/tmpfiles.d/*.conf,
but /etc/tmpfiles.d/*.conf can override both of them.

### DIFF
--- a/sh/tmpfiles.sh.in
+++ b/sh/tmpfiles.sh.in
@@ -245,7 +245,7 @@ PREFIX=
 FILE=
 fragments=
 # XXX: The harcoding of /usr/lib/ is an explicit choice by upstream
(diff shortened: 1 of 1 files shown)


williamh:
@robbat2:
Since you wrote the tmpfiles code, can you take a look at this before we merge it and confirm that it is correct?

Thanks,

William

Issue #54, opened 2015-05-14
services in 'boot' runlevel are not pulled in when switching to empty runlevel


ncopa:
The services in 'boot' and 'sysinit' runlevels should be pulled in when switching to any runlevel (except shutdown). This only happens if there are services to start in the runlevel.

When booting the alpine iso we have an empty 'default' runlevel and unlike Gentoo, we don't explicitly run the 'boot' level from /etc/inittab. This has worked fine til operc-0.13 and later.

Screenshot:

```
(none):~# rc-status --runlevel                                                  
default
(none):~# rc-update                                                             
             bootmisc | boot                                                    
                devfs |                       sysinit                           
                dmesg |                       sysinit                           
             hostname | boot                                                    
              hwclock | boot                                                    
            hwdrivers |                       sysinit                           
            killprocs |              shutdown                                   
                 mdev |                       sysinit                           
              modloop |                       sysinit                           
              modules | boot                                                    
             mount-ro |              shutdown                                   
            savecache |              shutdown                                   
               sysctl | boot                                                    
               syslog | boot
(none):~# rc-status                                                             
Runlevel: default                                                               
Dynamic Runlevel: hotplugged                                                    
Dynamic Runlevel: needed                                                        
 sysfs                                                             [  started  ]
 mdev-mount                                                        [  started  ]
Dynamic Runlevel: manual
(none):~# openrc default                                                        
(none):~# rc-update -s add boot default                                         
 * rc-update: cannot stack the sysinit runlevel
(none):~# openrc boot                                                           
 * Loading modules ...                                                    [ ok ]
 * Setting system clock using the hardware clock [UTC] ...                [ ok ]
 * Checking local filesystems  ...                                        [ ok ]
 * Remounting filesystems ...                                             [ ok ]
 * Mounting local filesystems ...                                         [ ok ]
 * Configuring kernel parameters ...                                      [ ok ]
 * Migrating /var/lock to /run/lock ...                                   [ ok ]
 * Migrating /var/run to /run ...                                         [ ok ]
 * Creating user login records ...                                        [ ok ]
 * Wiping /tmp directory ...                                              [ ok ]
 * Setting hostname ...                                                   [ ok ]
 * Starting busybox klogd ...                                             [ ok ]
 * Starting busybox syslog ...                                            [ ok ]
localhost:~# rc-status --runlevel                                               
boot
localhost:~# openrc default                                                     
localhost:~# 
```

ncopa:
[...]

PR #55, opened 2015-05-14
rc: pull in sysinit and boot as stacked levels when needed

ncopa:
We need start services from sysinit and boot runlevel, even if the new
runlevel is empty.

This fixes problem introduced with commit 7716bf31 (Fix stacked runlevel
support), at which the start_services list are no longer used to start
the services.

This also make sure that all services in sysinit and boot runlevels are
started before switching to next. This was not guaranteed when switching
to a non-empty runlevel.

Fixes issue #54.

### DIFF
--- a/src/rc/rc.c
+++ b/src/rc/rc.c
@@ -757,6 +757,7 @@ main(int argc, char **argv)
 	static RC_STRINGLIST *types_n;
 	static RC_STRINGLIST *types_nua;
 	static RC_DEPTREE *deptree;
+	RC_STRINGLIST *runlevel_chain;
 	RC_STRINGLIST *deporder = NULL;
 	RC_STRINGLIST *tmplist;
 	RC_STRING *service;
@@ -1016,6 +1017,7 @@ main(int argc, char **argv)
 	hotplugged_services = rc_services_in_state(RC_SERVICE_HOTPLUGGED);
 	start_services = rc_services_in_runlevel_stacked(newlevel ?
 	    newlevel : runlevel);
+	runlevel_chain = rc_runlevel_stacks(newlevel ? newlevel : runlevel);
 	if (strcmp(newlevel ? newlevel : runlevel, RC_LEVEL_SHUTDOWN) != 0 &&
 	    strcmp(newlevel ? newlevel : runlevel, RC_LEVEL_SYSINIT) != 0)
 	{
@@ -1033,6 +1035,7 @@ main(int argc, char **argv)
 				tmplist = rc_services_in_runlevel(bootlevel);
 				TAILQ_CONCAT(start_services, tmplist, entries);
 				free(tmplist);
+				rc_stringlist_add(runlevel_chain, bootlevel);
 			}
 			if (hotplugged_services) {
 				TAILQ_FOREACH(service, hotplugged_services,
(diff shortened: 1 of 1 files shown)


dwfreed:
Discussion from the IRC channel:
```
2015-05-15 11:39:48 < dwfreed> I'm not sure we should be automatically pulling in sysinit and boot runlevels
2015-05-15 11:40:13 < dwfreed> in gentoo, they're started by inittab before entering the default runlevel
2015-05-15 11:40:31 < dwfreed> and I think it's better to be explicit than implicit
2015-05-15 11:41:10 < dwfreed> but I'll leave that to WilliamH to decide
2015-05-15 11:42:04 < dwfreed> at the very least, your changes seem sane enough, and it compiles, and I'm sure you tested it, so if he agrees with your change, it can be merged
```

ncopa:
> 2015-05-15 11:39:48 < dwfreed> I'm not sure we should be automatically pulling in sysinit and boot runlevels

It behaved like this pre openrc-0.13. The mentioned commit 7716bf3 introduced this new behavior. Commit message does not say anything about desired new behavior. Nor does any documentation (that i could found) mention new behavior since 0.13.

This new (unintentional?) behavior did break things for Alpine Linux. So I consider this a regression.

> 2015-05-15 11:40:13 < dwfreed> in gentoo, they're started by inittab before entering the default runlevel

So my proposed patch will not hurt Gentoo users.

> 2015-05-15 11:40:31 < dwfreed> and I think it's better to be explicit than implicit
> 2015-05-15 11:41:10 < dwfreed> but I'll leave that to WilliamH to decide

If this is the desired behavior, then please document it so I can point Alpine users who upgrade to it if things break.

> 2015-05-15 11:42:04 < dwfreed> at the very least, your changes seem sane enough, and it compiles, and I'm sure you tested it, so if he agrees with your change, it can be merged

I have tested it and made sure the order is correct (so services in sysinit gets started before boot).

I would be very happy if it got applied, because I am afraid that this issue will case non-obvious breakages due some of the boot services not starting at boot. For example /tmp no longer getting wiped, and hwclock not being set. Those changes can easily go unnoticed.

Thanks

williamh:
@ncopa:
I'm curious, if you set up your inittab the way the example in support/sysvinit is set up, do you still have issues?

I'm sitting on the fence on this one, so fill me in on what changed for you.

williamh:
@ncopa:
I  just did a test here as follows:
I made an empty runlevel, then used openrc <runlevel> to switch to the new runlevel.
The only services that were stopped during the switch were the ones I have in the default runlevel; the services in sysinit and boot were not affected. Can you confirm this behavior?

williamh:
@nncopa:
I want to appologise for not looking at this. Can you please rebase it on current master so I can take a look now? Thanks much for your understanding.

dwfreed:
Having looked into this further, there is **a lot** of things OpenRC does in sysinit level only when it is explicitly started, such as printing its version, running init.early.sh, running init.sh, handling softlevel from the kernel command line, etc.  Refactoring that code to allow sysinit runlevel to be implicitly started and work correctly is a lot of work, and very little benefit, and not something I'd like to support.

williamh:
After speaking with @dwfreed, in a nutshell, we are talking about the do_sysinit function along with running /lib*/rc/init.*. This setup is only done when "openrc sysinit" is run. Your p/r does not make sure those things are run, and it would be difficult to refactor the code to do this. If you can do this refactor, please open another pull request and we'll take a look, but the best bet is to use inittab.

PR #61, opened 2015-07-29
Prototype path keyword for depend()

willmmiles:
This adds a prototype 'path' keyword for depend() functions that will look up where the specified path is mounted and emit an appropriate mount.xxx dependency line.

Testing was minimal, but it seemed to work OK on my QNX system.


willmmiles:
Ack, pushed the wrong second commit :(

PR #64, opened 2015-08-03
openrc-run: Fix default behavior when rc_hotplug is defined

dwfreed:
When rc_hotplug is undefined, service_plugable always returns false.

When rc_hotplug is defined, service_plugable previously would return
true unless an entry in rc_hotplug negated the behavior for a specific
service.

This change makes service_plugable return false unless an entry in
rc_hotplug says otherwise, and doesn't break the '!foo' case, which is
supposed to match all services except 'foo'.

Obsoletes #63

X-Gentoo-Bug: 554540
X-Gentoo-Bug-URL: https://bugs.gentoo.org/show_bug.cgi?id=554540

### DIFF
--- a/src/rc/openrc-run.c
+++ b/src/rc/openrc-run.c
@@ -1028,7 +1028,7 @@ static bool
 service_plugable(void)
 {
 	char *list, *p, *token;
-	bool allow = true, truefalse;
+	bool allow = false, truefalse;
 	char *match = rc_conf_value("rc_hotplug");
 
 	if (!match)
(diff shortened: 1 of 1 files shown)


floppym:
> and doesn't break the '!foo' case, which is supposed to match all services except 'foo'.

This part seems stupid to me. !foo should disable foo, not enable everything else. To quore rc.conf:

```
# To disable services, prefix with a !
```

The example below it then gives conflicting information.

```
# Example - rc_hotplug="net.wlan !net.*"
# This allows net.wlan and any service not matching net.* to be plugged.
```

williamh:
@dwfreed:
Is this still a valid p/r? I thought we may have decided that this was a documentation issue.
Let me know.

Thanks,

William

dwfreed:
The documentation matches the implemented behavior now, so I guess not.

PR #65, opened 2015-09-14
Fix for s6-svc handling

colinbooth:
s6 v2.2.0.0 changed the readyness notification flags for s6-svc from -D/-U to -wd/-wD/-wu/-wU in order to support more granular change types. This change updates the s6 helper script with the correct syntax. In addition, it will cause s6-svc to wait until the service is down and (if applicable) any follow-on ./finish script has completed before returning instead of returning after the service has terminated but before ./finish has exited.


dwfreed:
Please squash your commits into one commit

Issue #68, opened 2015-11-03
Add SSD_IONICELEVEL env var


kennethlakin:
SSD_NICELEVEL is nice and all, but start-stop-daemon can *also* control a daemon's ionice level. It might be nice to have an environment variable that controls IO niceness, too.

williamh:
This was fixed in #69.

PR #69, opened 2015-11-03
start-stop-daemon: Add SSD_IONICELEVEL

kennethlakin:
This is the disk io counterpart to SSD_NICELEVEL.

Resolves issue #68 .

Concerns with the code:

* Testing *indicates* that sscanf won't touch pointer arguments that don't have a corresponding match in the string, but the manpage is silent on the matter. OTOH, the code on line 746 does the same thing that I do in my code, so we're not in *more* danger than we were before.

* The ionicec and ioniced manipulation on line 720 is copied from code that handles parsing of the command-line ionice adjustment string, on line 746. If this now-duplicate code needs to be pulled out into a function, I'd be more than happy to make the required changes.

### DIFF
--- a/src/rc/start-stop-daemon.c
+++ b/src/rc/start-stop-daemon.c
@@ -712,6 +712,17 @@ start_stop_daemon(int argc, char **argv)
 		if (sscanf(tmp, "%d", &nicelevel) != 1)
 			eerror("%s: invalid nice level `%s' (SSD_NICELEVEL)",
 			    applet, tmp);
+		if ((tmp = getenv("SSD_IONICELEVEL"))) {
+			int n = sscanf(tmp, "%d:%d", &ionicec, &ioniced);
+			if (n != 1 && n != 2)
+				eerror("%s: invalid ionice level `%s' (SSD_IONICELEVEL)",
+				    applet, tmp);
+			if (ionicec == 0)
+				ioniced = 0;
+			else if (ionicec == 3)
+				ioniced = 7;
+			ionicec <<= 13; /* class shift */
+		}
 
 	/* Get our user name and initial dir */
 	p = getenv("USER");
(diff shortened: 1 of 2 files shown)


williamh:
@kennethlakin:
Please rebase this on current master and update the start-stop-daemon man page as well.

Thanks much,

William

src/rc/start-stop-daemon.c
```diff
@@ -712,6 +712,17 @@ start_stop_daemon(int argc, char **argv)
 		if (sscanf(tmp, "%d", &nicelevel) != 1)
 			eerror("%s: invalid nice level `%s' (SSD_NICELEVEL)",
 			    applet, tmp);
+		if ((tmp = getenv("SSD_IONICELEVEL"))) {
```
dwfreed: This indentation is wrong

src/rc/start-stop-daemon.c
```diff
@@ -1211,7 +1222,8 @@ start_stop_daemon(int argc, char **argv)
 			if ((strncmp(env->value, "RC_", 3) == 0 &&
 				strncmp(env->value, "RC_SERVICE=", 10) != 0 &&
 				strncmp(env->value, "RC_SVCNAME=", 10) != 0) ||
-			    strncmp(env->value, "SSD_NICELEVEL=", 14) == 0)
+				strncmp(env->value, "SSD_NICELEVEL=", 14) == 0 ||
```
dwfreed: This indentation makes things confusing, because it suggests the logic depends on the strncmp of RC_

src/rc/start-stop-daemon.c
kennethlakin: Oh?

I'm fairly sure that I'm following StyleGuide.md here. Notice that the preceding if (712) is indented with two tabs, the eerror statement that follows it (713) is indented with three.

What have I done wrong?

src/rc/start-stop-daemon.c
kennethlakin: The style guide offers no guidance here. Should lines 1225 and 1226 be dedented by one step, making them line up with if clause that they're a part of? Or dedent by one step, and then indent with three spaces, making them line up with the open paren? on line 1222?

dwfreed: It should match the indentation of the old line

src/rc/start-stop-daemon.c
dwfreed: If you expand the context, you'll see that line 712 is indented as far as it is because it's the child of an if statement (line 711).  Since line 715 is not part of that if statement, it should have the same indentation as the if statement.

kennethlakin: Good point. All the more reason to *never* omit optional curly braces around flow control statements.

PR #75, opened 2016-01-18
Move system detection code into rc_sys and use it

dwfreed:
Also some minor fixes to the detection code, and one related cleanup I found.


dwfreed:
This should probably be backported to 0.20, and possibly also 0.19 (0.19 doesn't handle the nothing special case correctly).

PR #82, opened 2016-03-04
librc: sh:  respect alternative INITDIR

jsbronder:
Remove hard-coded directory name for INITDIR, which was partially
respected by the build system but was being ignored at runtime.

Note that I didn't go through and mess with any of the documentation
where 'init.d' was currently referenced as they didn't have any SED rules
similar to the compiled/generated code.  I can do that if necessary.

Tested with ``/etc/openrc.d`` under Linux.

### DIFF
--- a/mk/scripts.mk
+++ b/mk/scripts.mk
@@ -19,7 +19,7 @@ _PKG_SED:=		$(shell ${_PKG_SED_SH})
 _LCL_SED_SH=		if test "${PREFIX}" = "${LOCAL_PREFIX}"; then echo "-e 's:@LOCAL_PREFIX@::g'"; else echo "-e 's:@LOCAL_PREFIX@:${LOCAL_PREFIX}:g'"; fi
 _LCL_SED:=		$(shell ${_LCL_SED_SH})
 
-SED_REPLACE=		-e 's:@SHELL@:${SH}:g' -e 's:@LIB@:${LIBNAME}:g' -e 's:@SYSCONFDIR@:${SYSCONFDIR}:g' -e 's:@LIBEXECDIR@:${LIBEXECDIR}:g' -e 's:@PREFIX@:${PREFIX}:g' -e 's:@BINDIR@:${BINDIR}:g' -e 's:@SBINDIR@:${SBINDIR}:g' ${_PKG_SED} ${_LCL_SED}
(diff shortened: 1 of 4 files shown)


williamh:
@jsbronder
At first look, this seems to break RC_CONFDIR.
If someone defines RC_INITDIR to be /etc1/init.d, RC_CONFDIR should be /etc1/conf.d.
Instead of replacing @SYSCONFDIR@/init.d with @INITDIR@, I recommend replacing init.d with @INITDIR@ and requiring it to be in @SYSCONFDIR@.

rindeal:
If `conf.d` is bound to `init.d`, but `init.d` is not bound to `SYSCONFDIR`, then the config should look like this:
```sh
INITDIR="/what/ever"
CONFDIR="$( dirname "${INITDIR}"/conf.d )"
```

jsbronder:
@williamh So, prior to this patch, someone could set INITDIR to anything that wasn't '/etc/init.d' and end up with a broken system.  How far would you like me to go in fixing this?  I assumed that a small patch addressing one issue at a time would be preferable.

If not, then I'll probably have to make some more significant edits to both INITDIR and CONFDIR to make them both work in every scenario.

dwfreed:
No they couldn't, because prior to this patch, INITDIR wasn't a thing; RC_INITDIR was always SYSCONFDIR/init.d, and thus RC_CONFDIR was always SYSCONFDIR/conf.d.  Whatever you call them, they should always reside in SYSCONFDIR, because that's what SYSCONFDIR means.  If you want to call init.d something else, like bananas.d or whatever, we'd probably accept a patch that makes the dir name configurable (though if you're doing INITDIR, you might as well do CONFDIR while you're at it).

jsbronder:
@dwfreed See https://github.com/OpenRC/openrc/blob/master/mk/sys.mk#L34

Anyways, it sounds like what you guys would like is for me to make sure both initdir and confdir are both inside of SYSCONFDIR but to allow the names 'init.d' and 'conf.d' to be changed.  I'll go to work on that, but please let me know if I've misunderstood.

Thanks all,

dwfreed:
Oh, that's cute; setting that is a great way to completely break the build, because it has no influence on RC_INITDIR, which is what is actually used at runtime to locate your init dir.  And yes, you're understanding correctly :).

PR #83, opened 2016-03-12
tmpfiles: Accept filenames as command line arguments

floppym:
This brings us closer to the behavior of systemd-tmpfiles.

My end goal is to be able to call this via an eclass function in pkg_postinst in ebuilds.

### DIFF
--- a/sh/tmpfiles.sh.in
+++ b/sh/tmpfiles.sh.in
@@ -264,6 +264,32 @@ _Z() {
 BOOT=0 CREATE=0 REMOVE=0 CLEAN=0 VERBOSE=0 DRYRUN=0 error=0 LINENO=0
 EXCLUDE=
 PREFIX=
+
+while [ $# -gt 0 ]; do
+	case $1 in
+		--boot) BOOT=1 ;;
+		--create) CREATE=1 ;;
+		--remove) REMOVE=1 ;;
+		--clean) CLEAN=1 ;; # TODO: Not implemented
+		--verbose) VERBOSE=1 ;;
+		--dryrun|--dry-run) DRYRUN=1 ;;
+		--exclude-prefix=*) EXCLUDE="${EXCLUDE}${1##--exclude-prefix=} " ;;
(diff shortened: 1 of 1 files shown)


dwfreed:
LGTM

floppym:
I pushed a small fix to the first commit.

robbat2:
Conditionally I accept it; there's one change needed:
You must be able to handle arguments that have whitespace in the filename.

floppym:
Heh, show me a reasonable way to do that with portable sh syntax and I'll get right on that.

robbat2:
Doing that with portable sh:
Method 1:
Use \0 as a seperator rather than a space, adjusting IFS when you want to read that variable.

Method 2:
Keep them in arguments, rather than shifting them out; then loop over arguments:
```
# Options pass, replaces $# -gt 0
FILEMODE=0
for opt in "$@" ; do
  if [ "${f:0:2}" != "--" ]; 
    FILEMODE=1
    continue
  fi
  # existing case statement, with $opt instead of $1
  # drop the shift
done
# Arguments pass
if [ $FILEMODE -eq 1 ]; then 
  for f in "$@" ; do
    [ "${f:0:2}" == "--" ] && continue
    process "$f"
  done
  exit # ...
fi
```

floppym:
Thanks. I think using `\0` as a separator might break some shells. I can probably make your second method work.

floppym:
In the case of having no files specified on the command line, we need to generate a sorted union of filenames from `/etc/tmpfiles.d/*.conf`, `/run/tmpfiles.d/*.conf` and `/usr/lib/tmpfiles.d/*.conf`. I cannot see any way to accomplish this in POSIX sh while also preserving whitespace.

Your idea of using `\0` as a delimiter simply does not work; both bash and dash eat any `\0` characters that are stored in variables.

tmpfiles.sh is currently broken when filesnames have whitespace, so my patchset does not introduce a regression here. If you want to fix it, please undertake it as a separate issue.

dwfreed:
@williamh any objections to merging this?

PR #97, opened 2016-09-02
Format code blocks and variable/path notations

frickler01:
Added markdown backticks for commands, variable names and path as well es indented code blocks for better readability.


williamh:
In the commit message, please use the present tense, so "add" instead of "added". Also, I believe there is a way to format code in markdown, so it might be better to use that instead of tabs.

dwfreed:
Start with 3 backticks (the character to the left of 1 on a standard US QWERTY keyboard), optionally specify a language, newline, code, newline, end with 3 backticks.

frickler01:
@williamh I'm aware of that, I just thought indented code is better distinguished in plain text than the fenced variant. But I can change that of course.

PR #102, opened 2016-10-24
sh/init.sh.Linux.in: skip /proc sanity test if md5sum is missing

floppym:
Bug: https://bugs.gentoo.org/594534#c7

### DIFF
--- a/sh/init.sh.Linux.in
+++ b/sh/init.sh.Linux.in
@@ -21,7 +21,7 @@ fi
 mountproc=true
 f=/proc/self/environ
 if [ -e $f ]; then
-	if [ "$(VAR=a md5sum $f)" = "$(VAR=b md5sum $f)" ]; then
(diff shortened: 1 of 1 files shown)


dwfreed:
meh

williamh:
If I have my choice, I would rather get rid of the /proc sanity test and just try to mount /proc if it isn't mounted.

robbat2:
As the person that wrote this code, and cares about the case, please merge floppym's PR, and don't drop the /proc check.

williamh:
Here's where I am on this p/r. We can keep the /proc check, but I question the value of the separate check for md5sum. The only time md5sum is not accessible is if /usr is separate and not mounted.

floppym:
This is a simple patch to keep OpenRC working without /usr mounted. If you want to actively break that by refusing patches, that's up to you.

Merge this or close it.

dwfreed:
Merged my PR.  Thanks again for your patch.

PR #103, opened 2016-10-24
init.d/loopback: drop scope on loopback

robbat2:
Busybox does not support the 'scope' argument on 'ip address add' or 'ip
route add', this is documented in BUSYBOX.md, but is no longer actually
needed, as the kernel does get it right without manual specification,
and the ifconfig variant already relies on the kernel to get it right.

X-Gentoo-Bug: 487208
X-Gentoo-Bug-URL: https://bugs.gentoo.org/show_bug.cgi?id=487208
Signed-off-by: Robin H. Johnson robbat2@gentoo.org

### DIFF
--- a/init.d/loopback.in
+++ b/init.d/loopback.in
@@ -21,8 +21,8 @@ start()
 	if [ "$RC_UNAME" = Linux ]; then
 		ebegin "Bringing up network interface lo"
 		if command -v ip > /dev/null 2>&1; then
-			ip addr add 127.0.0.1/8 dev lo brd + scope host
-			ip route add 127.0.0.0/8 dev lo scope host
+			ip addr add 127.0.0.1/8 dev lo brd +
+			ip route add 127.0.0.0/8 dev lo
 			ip link set lo up
 		else
 			ifconfig lo 127.0.0.1 netmask 255.0.0.0


dwfreed:
Can we drop the explicit route while we're at it?  The kernel should be automatically adding that route already.

robbat2:
Yes, we can drop it as well, if we're ok with it being just in the `local` routing table, rather than the `main` routing table.

williamh:
@robbat2 @dwfreed Does it matter which routing table the loopback route is in? If not I'll remove the explicit route in a separate commit.

robbat2:
@dwfreed 
we need to check if it's safe to drop the explicit route on BSD.
On Linux it's safe because the implicit route has a much higher priority.

dwfreed:
It should not, because more specific routes always take precedence over less specific routes (so an explicit 'default via <gateway>' doesn't take precedence over the 127/8 route for lo0

williamh:
I still need an update from the *bsd side. Thanks much.

pkgdemon:
I fail to see how this change is going to affect the BSD's at all.  The BSD's do not use ip addr.  They would use this block which I see no changes to.

```
ifconfig lo 127.0.0.1 netmask 255.0.0.0
route add -net 127.0.0.0 netmask 255.0.0.0 gw 127.0.0.1
```

dwfreed:
@pkgdemon We're asking if we can drop the explicit route call in the BSD codepath too, and still have people that bind to things like 127.0.0.2, etc, work.

pkgdemon:
I cannot speak for the other BSD distributions.  For FreeBSD 10.3, and higher if I start in single user mode to bypass /etc/rc:

``ifconfig up lo0```

```
lo0: flags=8049<UP,LOOPBACK,RUNNING,MULTICAST> metric 0 mtu 16384
        options=600003<RXCSUM,TXCSUM,RXCSUM_IPV6,TXCSUM_IPV6>
        inet6 ::1 prefixlen 128 
        inet6 fe80::1%lo0 prefixlen 64 scopeid 0x2 
        nd6 options=21<PERFORMNUD,AUTO_LINKLOCAL>
```

You can see it adds an ipv6 route but not an ipv4 route.

```
Internet6:
Destination        Gateway            Flags     Netif Expire
localhost          link#2             UH          lo0
fe80::feaa:14ff:fe link#1             UHS         lo0
fe80::%lo0/64      link#2             U           lo0
fe80::1%lo0        link#2             UHS         lo0
```

Issuing an ipv4 address by doing the following:

```
ifconfig lo0 127.0.0.1 netmask 255.0.0.0
```

Will add this route:

```
Internet:
Destination        Gateway            Flags     Netif Expire
localhost          link#2             UH          lo0
```

This is actually invalid for FreeBSD:
```
route add -net 127.0.0.0 netmask 255.0.0.0 gw 127.0.0.1
```

It has to be:
```
route add -net 127.0.0.0/8 127.0.0.1
```

This will add:
```
Internet:
Destination        Gateway            Flags     Netif Expire
yournet          localhost             UGS          lo0
```

To summarise I don't think route add -net `127.0.0.0 netmask 255.0.0.0 gw 127.0.0.1` in the case of FreeBSD is needed at all.  The inet6 routing is added just by bringing the interface up.  The inet routing is added just by issuing `ifconfig lo 127.0.0.1 netmask 255.0.0.0` alone.

dwfreed:
@pkgdemon Alright, thanks, I'll just drop it there, too, then.

dwfreed:
(Closing PR as Robin's changes have been merged)

PR #106, opened 2016-12-15
sh/openrc-run.sh: expose default start/stop/status

robbat2:
Supervisor setups break easily when start/stop/status functions are not
default.

Applications that write multiple PIDs to a pidfile (eg HAProxy as
described in bug 601540), can also benefit from being able to call the
default start/stop/status with modified environment variables.

Expose the default start/stop/status functions as
openrc_default_start/stop/status, and use them for the defaults
start/stop/status.

Trivial usage example:
```
  stop()
  {
    t=$(mktemp)
    for pid in $(cat $pidfile) ; do
      echo $pid >$t
      pidfile=$t openrc_default_stop
    done
    rm -f $t
  }
```

X-Gentoo-Bug: 601540
Signed-off-by: Robin H. Johnson <robbat2@gentoo.org>

### DIFF
--- a/sh/openrc-run.sh.in
+++ b/sh/openrc-run.sh.in
@@ -146,10 +146,9 @@ _status()
 	fi
 }
 
-# Template start / stop / status functions
 # These functions select the appropriate function to call from the
 # supervisor modules
-start()
+default_start()
 {
 	local func=ssd_start
 	case "$supervisor" in
@@ -163,7 +162,7 @@ start()
 	$func
 }
 
-stop()
+default_stop()
 {
 	local func=ssd_stop
 	case "$supervisor" in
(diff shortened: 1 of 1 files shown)


williamh:
I like this, but I think we can do without the openrc_ prefix on the function names.

robbat2:
Changed to `default_` prefix instead of `openrc_default_`.

williamh:
Please add the X-Gentoo-Bug-URL line to your commit message. Also, if you want you can remove
Signed-Off-By since it really doesn't mean anything in OpenRC.
I understand that you prefer to have the gpg signatures preserved on your commits, so please let me know when you want this merged and rebase it.

PR #107, opened 2016-12-17
Localmount shouldn't mount remote filesystems

asomers:
The /etc/init.d/localmount script has a syntax error that causes it to
attempt to mount remote filesystems, causing the boot to fail. The
script appends a "no" to each remote filesystem type, but it should only
be append the "no" to the beginning of the list.  This patch fixes
localmount on FreeBSD 12.0.  A review of the mount(8) manpage on Ubuntu
12.04 suggests that this patch is correct for Linux, too.

Downstream bug:
https://github.com/trueos/trueos-core/issues/220


dwfreed:
That is incredibly not intuitive...  I'll get this merged today.

PR #108, opened 2016-12-17
sh/init.sh.Linux.in: skip /proc test if no md5sum

dwfreed:
This will also warn users if md5sum is missing, which serves as a pretty
good indicator that /usr is not mounted.

Supercedes #102

### DIFF
--- a/sh/init.sh.Linux.in
+++ b/sh/init.sh.Linux.in
@@ -11,6 +11,16 @@ if [ -d "@SYSCONFDIR@/rc.conf.d" ]; then
 	done
 fi
 
+# check for md5sum, and probably /usr too
+if command -v md5sum >/dev/null; then
+	got_md5sum=true
+else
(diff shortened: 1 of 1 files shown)


williamh:
I wouldn't refer to coreutils in the message, because it could be a busybox system that doesn't have md5sum turned on.
Also, I'm wondering about the value of the else branch. Do we need to print out the message about /proc being already mounted?

dwfreed:
> I wouldn't refer to coreutils in the message, because it could be a busybox system that doesn't have md5sum turned on.

The people using busybox to replace coreutils will know they're using busybox to replace coreutils and can figure that out.

> Also, I'm wondering about the value of the else branch. Do we need to print out the message about /proc being already mounted?

Message parity.  We print a message saying we're mounting /proc if it's not mounted, so might as well print something when it is already mounted.

PR #110, opened 2016-12-24
fix confusing indentation

lainegholson:
they were causing -Wmisleading-indentation warnings with GCC 6


williamh:
I believe this is taken care of in 0.23.1.
Thanks much, and if you see anything else please let me know. :-)

PR #111, opened 2016-12-27
start-stop-daemon: Initial support EnvironmentFiles

mrueg:
This would enable services to read from a generic environment file besides conf.d/service.conf.

See also:
https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=

Two smaller parts of the specification are not implemented yet:
* Support multiline variable definitions
* Strip whitespaces unless double quoted

### DIFF
--- a/man/start-stop-daemon.8
+++ b/man/start-stop-daemon.8
@@ -111,6 +111,8 @@ Same as the
 option.
 .It Fl e , -env Ar VAR=VALUE
 Set the environment variable VAR to VALUE.
+.It Fl E , -envfile Ar FILE
+Set environment variables via a EnvironmentFile.
 .It Fl g , -group Ar group
 Start the daemon as in the group.
 .It Fl k , -umask Ar mode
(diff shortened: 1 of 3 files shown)


williamh:
I'm adding a note here after speaking with you in irc. We can't merge this yet since it hasn't bee tested. Please update the p/r once you have tested this and we'll go from there.

dwfreed:
I have several issues with this:

1. Not all services use start-stop-daemon (particularly ones that don't use daemons, or ones that do really tricky things).
2. Environment files as implemented by systemd are incredibly basic, whereas /etc/conf.d/service is sourced by shell, and thus can contain whatever shell code you want.
3. This leads to environment configuration in several different places, when OpenRC's canonical location for that has always been in /etc/conf.d/service, and this provides no way for that location to override the environment file.

If you need environment variables set for a daemon, just put `export SOMEVAR="somevalue"` in /etc/conf.d/service or the initscript as appropriate.

:-1:

mrueg:
Yeah I haven't thought of conf.d as an alternative when working on it. Closing it then.

PR #112, opened 2017-01-07
sysfs: don't pull modules to sysinit runlevel

maciejsszmigiero:
Commit 73cdf10f1f51 ("Deprecate automatic loading of modules") introduced
"want modules" dependency for the "sysfs" service.

By default the "sysfs" service is installed in the "sysinit" runlevel, is
one of the very first services to start on boot (often it is the first one)
and it is always started before udev since udev depends on it.
The "modules" service on the other hand is supposed to start a bit later,
in the "boot" runlevel.

However, since that mentioned earlier commit the "modules" service will be
pulled in to the "sysinit" runlevel because a "want" dependency will do it
regardless in which runlevel this service is actually installed.
This breaks things like USB modem mode switching.

Change this dependency to an "use" which does not automatically pull in the
"modules" service into the "sysinit" runlevel and inform an user that he
needs to add it manually to this runlevel in case that he boots an EFI
Linux system AND efivarfs support wasn't built in into the kernel AND
efivarfs module wasn't loaded already by an initrd.

### DIFF
--- a/init.d/sysfs.in
+++ b/init.d/sysfs.in
@@ -15,7 +15,7 @@ sysfs_opts=nodev,noexec,nosuid
 
 depend()
 {
-	want modules
+	use modules
 	keyword -docker -lxc -prefix -systemd-nspawn -vserver
 }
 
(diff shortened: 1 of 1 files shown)


maciejsszmigiero:
PCMCIA/CardBus cards that remained inserted during system startup are another victims of this issue.

williamh:
Honestly, given that pc cards are also affected, I would rather drop this dependency and not have sysfs pull in modules.

maciejsszmigiero:
Then efivarfs module loading in init.d/sysfs will need to be un-deprecated.

williamh:
My understanding is that the kernel will load the module automatically when you run the mount command, so I shouldn't have to do anything special to load it.

williamh:
This is fixed in http://github.com/openrc/openrc/commit/6a79aef

maciejsszmigiero:
init.d/sysfs from your commit still uses modules, but you wrote that you would like to drop this dependency.
Was this intentional?

PR #113, opened 2017-01-08
start-stop-daemon: allow all standard signals

dwfreed:
X-Gentoo-Bug: 604986
X-Gentoo-Bug-URL: https://bugs.gentoo.org/show_bug.cgi?id=604986

### DIFF
--- a/src/rc/start-stop-daemon.c
+++ b/src/rc/start-stop-daemon.c
@@ -197,26 +197,45 @@ parse_signal(const char *sig)
 		int signal;
 	} SIGNALPAIR;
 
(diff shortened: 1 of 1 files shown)


dwfreed:
Updated to protect with ifdef those signals not defined in POSIX.2

williamh:
I can't tell which signals you added as easily with the addition of the signalpair_item macro, so can we do that switch later? Also, supervise-daemon will need the same signals added.

Thanks,

William

williamh:
@dwfreed, sorry about the incorrect statement about supervise-daemon. I looked over your code and will merge it, also adding a comment about the macro to the commit message.

PR #122, opened 2017-03-31
start-stop-daemon: give user a warning if calling --start with --retry or --stop with --wait

austin987:
Found while trying to figure out why a long running, slow to start up java webapp was always showing as 'crashed' when trying to shut it down


austin987:
Changed to use a warn rather than an error

PR #142, opened 2017-06-09
scripts/Makefile: respect SBINDIR with MKSYSVINIT

udeved:
Use SBINDIR in the scripts Makefile and fix the symlinks if MKSYSVINIT=yes

### DIFF
--- a/scripts/Makefile
+++ b/scripts/Makefile
@@ -22,7 +22,7 @@ ifeq (${MKSYSVINIT},yes)
 	ln -sf	${DIR}/halt ${DESTDIR}/${SBINDIR}/halt
 	ln -sf	${DIR}/poweroff ${DESTDIR}/${SBINDIR}/poweroff
 	ln -sf	${DIR}/reboot ${DESTDIR}/${SBINDIR}/reboot
-	ln -sf	openrc-shutdown ${DESTDIR}/${SBINDIR}/shutdown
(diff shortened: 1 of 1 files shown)


williamh:
The first commit on this p/r looks good, but I have concerns about the second and third ones. The first commit fixes sbindir on the symlinks and that is correct. In the second one, why do you convert the symlinks to absolute symlinks instead of leaving them as relative? The third one breaks the shutdown wrapper by linking it to shutdown instead of openrc-shutdown.

udeved:
Hi @williamh 

I am just experimeting, I perhaps should have done it i a separate patch branch.

Reaso for second commit was, that all the symlinks in the resulting package were broke.
The were a dead end symlink, pointing to a non existing file.
Now what I don't know, if this is connected to using a prefix as opposed to gentoo.
The third commit was donne, because I have had trouble to poweroff the system from plasma. It shutdown, but the message states "reboot" annd the system just keeps powered on. Something is still wrong with the shutdown.

udeved:
Reason to make it absolute path was this...

`ln -sf ..//usr/lib/rc/bin/rc-sstat /build/openrc/pkg/openrc//usr/bin/rc-sstat`

The double slash renders the symlink broke.

williamh:
Correct, you should always keep the master branch clean; all pull requests should be done with branches.

udeved:
Can you close it and I'll reopen a patch PR.

PR #144, opened 2017-06-12
.travis.yml: Initial travis configuration

mrueg:
* Build with gcc, clang on glibc and gcc on musl libc via musl-gcc
wrapper
* Execute tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,19 @@
+language: c
+
+dist: trusty
+sudo: required
+
+compiler:
+  - gcc
+  - musl-gcc
+  - clang
+
+install:
+  # This ppa provides a more recent musl
+  - sudo add-apt-repository ppa:bortis/musl -y
+  - sudo apt-get update -qq
+  - sudo apt-get install -qq musl-tools musl musl-dev
+
+script:
+  - make 
+  - make test


williamh:
To the rest of the team -- do not merge this for now, but keep it open. I need to figure out whether travis is accessible to me or not.

williamh:
I have opened https://bugs.gentoo.org/show_bug.cgi?id=621628 to inquire about stabilizing the Travis commandline client. I will follow that bug and look at this more once that is stabilized.

mrueg:
@williamh Is there still interest in this change?

williamh:
Yes there is, I just haven't  looked at it much yet. On another note, I'm thinking about unit testing, I'm just not sure how to do that with OpenRC yet either.

.travis.yml
```diff
@@ -0,0 +1,19 @@
+language: c
```
vapier: can you add to the top of this file:

```
# Travis build integration.
# https://docs.travis-ci.com/
```

.travis.yml
```diff
@@ -0,0 +1,19 @@
...
+
+dist: trusty
+sudo: required
```
vapier: why do we need sudo ?  the build/tests shouldn't need this.  i guess it's because you want to use apt below ?

it'd be better if we could just rely on addons->apt->packages to list the packages we care about.  if that means using an older musl package, or dropping that coverage entirely, then i think that's OK.

.travis.yml
```diff
@@ -0,0 +1,19 @@
...
+  - clang
+
+install:
+  # This ppa provides a more recent musl
+  - sudo add-apt-repository ppa:bortis/musl -y
+  - sudo apt-get update -qq
+  - sudo apt-get install -qq musl-tools musl musl-dev
+
+script:
+  - make
```
vapier: this can easily grow.  rather than do this, how about we create a `test/travis.sh` script to handle this logic and this line simply reads:

```
script: ./test/travis.sh
```

mrueg: I'd prefer a step-level in travis. Otherwise it can be difficult to figure out what failed on travis.

.travis.yml
mrueg: Sure.

.travis.yml
mrueg: On the other hand, what's the harm using sudo here?

.travis.yml
vapier: that's purely an issue of the script and how it logs.  making it a standalone script makes it a lot easier to add extended/conditional logic as well as reproduce/execute the code locally.

.travis.yml
vapier: it adds overhead to the system that isn't needed otherwise, you're pulling from a repo that is effectively random from our perspective, and why do we care about whether musl is the latest ?  we don't do that for gcc, clang, glibc, or other packages.  arguably, we should be building against the versions other people have available.

Issue #121, opened 2017-03-26
agetty login display


cromerc:
I just converted to using agetty in openrc instead of sysvinit's inittab. However there is a visual issue with the agetty login display when run from openrc. What happens is the login prompt is shown in the middle of other text instead of at the end. To get around this I put --delay in my /etc/conf.d/agetty file like so:
`agetty_options="--noclear --delay 1"`

This does fix the graphical issue of the text overlapping, but introduces a delay every time that I logout of my session.

Is there a better way to fix this without using --delay?

williamh:
I'm closing this, only because I don't know of a better way to fix it than what you are doing. If someone does, or if someone thinks OpenRC should do something different with this by default, please feel free to comment or open another issue.

cromerc:
Another solution would be to create an agetty runlevel that is called after default, and when it is called, use the --quiet argument so that the agetty runlevel doesn't show anything on the screen. This solution would require modifiying the openrc-init to detect and run an agetty runlevel if it exists.

PR #147, opened 2017-06-15
add quiet option for agetty

cromerc:
This adds a quiet option that can be used to silence ebegin and eend for services. This fixes the display issue that happens under openrc-init using the agetty services: https://github.com/OpenRC/openrc/issues/121

### DIFF
--- a/conf.d/agetty
+++ b/conf.d/agetty
@@ -6,3 +6,6 @@
 
 # extra options to pass to agetty for this port
 #agetty_options=""
+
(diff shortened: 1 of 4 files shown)


dwfreed:
you can accomplish the same thing by just exporting EINFO_QUIET="yes" in the service script.

cromerc:
Thanks, I had no idea about that option. I will make changes to the PR accordingly.

PR #156, opened 2017-07-31
s-s-d: warn about non-root writable pidfile

dwfreed:
Since start-stop-daemon is run as root, non-root writable pidfiles can
be used to make start-stop-daemon signal processes other than the daemon
the pidfile is supposed to be for.

Reported-By: Michael Orlitzky <mjo@gentoo.org>
Tested-By: Michael Orlitzky <mjo@gentoo.org>

### DIFF
--- a/src/rc/start-stop-daemon.c
+++ b/src/rc/start-stop-daemon.c
@@ -452,6 +452,7 @@ run_stop_schedule(const char *exec, const char *const *argv,
 	pid_t pid = 0;
 	const char *const *p;
 	bool progressed = false;
+	struct stat pidfile_stat;
 
 	if (exec)
 		einfov("Will stop %s", exec);
@@ -475,6 +476,13 @@ run_stop_schedule(const char *exec, const char *const *argv,
 		pid = get_pid(pidfile);
 		if (pid == -1)
 			return 0;
+
+		if (stat(pidfile, &pidfile_stat) == 0) {
+			if (pidfile_stat.st_uid != 0 ||
+			    pidfile_stat.st_mode & (S_IWGRP | S_IWOTH)) {
+				ewarn("%s is writable by non-root, which poses a security risk", pidfile);
+			}
+		}
 	}
 
 	while (item) {


vapier:
s-s-d can run as non-root just fine, so warning in that case is incorrect

also, can't s-s-d let the target program create/manage the pid, and then re-use it to stop ?  in which case it'll be owned by non-root ?

dwfreed:
> s-s-d can run as non-root just fine, so warning in that case is incorrect

I'll make the st_uid != 0 check also require geteuid() == 0.

> also, can't s-s-d let the target program create/manage the pid, and then re-use it to stop ? in which case it'll be owned by non-root ?

Daemons *should* be creating their pidfiles before they drop privileges, so the pidfile would be root-owned.  s-s-d's own --make-pidfile writes the pidfile before applying any --user/--group changes

dwfreed:
> I'll make the st_uid != 0 check also require geteuid() == 0.

Thinking further, st_uid != geteuid() would probably be better.  Thoughts?

vapier:
i don't see how being root owned is a requirement or even an improvement.  if s-s-d was launched as root, and it was responsible for creating the pid file, then sure, it'll be root owned.  but it's not inconceivable that the main daemon sets up the pid file so an unprivileged child can manage it.  i don't see that being a bug.

on my active system, i see transmission running with its pid file as owned by non-root.

i'm not really seeing the bug/issue this is trying to fix.

dwfreed:
A simple arbitrary code execution vulnerability in transmission could lead to a system-wide DoS when running-as-root s-s-d reads a pid from the non-root writable pidfile and signals that process, and somebody has used that vulnerability to write, say, '1' in that pidfile.

dwfreed:
And it doesn't even have to be arbitrary code execution; writing to a path that the attacker can specify with content the attacker can provide would be sufficient.

vapier:
that's why you pass the program name to the --stop function -- it does a sanity check on the target process.  if you aren't passing that, your init script is already broken and needs fixing.

there is no diff between "server crashed and pid was recycled to a diff program" and "a diff pid was manually injected to an existing process".

dwfreed:
It doesn't.  If --pidfile is provided to --stop or --signal, it is considered the only source of truth.  I imagine this was inherited from Debian's implementation when OpenRC forked it many many moons ago.

dwfreed:
https://github.com/OpenRC/openrc/blob/master/src/rc/start-stop-daemon.c#L407 is where that happens

williamh:
Can someone check Debian's s-s-d for sure and let us know if it just treats the pidfile as the source of truth or if it requires the program name as well?

orlitzky:
I've found another common mistake along the same lines. Often, people don't realize that `command_background=true` will write the PID file as `root:root` regardless of `command_user`. As a result, they change ownership of the PID file's containing directory to the daemon user, and the PID file (owned by `root:root`) gets written into a directory that a  non-root user controls. For example:

* [Bug 628542, app-admin/consul-template](https://bugs.gentoo.org/show_bug.cgi?id=628542)
* [Bug 628544, app-admin/consul](https://bugs.gentoo.org/show_bug.cgi?id=628544)
* [Bug 628546, app-admin/filebeat](https://bugs.gentoo.org/show_bug.cgi?id=628546)
* [Bug 628548, app-admin/glance](https://bugs.gentoo.org/show_bug.cgi?id=628548)

Even if it's only as a QA warning, it's looking like it would be extremely useful to extend this warning to the containing directory (i.e. if a non-root user can _replace_ the PID file).

orlitzky:
> Can someone check Debian's s-s-d for sure and let us know if it just treats the pidfile as the source of truth or if it requires the program name as well?

I haven't tested it, but reading the source at

  https://anonscm.debian.org/cgit/dpkg/dpkg.git/tree/utils/start-stop-daemon.c

suggests that the PID file takes precedence over the program name.

williamh:
Can this be updated so I can merge it, or should we close it?

Thanks,

William

williamh:
@dwfreed What is the status of this p/r? should I close it, or can you give me an update?

PR #162, opened 2017-09-05
SERVICE-DOS-AND-DONTS.md: new tips document for service writers

orlitzky:
Looking for initial feedback on this.

I'm willing to change the name, but be aware that my second choice is `TIPS4SCRIPS.md` =)

### DIFF
--- a/SERVICE-DOS-AND-DONTS.md
+++ b/SERVICE-DOS-AND-DONTS.md
@@ -1,0 +1,381 @@
+This document is aimed at upstream and distribution developers who
+write OpenRC service scripts, either for their own projects, or for
+the packages they maintain. It contains advice, suggestions, tips,
+tricks, hints, and counsel; cautions, warnings, heads-ups,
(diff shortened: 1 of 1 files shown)


mrueg:
I would suggest to call the file Service, "ServiceDefinition" or "InitDefinition"

williamh:
This was just brought to my attention, and I see a couple of things:

/run is only good for Linux, /var/run is more likely what people will want on other OS's.
The document really doesn't explain its purpose, so you might want to start by doing that.
A number of the tips about backgrounding and the pid files only apply if using start-stop-daemon to start the daemon. Take a look at the other markdown documents to see what I'm talking about for this, and I'll answer any questions I can.

orlitzky:
I finally got back to work on this. I added a short "statement of purpose" to the top that explains what the document is about, mentioned `/var/run`, added a new section on "need net" dependencies, and cleaned up some more minor issues.

orlitzky:
I renamed the file to something less cheeky but more awkward *shruuuuug*

PR #169, opened 2017-10-16
start-stop-daemon: check for errors when opening /dev/null and /dev/tty

floppym:


williamh:
@floppym Do we need to do this in supervise-daemon as well? if so, can you update it as well on this p/r?

PR #175, opened 2017-10-31
add meson build files

williamh:


liugang:
```
rootlibdir = join_paths(rootprefix, get_option('libdir').split('/')[-1])
```
on x86_64, it will be `x86_64-linux-gnu`, after install, it will be `/x86_64-linux-gnu`. 


Need use `add_project_arguments(cc_flags,  language : 'c')` instead of `add_global_arguments(cc_flags,  language : 'c')`, because any project can be used as a subproject.

williamh:
I changed add_global_arguments to add_project_arguments.

Also, I added messages at the bottom of the build so we can see the default root paths. It looks to me like the default libdir is  lib64 on linux/amd64, which means the default rootlibdir is /lib64.

williamh:
@liugang This is the start of a new release cycle, so I want to re-visit getting this in the tree. Can  you take a look and let me know what you can see that I may have missed?

Issue #180, opened 2017-11-05
start-stop-daemon: can't stop daemon running using non-root user inside lxc container with sys_ptrace capability dropped


iDarkTemplar:
Openrc version 0.34.5.

Steps to reproduce:

1) use lxc-2.* (not sure if it reproduces with other versions, I've tried only 2.0.8 and 2.1.1)
2) create gentoo container (lxc-create -n $name -t gentoo)
3) add to container config:
lxc.mount.entry=run run tmpfs rw,nosuid,nodev,relatime,mode=755 0 0
lxc.cap.drop = sys_ptrace
4) add network to config and run container
5) run 'ACCEPT_KEYWORDS="~*" emerge -1 openrc' inside container
6) add line to etc/rc.conf inside container:
rc_sys="lxc"
7) restart container, attach to it and do rest of steps inside of container
8) emerge dev-vcs/git
9) mkdir /var/git
10) /etc/init.d/git-daemon start
11) /etc/init.d/git-daemon stop

Stop action fails.

Without 'sys_ptrace' capability start-stop-daemon can't find out cgroups namespace of process belonging to another user when pid is provided, and thus fails to stop daemon. I'm also dropping other capabilities like sys_admin, but dropping this one specifically make openrc fail. If I rerun start-stop-daemon command again manually after that, it successfully stops git-daemon (since there's no longer a pidfile present, and it looks using other identifiers), but since it uses executable name, it will just kill every git daemon process.

```
test4 ~ # /etc/init.d/git-daemon status
 * status: started
test4 ~ # whoami
root
test4 ~ # ps aux | grep git
nobody     306  0.0  0.0  17052  1068 ?        Ss   12:58   0:00 /usr/bin/git daemon --syslog --base-path=/var/git
nobody     309  0.0  0.0  16044   988 ?        S    12:58   0:00 /usr/libexec/git-core/git-daemon --syslog --base-path=/var/git
root       341  0.0  0.0  10708   968 ?        S+   13:02   0:00 grep --colour=auto git
test4 ~ # ls -la /proc/306/ns
ls: cannot read symbolic link '/proc/306/ns/net': Permission denied
ls: cannot read symbolic link '/proc/306/ns/uts': Permission denied
ls: cannot read symbolic link '/proc/306/ns/ipc': Permission denied
ls: cannot read symbolic link '/proc/306/ns/pid': Permission denied
ls: cannot read symbolic link '/proc/306/ns/pid_for_children': Permission denied
[...]

PR #181, opened 2017-11-12
Temporarily change euid and egid when searching for processes with sp…

iDarkTemplar:
…ecific pid.

This should fix https://github.com/OpenRC/openrc/issues/180

https://github.com/gentoo/gentoo/pull/6176 is also needed to fix issue with git-daemon

### DIFF
--- a/sh/start-stop-daemon.sh
+++ b/sh/start-stop-daemon.sh
@@ -83,6 +83,7 @@ ssd_stop()
 		${command:+--exec} $command \
 		${procname:+--name} $procname \
 		${pidfile:+--pidfile} $chroot$pidfile \
+		${command_user+--user} $command_user \
 		${stopsig:+--signal} $stopsig \
 		${_progress}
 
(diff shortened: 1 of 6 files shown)


williamh:
Can you please check the patch I posted on the original issue? If you are just running into a permission issue when checking the name spaces, I would rather just skip checking the name spaces than change uid/gid to avoid the permission issue.

iDarkTemplar:
I already did and responded in original issue: https://github.com/OpenRC/openrc/issues/180#issuecomment-343987970

williamh:
I will keep this open for a release or so to monitor for more issues, but I don't think we need to do this since the issue that it addresses is actually in librc itself.

williamh:
The searching is done in rc_find_pids, so it would probably be better to do anything like this there than in either of the supervisors.

iDarkTemplar:
Closing this request since issue is fixed in 0.34.7.

PR #182, opened 2017-11-15
add "modules-late" service

maciejsszmigiero:
Some modules, when loaded, trigger hotplugging of services that in turn
need some services from the "default" runlevel (like dbus) or which need
local filesystems to be mounted (including the root filesystem (re)mounted
read / write).

This is especially true for devices like NICs or USB modems, but almost
certainly not limited to them.

For example, in current netifrc a wireless NIC handled by wpa_supplicant
module needs dbus service to be started (if wpa_supplicant was built with
dbus support) and these NICs which are handled by dhcpcd need
/var/lib/dhcpcd/ writable (in practice this most often means a need to
have the root filesystem writable).
We order the normal "modules" service before "fsck" service (and so before
remounting the root fs read / write) so filesystems that are on devices
that need some modules loaded can be checked.

There used to be also a problem with usb_modeswitch needing a write access
to the root filesystem when switching mode of an USB modem (see Gentoo bug
565262). It was possible to work it around that time, but since programs
are clearly expecting a root fs being writable it can be reasonably
expected that similar issues will pop up in the future.

To fix this once for all let's introduce a "modules-late" service started
early in the "default" runlevel which can be used to load modules that have
the requirements described above.
This service is implemented via a slight modification of the existing
"modules" script.


williamh:
Hello,

The issue is not with loading the modules. All modules should be loaded as early as possible. That is why we order the modules and modules-load services the way we do.

The problem is that when services get hotplugged, they get pulled out of the normal boot sequence. So you get situations where services are starting much earlier than they would when they are not hotplugged.

This is off by default because we have no way of knowing how your system is configured. You are discussing the root file system, but /var can also come into play, and if you are using /var on a separate file system, is it a local file system, or is it a network filesystem? We don't know so we don't try to set up hotplugging services by default.

You are saying that the root filesystem being r/w would solve a lot of this, but that's not really the case. Remember that /var can be on its own file system as well, so we would have to check for that, and in theory /var can be a local filesystem or even a network file system.

Also, since the services you are hotplugging are being pulled out of the default runlevel to begin with, this modules-late service wouldn't be the answer.

The answer would be to add the appropriate rc_net_if_need="foo bar bas" lines to /etc/conf.d/net if you are doing this to make sure the appropriate services are run.

I will warn you though, that hotplugging services and sorting out what should run when will be messy; this is why hotplugging services is off by default.

maciejsszmigiero:
> The problem is that when services get hotplugged, they get pulled out of the normal boot sequence.
> So you get situations where services are starting much earlier than they would when they are not
> hotplugged.

Yes, that's the problem.
Some services should start only when the "default" runlevel is reached, but due to hotplugging are started when the "boot" runlevel is still running.
That's why this pull request introduces a "modules-late" service that is supposed to be in the "default"
runlevel.
Modules loaded by this service can then safely cause hotplugging of services from the "default"
runlevel.

> You are discussing the root file system, but /var can also come into play, and if you are using /var
> on a separate file system, is it a local file system, or is it a network filesystem?

This (root file system) was used only as an example, as you said the problem could happen also when some other file system (that is not mounted yet or not mounted read / write yet) is needed by some hotplugged service triggered by a module loaded by the "modules" service.

> Also, since the services you are hotplugging are being pulled out of the default runlevel to begin with, 
> this modules-late service wouldn't be the answer.

This "modules-late" service is supposed to be in the "default" runlevel so it is fine if it causes some
services from this runlevel to start.

> I will warn you though, that hotplugging services and sorting out what should run when will be messy;
> this is why hotplugging services is off by default.

All right, but hotplugging is needed for things like USB network cards that can be plugged in at any time (or alternatively could remain plugged in all the time during a system startup).

What I would like to underline is that this is a second time such problem occurred (previously there
was a problem with USB modems and usb_modeswitch, now with WiFi NICs and wpa_supplicant / dhcpcd) which is solved by introducing this service (I have tested this, naturally).

As I wrote in the description of this pull request, it can be reasonably expected that similar issues will pop up in the future, again.

LW-archlinux:
> Some modules, when loaded, trigger hotplugging of services

I am not aware of any kernel modules that require services.
Do you mean devices that depend on a certain module ALSO need services running to be used succesfully ?

This sounds like something that could be solved much easier with a custom (e)udev rule.
http://www.reactivated.net/writing_udev_rules.html , check the advanced topics .

maciejsszmigiero:
> I am not aware of any kernel modules that require services.

Examples: USB network cards, USB modems, network cards that can be either present or not present during particular system startup in general...

> This sounds like something that could be solved much easier with a custom (e)udev rule

That's exactly how network interface service hotplugging is currently triggered (via udev) - see
/lib/udev/rules.d/90-network.rules.
This problem is not dependent on what exactly triggers the hotplugging, only that it can be triggered
when a module loads.

maciejsszmigiero:
It is also worth noting that even with service hotplugging disabled this problem will still occur if a udev
rule triggered by loading a kernel module needs something like root file system mounted read / write 
(or more generally, a completed "boot" runlevel) - in fact that's exactly what happened previously with usb_modeswitch.

LW-archlinux:
I have  a lot of expereince with IT troubleshooting, hardeware/software, cdesktop / servers and there's one rule i try to stick to always : treat the root cause , not the symptoms .

module already loaded is not enough to use device effectively is  a symptom .
The root cause : insertion of a device when the system doesn't have everything in place needed.

Solve the root cause by triggering something on device insertion that makes sure everything that's needed is started.

maciejsszmigiero:
> Solve the root cause by triggering something on device insertion that makes sure everything that's
> needed is started.

You get a circular dependency this way.
From this PR description:
> We order the normal "modules" service before "fsck" service (and so before
> remounting the root fs read / write) so filesystems that are on devices
> that need some modules loaded can be checked.

The "root" service needs to be started after "fsck",
the "fsck" service needs to be started after "modules".

But due to issues described in this PR the "modules" service needs to be started after "root".

This PR solves this circular dependency by splitting the "modules" service into two parts:
1) The first one which is needed by the "fsck" service (the old "modules" service),
2) The second one which needs the "root" service (the new "modules-late" service).

LW-archlinux:
You claim that some modules need to be started after "root" , but i think that's due to a flawed analysis.

Let's see if i can help with the problem analysis.
First let's get a clear picture of the problem.

examples you gave : 

- wireless  card needs wpa_supplicant, wpa_supplicant needs dbus service 
- usb modem needs write access to /var (probably needs localmount service for that )


There are (atleast) 2 ways to have openrc start services on demand : 
- use of rc_hotplug variable in rc.conf
- triggered by an external program like udev (using rc-service )

What triggers the problem
- external program
- device insertion detected by kernel/udev 
- both
- something else  ?

maciejsszmigiero:
> wireless card needs wpa_supplicant, wpa_supplicant needs dbus service

Also dhcpd needs /var/lib/dhcpcd/ writable (provided by "root" or "localmount" service) both for wireless and wired NICs as long as they use dhcpd for their configuration.

> usb modem needs write access to /var (probably needs localmount service for that )

usb_modeswitch used to need write access to /tmp (this has been worked around for now, however an
underlying problem remains).

> What triggers the problem

Usually it is triggered by a module being loaded by "modules" service on boot but you can assume that if you plug in a device in wrong moment (when the boot runlevel is still being started) it can happen then via a module loaded on kernel request, too.

Generally, what triggers it is a module load which then triggers a udev event which triggers some application (directly or via some hotplugged service) which needs the things described above.

robbat2:
As the netifrc maintainer, I would like to bring a further view to the table. Because even the modules-late proposal is problematic when a given module is built into the kernel. It's also not limited to networking.

What we know:
* The system is in the `boot` or `sysinit` runlevel.
* udev is running.
* The module gets loaded or is already built into the kernel.
* The kernel generates an event for the device appearing.
* udev gets the event.
* udev fires a hotplug script fires for the event, per some rules:
  * `25-gpsd.rules`
  * `70-spice-webdavd.rules`
  * `90-network.rules`
  * `99-pcscd-hotplug`
  * `99-space-navigator.rules`
  * others...
* The hotplug script starts or tries to start a service.
* The service is meant for usage during the `default` runlevel, and causes many other services to start because it's too early.

The `modules-load` proposal tries to defer the problem, and only load the module when we are in the default runlevel. This won't work if the kernel code was built-in instead of a module, or already loaded by the initramfs.

I think that OpenRC should capture the start event, and delay starting the service until a later runlevel.

maciejsszmigiero:
Thanks for your input Robin.

> This won't work if the kernel code was built-in instead of a module, or already loaded by the 
> initramfs.

I think these cases are a parallel, but a bit different problem, since they are handled by coldplugging 
as OpenRC isn't running yet when these devices are instantiated.
Some coldplugging could be deferred, too.

> I think that OpenRC should capture the start event, and delay starting the service until a later
> runlevel.

I agree that this would be the best solution to this issue, as long as udev events can be delayed, too
(since they can launch programs like usb_modeswitch).

However, since I don't know OpenRC internals I don't know how big a change is needed to implement
such functionality.
Since the problem is occurring right now if that solution will only be available at an unspecified future time then it would be best to have for now at least something that works.

maciejsszmigiero:
Rebased this pull request onto current master.

It is worth noting that I have OpenRC with this change deployed for few months already on few 
machines without any issues.
Also, this is a backward-compatible change: nothing changes for existing OpenRC setups, unless
a user specifically starts to make use of new functionalities added by it.

maciejsszmigiero:
Rebased this pull request onto current master.

williamh:
Let's revive discussion here a bit.

To be honest, I've never set RC_HOTPLUG to any value in OpenRC, so I do not see the value of hotplugging services.
Given how problematic hotplugging services is as discussed in this mr, I think I would rather go the opposite direction and not allow it.

Since it isn't on by default in the first place, are there a number of people using it? Would this feature be missed?

Let me know what you think.

maciejsszmigiero:
Hotplugging is needed for devices on hot-pluggable buses like USB or even PCIe (ExpressCard or virtualization) since they can show up at any time, and then the relevant service needs to be started (network, pcsc-lite, usbmuxd, ...).

I am using this functionality to have reliable wireless and 3G modem support at boot for more than 3 years now.
Before that, I was hitting race conditions with respect to the services in the "boot" runlevel.

Rebased this pull request onto current master.

williamh:
I'm just brain-storming, but I wonder if the better answer would be to make sure the hotplugged services don't start until after the boot runlevel is completed?

maciejsszmigiero:
Not only hotplugged services would need to be delayed until the "boot" runlevel completes, but udev events too since some of them need root fs mounted read / write (for example these that call usb_modeswitch).

But then there is a problem that some of these udev events might be needed for the "boot" runlevel services.
For example, events configuring block devices (and their handlers) may be needed for the "localmount" service.

These udev events would need to be split into instant and delayed ones.
In the end it basically would be the same solution as when splitting the "modules" service (which is what this pull request does).
Only it would probably be harder to configure for an user, since udev events aren't normally user-visible. Unlike module names.

lu-zero:
We'd need to consider adding the concept of delayed dependency so when we are made aware of an event and the service that would trigger is in another runlevel not yet reached its activation would be scheduled once you get in that runlevel.

Would be nice to add the whole machinery so you can 
```
rc-update add-hotplugged foo runlevel
rc-update del-hotplugged foo runlevel
```

I guess would have the simplest user-experience

maciejsszmigiero:
As I wrote in my previous message, delaying just OpenRC services would not solve the udev event part.

lu-zero:
If the `udev` part depends on a specific system state that depends on an `openrc` service in general, it would be good to make so it works correctly by having `udev` trigger `openrc` and having `openrc` make sure the required services are available before.

Alternatively I would push your approach further and have the modules initscript behave like `netifrc`. This way the user would a good way to load groups of modules with a good amount of granularity.

williamh:
I'm still working on wrapping my head around this; I am trying to understand how other init systems handle it.
Here is what I think I know so far.

Kernel modules can be loaded at several points during the boot sequence:

0) they can be built in
1) they can be loaded from an initramfs
2) they can be loaded by udev
3) the modules service can load them

No matter when they are loaded in the sequence, they trigger events, which cause udev rules to be run, and these rules can do things that are outside our control and need access to things that may not be available during early boot.
An I correct so far?

williamh:
@maciejsszmigiero  I have another question specifically for you.
I'm not sure it matters, but please bear with me. In ```/etc/rc.conf```, do you have ```RC_HOTPLUG``` set? If so, what is it set to?

Thanks much.

maciejsszmigiero:
> If the `udev` part depends on a specific system state that depends on an `openrc` service in general, it would be good to make so it works correctly by having `udev` trigger `openrc` and having `openrc` make sure the required services are available before.

The problem is that some udev events are needed also for the services in the "boot" runlevel.
For example, events configuring block devices (and their handlers) may be needed for the "localmount" service.

So, these udev events would need to be split into instant and delayed ones.
In the end it basically would be the same solution as when splitting the "modules" service.

> No matter when they are loaded in the sequence, they trigger events, which cause udev rules to be run, and these rules can do things that are outside our control and need access to things that may not be available during early boot.
> An I correct so far?

Yes.
A small footnote here: please note that situation for the modules that are built-in into the kernel (and, to some extent, these loaded by initramfs, too) is different: their events are processed as "coldplug events" by the "udev-trigger" service when this service starts.
While the udev events generated by the modules loaded by the "modules" service are processed directly when they get loaded (well, more or less so, technically the event processing is asynchronous).


> In `/etc/rc.conf`, do you have `RC_HOTPLUG` set? If so, what is it set to?

I have it set to allow hotplugging of all services:
> rc_hotplug="*"

maciejsszmigiero:
Rebased this pull request onto current master.

Since this patch solves a real issue (that I explained in many comments here over the last 4+ years) I would like to have it finally merged.
Thanks.

PR #184, opened 2017-11-17
Add more arguments available as variables

j8r:
Brings more arguments available as variables in the service's configuration.

The small breaking change is the `$command_user` variable changed simply to `$user`, with the same name as its related argument, `--user`.

### DIFF
--- a/man/openrc-run.8
+++ b/man/openrc-run.8
@@ -142,12 +142,12 @@ List of arguments to pass to the daemon when starting via
 .Nm supervise-daemon .
 to force the daemon to stay in the foreground
 .It Ar command_background
-Set this to "true", "yes" or "1" (case-insensitive) if you want 
+Set this to "true", "yes" or "1" (case-insensitive) if you want
 .Xr start-stop-daemon 8
 to force the daemon into the background. This forces the
 "--make-pidfile" and "--pidfile" options, so the pidfile variable must be set.
 .It Ar command_progress
-Set this to "true", "yes" or "1" (case-insensitive) if you want 
+Set this to "true", "yes" or "1" (case-insensitive) if you want
 .Xr start-stop-daemon 8
 to display a progress meter when waiting for a daemon to stop.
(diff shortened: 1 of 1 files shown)


williamh:
Please stay with command_user.
Also, please document all new variables you add on the openrc-run.8 man page.

Thanks,

William

j8r:
Hi William,

Ok I will add docs on the man pages.

Why having `--user` and `--group` and not having their equivalent variables, like all other ones? In order to be consistent, we should change to `--command_user` too.

By the way, I suppose we will need to change to `command_group`, `command_stdout`, `command_stderr` etc... That seems quite verbose IMHO.

j8r:
Voila, docs in the `openrc-run.8` added with `command_user`, `command_group`,  `command_stdout`,  `command_stderr`.

williamh:
@j8r Not necessarily, if we want to move toward getting rid of the "command_" prefix, just don't add it to any new variables you add then we can transition away from it later.

j8r:
@williamh Really? Nice! I didn't know what direction you would like to take - adding `command_` to each argument directly related to the command itself. We can get rid of all `command_` and keep accepting both `user` and `command_user` for the moment.

williamh:
@j8r I added most of these, with slightly different names, but I didn't add group. The reason for that is I'm considering another path. Currently command_user can be either "user" or "user:group", but I'm thinking about adding a third form, ":group" which would not change the user but would change the group.

j8r:
@williamh Why not, but I personally think that having one information per variable, in the form of *key=value* is clearer -  `command_user=user` and `command_group=group` compared to `command_user=user` and `command_user=:group` or `command_user=user:group`.

j8r:
I see your commits - I would like to add also that if we keep the same logic of `command_user`, `command_args` etc, we should also have `log_error`, `log_output`. In the future who knows there could have new log related variables that will be able to follow the `log_` schema.

PR #189, opened 2017-12-09
Gracefully handle unreadable /sys/class/net/dev/ nodes

mgreter:
Fixes https://bugs.gentoo.org/629228

The fix mentioned in the gentoo bug report is not enough.
On the first failing try, the variable `x` would be assigned the
value `down` from the second check. On a second failing try,
the conditional `[ $x -eq 1 ]` will print an error:
`[: down: integer expression expected`

### DIFF
--- a/init.d/net-online.in
+++ b/init.d/net-online.in
@@ -44,10 +44,10 @@ start ()
 	ifcount=0
  	for dev in ${interfaces}; do
 		: $((ifcount += 1))
-		read x < /sys/class/net/$dev/carrier
-		[ $x -eq 1 ] && : $((carriers += 1))
-		read x < /sys/class/net/$dev/operstate
-		[ "$x" = up ] && : $((configured += 1))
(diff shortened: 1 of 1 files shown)


williamh:
Hi, can you please take a look at the Gentoo bug? I added another modification there that I would like to use if it works.

The change I am suggesting will check for existance of the sys files before it reads them instead of suppressing the error messages.

Please respond on the bug and let me know if that works.

mgreter:
Hi, unfortunately all file tests return true, even though you will get an error on read.

- https://unix.stackexchange.com/questions/252002/help-testing-special-file-in-sys-class-net

> I have also used the tests -b, -c, -f all of which fail to be able to distinguish between the 'readable' state of this file in /sys/class/net/wwan1/ - the 'bad' file, and /sys/class/net/wwan0/ - the 'good' file.

Will add this comment to the gentoo bug too once I'm home ...

PR #205, opened 2018-02-06
Fix hardcoded sbin paths, improve file-access error messages

AgentD:
A little bit of background: I was fiddling around with OpenRC on custom, cross compiled system that
doesn't have a /sbin directory. Setting SBINDIR=/bin compiled fine and installed everything to /bin.

However, I discovered that there were a few hard coded "/sbin/<foo>" paths in the code. This is what
the first commit tries to fix.

The reason I had to dig into the source code in the first place was because "exec: no such file or directory" wasn't exactly helpful in trying to find the cause of the problem (more precisely, such
an error message is utterly meaningless). I could have found the problem much faster if the program had simply told me that that it was trying to access something in /sbin. This is what the second
commit tries to fix.

### DIFF
--- a/src/librc/rc.h.in
+++ b/src/librc/rc.h.in
@@ -23,6 +23,7 @@ extern "C" {
 #endif
 
 #define RC_PREFIX "@PREFIX@"
+#define RC_SBINDIR		"@SBINDIR@"
 #define RC_SYSCONFDIR		"@SYSCONFDIR@"
 #define RC_LIBDIR               "@PREFIX@/@LIB@/rc"
 #define RC_LIBEXECDIR           "@LIBEXECDIR@"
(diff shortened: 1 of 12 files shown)


williamh:
@AgentV: I think the better way to fix the hard coded sbin path is to remove the path from the executable name and use the exec functions that check the path (execlp, execlpe, execvp or execvpe) as appropriate. I believe all of the cases are now covered in current master. Please update this p/r based on current master.

Thanks,

William

N-R-K:
Most of the hardcoded "/sbin" are no more in current master. But in openrc-init, there's still this:

https://github.com/OpenRC/openrc/blob/028eadec44f6321652fede7fb6eae77b0131bd90/src/openrc-init/openrc-init.c#L47

Maybe we should pass the user given `--sbindir=` as a macro so that it can be added at the front of the `path_default`?

PR #206, opened 2018-02-07
Fix vulnerabilities switch from insecure strcpy, strcat, sprintf to strncpy, strncat, snprintf

wltjr:
Fix vulnerabilities switch from insecure strcpy, strcat, sprintf to strncpy, strncat, snprintf.

### DIFF
--- a/src/librc/librc-daemon.c
+++ b/src/librc/librc-daemon.c
@@ -507,6 +507,7 @@ rc_service_daemons_crashed(const char *service)
 	char *exec = NULL;
 	char *name = NULL;
 	char *pidfile = NULL;
+	size_t pidfile_len = 0;
 	pid_t pid = 0;
 	RC_PIDLIST *pids;
 	RC_PID *p1;
@@ -517,6 +518,7 @@ rc_service_daemons_crashed(const char *service)
 	RC_STRINGLIST *list = NULL;
 	RC_STRING *s;
 	size_t i;
+	size_t ch_root_len = 0;
 	char *ch_root;
 	char *spidfile;
 
@@ -566,9 +568,11 @@ rc_service_daemons_crashed(const char *service)
 		ch_root = rc_service_value_get(basename_c(service), "chroot");
 		spidfile = pidfile;
 		if (ch_root && pidfile) {
-			spidfile = xmalloc(strlen(ch_root) + strlen(pidfile) + 1);
-			strcpy(spidfile, ch_root);
-			strcat(spidfile, pidfile);
+			ch_root_len = strlen(ch_root);
+			pidfile_len = strlen(pidfile);
+			spidfile = xmalloc(ch_root_len + pidfile_len + 1);
+			strncpy(spidfile, ch_root, ch_root_len + 1);
+			strncat(spidfile, pidfile, pidfile_len + 1);
 			free(pidfile);
 			pidfile = spidfile;
 		}
--- a/src/librc/librc-misc.c
+++ b/src/librc/librc-misc.c
@@ -351,7 +351,11 @@ static RC_STRINGLIST * rc_config_directory(RC_STRINGLIST *config)
 			TAILQ_FOREACH(fname, rc_conf_d_files, entries) {
 				if (! fname->value)
 					continue;
-				sprintf(path, "%s/%s", RC_CONF_D, fname->value);
+				snprintf(path,
+					strlen(RC_CONF_D) + strlen(fname->value) + 2,
+					"%s/%s",
+					RC_CONF_D,
+					fname->value);
 				rc_conf_d_list = rc_config_list(path);
 				TAILQ_FOREACH(line, rc_conf_d_list, entries)
 					if (line->value)
--- a/src/rc/kill_all.c
+++ b/src/rc/kill_all.c
@@ -170,7 +170,7 @@ static int signal_processes(int sig, RC_STRINGLIST *omits, bool dryrun)
 			continue;
 
 		/* Is this a process we have been requested to omit? */
-		sprintf(buf, "%d", pid);
+		snprintf(buf, sizeof(buf), "%d", pid);
 		if (rc_stringlist_find(omits, buf))
 			continue;
 
--- a/src/rc/supervise-daemon.c
+++ b/src/rc/supervise-daemon.c
@@ -237,8 +237,9 @@ static void child_process(char *exec, char **argv)
 	char **c;
 	char cmdline[PATH_MAX];
 	time_t start_time;
-	char start_count_string[20];
-	char start_time_string[20];
+	size_t start_len = 20;
+	char start_count_string[start_len];
+	char start_time_string[start_len];
 
 #ifdef HAVE_PAM
 	pam_handle_t *pamh = NULL;
@@ -252,9 +253,9 @@ static void child_process(char *exec, char **argv)
 		start_time = time(NULL);
 		from_time_t(start_time_string, start_time);
 		rc_service_value_set(svcname, "start_time", start_time_string);
-		sprintf(start_count_string, "%i", respawn_count);
+		snprintf(start_count_string, start_len, "%i", respawn_count);
 		rc_service_value_set(svcname, "start_count", start_count_string);
-		sprintf(start_count_string, "%d", getpid());
+		snprintf(start_count_string, start_len, "%d", getpid());
 		rc_service_value_set(svcname, "child_pid", start_count_string);
 	}
 
@@ -399,8 +400,8 @@ static void child_process(char *exec, char **argv)
 	*cmdline = '\0';
 	c = argv;
 	while (c && *c) {
-		strcat(cmdline, *c);
-		strcat(cmdline, " ");
+		strncat(cmdline, *c, PATH_MAX-strlen(cmdline)-1);
+		strncat(cmdline, " ", PATH_MAX-strlen(cmdline)-1);
 		c++;
 	}
 	syslog(LOG_INFO, "Child command line: %s", cmdline);
@@ -596,8 +597,8 @@ int main(int argc, char **argv)
 	*cmdline = '\0';
 	c = argv;
 	while (c && *c) {
-		strcat(cmdline, *c);
-		strcat(cmdline, " ");
+		strncat(cmdline, *c, PATH_MAX-strlen(cmdline)-1);
+		strncat(cmdline, " ", PATH_MAX-strlen(cmdline)-1);
 		c++;
 	}
 	if (svcname) {
@@ -761,7 +762,7 @@ int main(int argc, char **argv)
 		child_argv = xmalloc((child_argc + 1) * sizeof(char *));
 		memset(child_argv, 0, (child_argc + 1) * sizeof(char *));
 		for (x = 0; x < child_argc; x++) {
-			sprintf(name, "argv_%d", x);
+			snprintf(name, PATH_MAX+1, "argv_%d", x);
 			str = rc_service_value_get(svcname, name);
 			child_argv[x] = str;
 		}
@@ -852,11 +853,11 @@ int main(int argc, char **argv)
 		fclose(fp);
 
 		rc_service_value_set(svcname, "pidfile", pidfile);
-		sprintf(name, "%i", respawn_delay);
+		snprintf(name, PATH_MAX+1, "%i", respawn_delay);
 		rc_service_value_set(svcname, "respawn_delay", name);
-		sprintf(name, "%i", respawn_max);
+		snprintf(name, PATH_MAX+1, "%i", respawn_max);
 		rc_service_value_set(svcname, "respawn_max", name);
-		sprintf(name, "%i", respawn_period);
+		snprintf(name, PATH_MAX+1, "%i", respawn_period);
 		rc_service_value_set(svcname, "respawn_period", name);
 		child_pid = fork();
 		if (child_pid == -1)
@@ -875,12 +876,12 @@ int main(int argc, char **argv)
 			c = argv;
 			x = 0;
 			while (c && *c) {
-				snprintf(name, sizeof(name), "argv_%-d",x);
+				snprintf(name, PATH_MAX+1, "argv_%-d",x);
 				rc_service_value_set(svcname, name, *c);
 				x++;
 				c++;
 			}
-			sprintf(name, "%d", x);
+			snprintf(name, PATH_MAX+1, "%d", x);
 				rc_service_value_set(svcname, "argc", name);
 			rc_service_value_set(svcname, "exec", exec);
 			supervisor(exec, argv);


vapier:
this seems to be a blind set of changes w/out actually understanding the code.  the fact that code uses strcpy/strcat doesn't mean the code is "insecure".  the librc-daemon.c changes for example are not necessary and the existing code is fine (and if you look closely, the math is off by one).  the librc-misc.c changes are def incorrect because you're not passing the correct length for the 2nd arg.  i gave up reading the code after that point.

please re-evaluate every proposed change and see what is actually necessary.

wltjr:
Not really, I ran a build through [SonarCloud/SonarQube static analysis](https://sonarcloud.io/dashboard?id=openrc). I made a similar set of changes to one of my own projects [asspr](https://sonarcloud.io/dashboard?id=asspr) and will be doing others. I was using Coverity, still have some projects setup there before migration to SonarCloud. Asspr had no issues under Coverity, but showed other issues under Sonar. I am sure other applications will be the same.

I had [Travis setup](https://github.com/Obsidian-StudiosInc/openrc/commit/a2acc2831e90beda29ba61322ff590c2d84f2cd0#diff-354f30a63fb0907d4ad57269548329e3) to send builds to Coverity. It identified some 55 issues, but I was denied ability to see scans since I am not part of the OpenRC project. Also another, pinentry, unrelated. Thus ended up on Sonar, etc.

The wording "insecure" comes from [Sonar S1081](https://www.sonarsource.com/products/codeanalyzers/sonarcfamilyforcpp/rules-c.html#RSPEC-1081). The changes are not so much blind, and if I am of I can correct that. Though if the code was safe before, any limits should not be reached. As it would have overrun before without such.

I am happy to correct as needed.

wltjr:
Think I reversed a few null paddings, +1 vs -1 I will correct and re-push.

wltjr:
FYI, some results from Coverity scan. I can send over .travis.yml for general test builds and upload to Coverity if you want to run a scan and see the results. I was denied ability to see detailed scan results, 55 defects. Sonar seems to report more than Coverity, not sure if it catches the same stuff or not.

shot-2018-02-05_13-37-31

vapier:
"i ran this static analyzer and it complained" isn't a valid reason to make a change.  if the checker is basically doing `grep strcat && die`, then the checker is garbage and we don't care.  you have to actually look at the code and understand it.

also, just because some checker used scary/nonsense language isn't a reason to copy & paste it into your commit messages.  simply using strcat is not "insecure".

as it stands, these changes are either unnecessary, or plain wrong.  i looked at the next file in your changes (kill_all.c) and that change is absolutely wrong (and introduces an uninitialized read of a stack buffer).  after that, the attempts to use strncat in supervise-daemon.c are wrong and a pretty common mistake -- the length to strncat is how much of the buffer is free, not the size of the entire buffer.  i gave up at this point again on reading more.

making changes w/out actually understanding things is a recipe for disaster, and this PR adds way more bugs that it might have tried to legitimately fix.

what would be interesting is adding support for something like asprintf which is way harder to screw up and doesn't rely on fixed size stack buffers.

wltjr:
Multiple analyzers that many projects rely on... I think you are making wide assumptions what these advanced checkers are doing. Its hardly as trivial as a grep... If you think Coverity and SonarQube scans are garbage. Then you would be at odds with quite many, a growing number. I suggest you look at the [issues Sonar found](https://sonarcloud.io/project/issues?id=openrc&resolved=false). It is also open source, so you can see the [code is hardly trivial](https://github.com/SonarSource/sonarqube).

I care less about the language, though its common, and I provided links. Which one has several other links to why coding like that should be avoided. Sonar is not making suggestions with facts to support.

"   MITRE, CWE-120 - Buffer Copy without Checking Size of Input ('Classic Buffer Overflow')
    MITRE, CWE-676 - Use of Potentially Dangerous Function
    SANS Top 25 - Risky Resource Management
    CERT, STR07-C. - Use the bounds-checking interfaces for string manipulation"
https://www.sonarsource.com/products/codeanalyzers/sonarcfamilyforcpp/rules-c.html#RSPEC-1081

I find it odd that Sonar would not catch an uninitialized read of a stack buffer. In fact these scanners catch such things. That maybe something Coverity catches that Sonar does not. I did make changes per initial comments and I may have made things worse by second guessing. Again I can correct as needed, just as I believe the PR can be modified as part of any merge.

I have been working on several C applications for the last couple years. Sure I make some mistakes, but I am not clueless when it comes to reading code. Mostly been going around fixing others code with help of things like Coverity and Sonar.
https://scan.coverity.com/projects/obsidian-studiosinc-jem
https://scan.coverity.com/projects/obsidian-studiosinc-ecrire
https://scan.coverity.com/projects/obsidian-studiosinc-entrance

Just what is remaining. Coverity removed Clipboard I had for months, and others pinentry and openrc. They have some odd policies on who can scan what, etc. Sonar is less restrictive. Really amazed at the issues in various code bases that seem to be unknown.

My changes are going through Travis and Sonar. They cannot be that unsafe of introduce such bugs you describe or they would or should be caught by Sonar. I will go a step further, package, and test my changes on real systems. Then it will be triple covered. Sorry for trying to make things better. I can just further things on my own and not submit PRs. Clearly many issues that people are not aware of and may discount or not care about but are real and exist. Likely many more projects with issues as well.

I agree on asprintf and use it in jem. Though I believe that is non-standard and a GNU extension or something. Which may or may not be issues for openrc. Depending on what all its targeting.

wltjr:
Seems my code in kill_all.c maybe off. What is interesting is there is another existing line doing the same thing. I do not see [path](https://github.com/OpenRC/openrc/blob/master/src/rc/kill_all.c#L92) being initialized at all before a similar [strlen](https://github.com/OpenRC/openrc/blob/master/src/rc/kill_all.c#L101). Thus if one is wrong, both are wrong. No?

wltjr:
Whoops, difference between size of and strlen, my bad...

vapier:
i'm just going to punt this.  these changes still make no actual improvements and if anything, make the code worse to read & understand.  your latest PR still *introduces* real bugs like buffer overflows (e.g. your supervise-daemon.c change to inline PATH_MAX+1 instead of using sizeof(name)).

i'll reiterate what i said earlier: using funcs (e.g. strcat) does not mean the code is insecure.  in all of the instances you highlighted, it's either already written correctly (e.g. librc-daemon.c), or it's extremely unlikely to be a problem (e.g. kill_all.c's buf is PATH_MAX+1 bytes, and we're only writing a 31-bit signed int, and if PATH_MAX is less than 20, everything is already going to be completely broken).

there are legitimate improvements that could be made in openrc wrt string handling (such as converting many things to asprintf as i suggested earlier which would subsume the majority if not all existing strcat/sprintf users), but none of those improvements would look like the cargo cult changes you're blindly making here.

wltjr:
No worries I have forked and will fix and further on my own, for my own needs. These were really minor issues and mostly a feel for how things would be with future submissions.

Extremely unlikely to be an issue is no reason to keep using functions most every documentation, definitely any security related document on best practices, secure coding, etc, will say do not use that stuff. I think its unlikely people can exploit Meltdown or Spectre, so those should not be addressed? That logic does not make sense.

Moving to asprintf will mean dropping POSIX compliance if that matters. It is not standard C and is GNU specific. That is a choice I made for jem, despite clams to be Posix compliant, I may need to move away from that. I did not start out using asprintf it was recommended by another.

How many security audits has OpenRC codebase gone through? Any? Seems a ton of issues in the codebase to be addressed. These were minor and dismissed. Not sure what that says about the rest. Maybe I should be like the rest and switch to systemd? These types of issues will not bring more people to running OpenRC. If that matters.

Anyway thanks for feedback. I will not bother with future submissions.

vapier:
you seem to be confusing actual security bugs and real fixes with blind changes that make absolutely no difference to security.  if you have a *real* bug to point out, feel free, but none of these are bugs.

if you want to participate in the bs security space replete with scare tactics, snake oil, and meaningless checkboxes, feel free.  just don't mistake it for actual security.

wrt asprintf, it's trivial to add a fallback local implementation built on top of two snprintf & one malloc call.  i'm not suggesting we throw away portability.

vapier:
plus, considering this PR was supposed to be "simple" but somehow managed to replace working/correct code with at least 4 real bugs, it's not exactly a glowing review of the fork work.  i might hold off on selling it as a "improved" code base security wise.  or not ... anyone who blindly trusts that won't know the difference i guess.

still don't confuse "nonsense goes nowhere" with "improvements aren't welcome".

wltjr:
I did not create Sonar, it is not my wording... On a SonarQube scan such things show up as "vulnerabilities". I guess Sans, Cert, Mitre, etc are not in the security business. That is their wording not mine. None the less I see no reason not to avoid using such. Only in some rare cases for maybe embedded devices where efficiency matters.

FYI a similar thing came up over the summer for [Tizen/EFL use of strcpy](https://motherboard.vice.com/en_us/article/xy9p7n/samsung-tizen-operating-system-bugs-vulnerabilities). The first things I read long ago about coding in C said to avoid such things. If anything the secure s_ versions should be used when available if not using the fixed sized buffer alternatives.

Either way no worries. All projects have their own understanding of the code and exploits. I just prefer to make it clear to all. Even dump scanners.

wltjr:
Can you be any ruder? If these are so minor you could have merged with your fixes and commented on the fixes. That would require less effort than talking or going back and forth. Or simply pointing out the correct way over the wrong way. Rather than just being negative all around about the contribution. Correct or not, I tend to thank people for at least trying to help...

These are not even bugs. I haven't even started on what Sonar reports as bugs, or blockers, etc. There are tons of issues.... I would be more looking into such. I am not selling anything as improved or secure. I said I would do it for my own purposes.... Though I will be using more than my own opinion on such things, like third party tools to scan code, etc. I can only imagine what showed on Coverity that may not be under Sonar. Does not seem like you care about such. That to me says more about the current codebase than anything I do with it going forward. For my own needs.

No worries

vapier:
> could have merged with your fixes and commented on the fixes

you're assuming the changes you proposed were even desirable from the start.  they weren't -- as i've said numerous times, you aren't actually "fixing" anything other than making a brain dead security scan go away.  it also assumes we want to merge known broken code before applying fixes on top ... that would needlessly clutter the git history with known bad code.

i suggested a sane route to addressing the issues you pointed out.  whether you want to implement it properly or just ignore or bandaid is your prerogative.

if you consider frankness to be the same as rudeness, then i guess that's that.

wltjr:
I said numerous times I would fix and pushed fixes, which I did a couple times. You could also fix as part of any commit/merge. Which any issue you mention with history, broken code, etc would have been negated with corrections. Simple squash to make for a clean history etc. None really valid points. I think the real issue is your first point, the changes were not desired from the start. You do not deem them necessary. That I can see as subjective. If things are not done the way the project developers want, or seen as issues, they are deemed otherwise.

What your saying is write code the way I want it written to be accepted. You suggested asprintf, thus no other solution not using that would be accepted. Nor do I see you taking the time to address. Likely just sweep under the rug and disregard. If I failed at simple replacements. I am not going to attempt a universal asprintf for platforms that have it and those that do not. Not sure that is an approach I would take myself. I am not here to code per others wishes. Not unless paid, and even then...

What I consider rude is saying something is wrong, without showing a correct way. Being overly negative towards someone simply trying to help and improve things. While in your opinion they may not be improvements. Others may feel differently. Plus your quotes of things I never said... Saying checkers are garbage, and numerous disparaging comments in each reply.

No worries, plenty of places to contribute where contributions are handled differently. Really no need for any of this back and forth. Its all more effort than to fix in code...

williamh:
@wltjr I just saw this myself, and I hope you will change your mind about future submissions. Forking and doing things on your own never helps the community; it always is not good.

wltjr:
@williamh I agree, but I rather not be seen as argumentative debating on different paths forwards. There is more than one way to skin a cat. Peoples coding styles and techniques differ. I accept that, to each their own. I am not seeking to upset anyone or waste anyone's time in lengthy discussions. Why I left is short from the start.

Sorry for the noise... I will continue my efforts. If you all want to cherry pick that is fine. Otherwise rather not bother with my garbage code and submissions... I will keep my bugs to myself.

wltjr:
Must say not sure how I failed so bad here. I made [similar changes](https://github.com/Obsidian-StudiosInc/asspr/commit/242b7af5eaee65ce9ba1690c14800afd6f29ea5b) to a project of mine, asspr. From actual usage after the changes. I am not seeing any errors or issues. If I did not properly covert from strcat to strncat I think I would have issues.... Making me second guess myself and may go introduce bugs in the process... Still trying to find bugs in asspr from my changes. Surely possible but thus far no issues.

vapier:
i never said "changes are not accepted".  i did say your changes are not only bad & broken, they're unnecessary, and there are better ways to fix both (i.e. asprintf).  i did point out the actual errors in your code (with the fixes being self-explanatory), as well as a sane way forward.  if you don't want to properly fix code, that's your choice.

assuming "it didn't crash therefore the code is correct" makes no sense.  that's simply not how memory corruption works, and lack of edge cases to exercise the faulty paths does not translate to correctness.

wltjr:
I did fix and re-pushed a couple times. Most should be better or fixed, short of the switch to asprintf. All of which can be done as part of a merge, it can be edited. IMHO less time to fix in code than type here.

I am relying on a several things, not just runtime tests. If it takes edge cases to trigger than it is not likely in most cases. In such edge cases given the code now, you may have other such issues.

vapier:
once again, changing strcat to strncat by itself fixes *nothing*.  we already made sure the buffers are the correct size.  all you've done is made the code harder to understand in order to blindly please garbage static checkers.  if you don't understand that, then clearly trying to discuss the matter isn't worth the time.

your argument wrt edge cases makes no sense.  if you're assuming they don't get hit after your changes, then what's the point of making the changes in the first place if the edge cases can't happen ?  that's not how real security works.

wltjr:
When you call scanners garbage, it is hard to put weight to all the rest. I guess you do not use repoman either in Gentoo right? Or CI, etc?

I never said I was doing a security audit or fixing the code for secure purposes. I simply was passing along Sonar's text. Why you cannot seem to understand that simple point is beyond me. All around I am not here to argue with anyone... I rather spend my time doing others things like fixing code or coding vs talking about it.

wltjr:
You keep discounting others just the same. I am not the one making negative remarks. I never asked for your opinion on scanners.

My understanding of not using strcpy/strcat,sprintf preceeds scanners. It was like C101 stuff for any modern coding. It simply should not be used unless like I said for performance reasons. None of those make code any harder to read with setting a length. You have yet to make any good arguments as to why it was coded that way in the first place, or why they are even used.

I am not blindly following scanners results, and thank you for such negative remarks again. Like I said and showed. I am making such changes to other projects as well. I will not bother with future contributions.

But given what I have seen in some of the scanners. To say the code is bug free is only because you have not run into those yet. It is not like openrc is widely used in a systemd world.... Most have no clue about it...

williamh:
@wltjr I would appreciate it if you follow the code and offer contributions. Just because one isn't taken doesn't mean none will be. You are as welcome on this project as anyone is.

wltjr:
@williamh I am not as concerned about a single contribution as I am the process. I am not sure contributing is good use of my time given the experience.

vapier:
i'm saying the code is bug free because i can read and understand what it's doing.  the fact it's using strcat is irrelevant.  it's hard to say you understand any of this based on your code submissions.

wltjr:
Thanks for the insult... You have yet to make a compelling argument as to why it was coded that way in the first place. I have provided several external links as to why it should not be used. Despite YOUR opinion, I will go with factual documentation and best practices.

vapier:
no, you've linked to pages clearly without having read or understood them.  for example, your first link:
https://www.sonarsource.com/products/codeanalyzers/sonarcfamilyforcpp/rules-c.html#RSPEC-1081
>When using legacy C functions such as strcpy, it's up to the developer to make sure the size of the buffer to be written to is large enough to avoid buffer overruns.

guess what ?  we did make sure the size is large enough.

wltjr:
Then with bother with asprintf.... Guess what your defeating your own argument... Enjoy your code. I can see what other distros went with Systemd...

williamh:
@wltjr You are welcome to participate in this project, but snippy comments like the last one you posted on this pull request are not welcome. If you have an issue with Me, Mike or anyone else, take it private.

Thanks,

William

wltjr:
@williamh given the amount of condescending remarks regarding static analysers, typing quotes of things I did not say, and then flat out making comments as to ones understanding. I guess those are all acceptable but my response after receiving several negative derogatory remarks. I think you should all should conduct yourself as your expecting others to....

Maybe lead by example rather than do as I say, while insulting others... Despite several valid points being dismissed without any facts to support. Beyond a single persons opinion... I am hardly in the wrong here.

But thanks for giving me further reason to avoid the project and fork for any of my needs. Once again I can see why others choose to avoid OpenRC. I seriously questioned that for many years. Now I have some understanding as to why. You can take that how ever you want. It is considerably more polite than any comments lobbed my direction. Without a single bit of thanks.

No matter if someone's contributions are correct or not, valid or not, I always THANK people for at least trying.... From the first reply, there has never been a single bit of politeness or respect. If anything rudeness and disrespectful replies.

I will not comment further or open issues, PRs, etc. Please do not reference my name in any tickets or new ones. I will be unsubscribing and not receiving future messages. Thanks!

PR #208, opened 2018-02-10
helpers.h: xasprintf function, replace ... with ap_list ap, init vars

wltjr:
Minor changes, feel free to drop var init stuff if you want. The other replacement is per the following.

MISRA C:2004, 16.1 - Functions shall not be defined with a variable number of arguments.
CERT, DCL50-CPP. - Do not define a C-style variadic function

### DIFF
--- a/src/includes/helpers.h
+++ b/src/includes/helpers.h
@@ -132,12 +132,11 @@ _unused static bool existss(const char *pathname)
  * functions to handle memory allocation.
  * this function was originally written by Mike Frysinger.
  */
-_unused static int xasprintf(char **strp, const char *fmt, ...)
+_unused static int xasprintf(char **strp, const char *fmt, va_list ap)
 {
-	va_list ap;
-	int len;
-	int memlen;
-	char *ret;
+	int len = 0;
+	int memlen = 0;
+	char *ret = NULL;
 
 	/*
 	 * Start with a buffer size that should cover the vast majority of uses


vapier:
please actually test/understand code before making suggestions.  this code change is unnecessary, makes no sense, as is objectively wrong.  you're confusing asprintf and vasprintf.

williamh:
@wltjr The concern about this p/r is the way we are going to call xasprintf(). It will be called like this:

```
xasprintf(buffer, format, arg1, arg2, ...);
```

So think of it as being like sprintf() but with allocation.
So, I'm not sure why  your scanner is objecting to it.

wltjr:
I understand the intentioned usage. It is the following that says it should not be done. I am not sure what the alternative replacement is at this time.

MISRA C:2004, 16.1 - Functions shall not be defined with a variable number of arguments.
MISRA C++:2008, 8-4-1 - Functions shall not be defined using the ellipsis notation.
CERT, DCL50-CPP. - Do not define a C-style variadic function

Per
https://www.sonarsource.com/products/codeanalyzers/sonarcfamilyforcpp/rules-c.html#RSPEC-923

It is also in MISRA 2012
https://en.wikipedia.org/wiki/MISRA_C

Some further documents say

"There are a lot of potential problems with this feature. Users shall not write additional functions that use a variable number of arguments. This precludes the use of use of stdarg.h, va_arg, va_start and va_end. "
http://clinker.soltel.es/sonar/rules_configuration/index/13#rule_1826

williamh:
@wltjr ok, there is a lot of info there, but I need a way to achieve the same functionality if I am going to remove the variable argument list. so please let me know once you know what the replacement should be.

vapier:
openrc does not aim to be MISRA compliant, nor do i think there is any usefulness for the project to do so.  its goals are specifically for embedded systems where reliability is paramount and assumes/imposes limitations that make no sense for openrc (such as focusing on limiting memory heap usage).

wltjr:
I can appreciate not wanting to be MISRA compliant. Though in most cases I do not see harm in much of the recommendations. That said, some recommendations like the variable argument and related preprocessor handling do not seem to have examples of alternatives. Where most other things do. I cannot find replacements for such. Which makes compliance that much harder, if possible in some cases.

Regarding embedded I think there is potential for usage in embed. Systemd is used in some embedded systems. [Tizen](https://wiki.tizen.org/Porting_Guide/Tizen_Bootup_Overview) uses it for some devices, not sure which all. Not sure where mobile/watches fall wrt to embedded.

PR #209, opened 2018-02-12
check whether /sys/fs/cgroup is a mountpoint

brauner:
The current check only tries to detect whether /sys/fs/cgroup exists and
whether it is writable or not. But when the init system doesn't mount
cgroups then /sys/fs/cgroup will just be an empty directory. When paired
with unprivileged containers that mount sysfs this will cause misleading
errors to be printed since /sys/fs/cgroup will be owned by user
nobody:nogroup in this case. Independent of this specific problem this
check will also be misleading when the /sys/fs/cgroup exists and is in
fact writable by the init system but isn't actually a mountpoint.


asarubbo:
the suggested fix silence the warning for me.

brauner:
Thanks, @williamh!

PR #226, opened 2018-05-08
librc: fix realpath() return value check

edanbe:
The return value of `realpath()` is checked by dereference, although it returns `NULL` if there is an error. This CL checks for a `NULL` return value instead of dereferencing the pointer.


philhofer:
Fixes #243

PR #229, opened 2018-06-21
hwclock: exit early if no rtc device is found

clandmeter:
When using a rpi without rtc I always see this error:

hwclock: Cannot access the Hardware Clock via any known method.
hwclock: Use the --verbose option to see the details of our search for an access method.

### DIFF
--- a/init.d/hwclock.in
+++ b/init.d/hwclock.in
@@ -97,7 +97,6 @@ start()
 		return 0
 	fi
 
-	ebegin "Setting system clock using the hardware clock [$utc]"
 	if [ -e /proc/modules ]; then
 		if ! rtc_exists; then
 			for x in rtc-cmos rtc genrtc; do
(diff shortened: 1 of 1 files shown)


clandmeter:
ah it seems i need to add swclock to provide an alternative clock or else it will depend on hwclock.

williamh:
If you don't need a clock provider at all, you should be able to delete both of them from the runlevels unless you have a service outside of OpenRC that has a "need clock" in its dependencies.

Try running these commands:

```
# rc-update del hwclock boot
# rc-update del swclock boot
```

vapier:
in general, making the service silent by default like this is bad practice imo.  in the case where a user has an RTC and expects it to work, but perhaps by some configuration setting it's not accessible, then we'd want those failures to be visible so they can take action.

if you don't have a hwclock, then it shouldn't be used/started at all.

PR #236, opened 2018-07-25
librc-misc: remove leftover dead variable/assignment

hhoffstaette:
I was looking at some of the warnings emitted when building with -Wall and this one was easy
enough. It's likely a leftover from the previous commit b2f5531194 ("librc-misc: convert snprintf calls to xasprintf".


hhoffstaette:
As an aside, the only other warnings I see are all in librc.c which doesn't seem to have gotten any xasprintf love yet. Any hard rules to follow when converting, other than changing the stack array to a pointer and freeing when the buffer goes out of scope?

williamh:
Sorry I missed this pr.

If you are still interested in helping convert to xasprintf, you are right about the rules for it. Just make sure to free() the buffer when it goes out of scope.

PR #246, opened 2018-10-12
add a basic TravisCI setup

austin987:
This would let you automate 'make test' checks, including when PRs are sent.

For an example, see https://travis-ci.org/austin987/openrc


.travis.yml
```diff
@@ -0,0 +1,7 @@
+language: c
```
vapier: can you add to the top of this file:

```
# Travis build integration.
# https://docs.travis-ci.com/
```

.travis.yml
```diff
@@ -0,0 +1,7 @@
...
+os:
+  - linux
+script:
```
vapier: this can easily grow. rather than do this, how about we create a test/travis.sh script to handle this logic and this line simply reads:

```
script: ./test/travis.sh
```

.travis.yml
austin987: Ok.

.travis.yml
austin987: Done, please review.

test/travis.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
```
vapier: please add a copyright/license block to this file

test/travis.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
```
vapier: this works, but i expect this script will grow, and using bash rather than POSIX sh is prob easier for people to work under.  so i'd suggest using /bin/bash now.

test/travis.sh
```diff
@@ -0,0 +1,9 @@
...
+
+set -e
+set -u
+set -x
+
+# These are steps to run on TravisCI under a containerized Ubuntu system
+make
```
vapier: should we detect the number of cpus the system has using getconf and add that to -j here ?

austin987:
I also went a bit further and made whitespace/style stuff fatal (if $FATAL_CHECKS is not empty). Would be simple to make it fail always, but I assumed there was a reason it didn't already.

austin987:
So I tried adding @mrueg's changes, but looks like openrc doesn't build with musl-gcc:
```
musl-gcc -fPIC -DPIC  -I../includes -D_DEFAULT_SOURCE -D_POSIX_C_SOURCE=200809L  -O2 -g -std=c99 -Wall -Wextra -Wimplicit -Wshadow -Wformat=2 -Wmissing-prototypes -Wmissing-declarations -Wmissing-noreturn -Wmissing-format-attribute -Wnested-externs -Winline -Wwrite-strings -Wcast-align -Wcast-qual -Wpointer-arith -Wdeclaration-after-statement -Wsequence-point -Werror=implicit-function-declaration   -c libeinfo.c -o libeinfo.So
libeinfo.c: In function ‘colour_terminal’:
libeinfo.c:466:4: error: implicit declaration of function ‘strlcpy’ [-Werror=implicit-function-declaration]
    strlcpy(tmp, tgoto(bold, 0, 0), sizeof(tmp));
    ^
libeinfo.c:466:4: warning: nested extern declaration of ‘strlcpy’ [-Wnested-externs]
libeinfo.c:467:4: error: implicit declaration of function ‘strlcat’ [-Werror=implicit-function-declaration]
    strlcat(tmp, tgoto(_af, 0, c & 0x07), sizeof(tmp));
    ^
libeinfo.c:467:4: warning: nested extern declaration of ‘strlcat’ [-Wnested-externs]
libeinfo.c: In function ‘elogv’:
libeinfo.c:525:3: error: implicit declaration of function ‘vsyslog’ [-Werror=implicit-function-declaration]
   vsyslog(level, fmt, apc);
   ^
```

I've enabled gcc/clang, since those work. That passes:
https://travis-ci.org/austin987/openrc

@vapier @williamh is there anything else that's needed before this can go in?

test/travis.sh
```diff
@@ -0,0 +1,29 @@
...
+set -u
+set -x
+
+# These are steps to run on TravisCI under a containerized Ubuntu system
+# See $TOP/.travis.yml
+
+if command -v nproc; then
+    cpus="$(nproc)"
+else
+    cpus=1
```
vapier: here you can write:

 cpus=$(getconf _NPROCESSORS_CONF || echo 1)

test/travis.sh
```diff
@@ -0,0 +1,29 @@
...
+fi
+
+make -j"${cpus}"
```
vapier: add the `-O` flag so the parallel output is less confusing

test/travis.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# This make whitespace/etc. checks fatal
```
vapier: add a period to the end please

test/travis.sh
austin987: Cool, didn't know about that option, thanks!

austin987: That's because it's not everywhere yet. TravisCI fails with that:
```
+make -O -j2
make: invalid option -- 'O'
Usage: make [options] [target] ...
Options:
  -b, -m                      Ignored for compatibility.
  -B, --always-make           Unconditionally make all targets.
  -C DIRECTORY, --directory=DIRECTORY
                              Change to DIRECTORY before doing anything.
  -d                          Print lots of debugging information.
  --debug[=FLAGS]             Print various types of debugging information.
  -e, --environment-overrides
                              Environment variables override makefiles.
  -f FILE, --file=FILE, --makefile=FILE
                              Read FILE as a makefile.
  -h, --help                  Print this message and exit.
  -i, --ignore-errors         Ignore errors from commands.
  -I DIRECTORY, --include-dir=DIRECTORY
                              Search DIRECTORY for included makefiles.
  -j [N], --jobs[=N]          Allow N jobs at once; infinite jobs with no arg.
  -k, --keep-going            Keep going when some targets can't be made.
  -l [N], --load-average[=N], --max-load[=N]
                              Don't start multiple jobs unless load is below N.
  -L, --check-symlink-times   Use the latest mtime between symlinks and target.
  -n, --just-print, --dry-run, --recon
                              Don't actually run any commands; just print them.
  -o FILE, --old-file=FILE, --assume-old=FILE
                              Consider FILE to be very old and don't remake it.
  -p, --print-data-base       Print make's internal database.
  -q, --question              Run no commands; exit status says if up to date.
  -r, --no-builtin-rules      Disable the built-in implicit rules.
  -R, --no-builtin-variables  Disable the built-in variable settings.
  -s, --silent, --quiet       Don't echo commands.
  -S, --no-keep-going, --stop
                              Turns off -k.
  -t, --touch                 Touch targets instead of remaking them.
  -v, --version               Print the version number of make and exit.
  -w, --print-directory       Print the current directory.
  --no-print-directory        Turn off -w, even if it was turned on implicitly.
  -W FILE, --what-if=FILE, --new-file=FILE, --assume-new=FILE
                              Consider FILE to be infinitely new.
  --warn-undefined-variables  Warn when an undefined variable is referenced.
```

See https://travis-ci.org/austin987/openrc/jobs/456832938. If you feel strongly about it, we can remove `make -j`, but I think anyone looking at TravisCI logs is smart enough to figure out what directory a file is in with find.

test/travis.sh
austin987: Done here (and adjusted TravisCI comment as well).

test/travis.sh
austin987: Okay.

test/travis.sh
vapier: that's fine for now.  as you said, hopefully openrc output isn't so voluminous to be a problem currently.  once travis updates to include `-O` we can add it then.

PR #253, opened 2018-11-06
[runit integration] give a bit of time to start runsv

abchk1234:
Currently, we run sv start immediately after linking the service.
The runsv process may not be up at the moment, as a result of which
openrc will mark the service as stopped, even though it may be brought up
by runit at the next scan.

This is documented in the gentoo wiki:
https://wiki.gentoo.org/wiki/Runit#OpenRC.27s_runit_integration_feature

This PR adds a timeout so that correct process state can be reported.

Before:
~~~~
# rc-service netdata-runit start
 * Starting netdata-runit ...
fail: /run/openrc/sv/netdata: runsv not running
 * Failed to start netdata-runit                                          [ !! ]
 * ERROR: netdata-runit failed to start
~~~~

After:
~~~~
 rc-service netdata-runit start
 * Starting netdata-runit ...
fail: /run/openrc/sv/netdata: runsv not running
ok: run: /run/openrc/sv/netdata: (pid 15725) 1s                           [ ok ]
~~~~

### DIFF
--- a/sh/runit.sh
+++ b/sh/runit.sh
@@ -21,15 +21,15 @@ runit_start()
 	ebegin "Starting ${name:-$RC_SVCNAME}"
 	ln -snf "${service_path}" "${service_link}"
 	local i=0 retval=1
+	# it can take upto 5 seconds for runsv to start
 	while [ $i -lt 6 ] ; do
-		if sv -w 6 start "${service_link}" > /dev/null 2>&1; then
+		if sv start "${service_link}" > /dev/null 2>&1; then
 			retval=0
 			break
 		fi
 		sleep 1 && i=$(expr $i + 1)
 	done
-	sv start "${service_link}" > /dev/null 2>&1
-	eend $? "Failed to start ${name:-$RC_SVCNAME}"
+	eend $retval "Failed to start ${name:-$RC_SVCNAME}"
 }
 
 runit_stop()
--- a/sh/runit.sh
+++ b/sh/runit.sh
@@ -29,6 +29,10 @@ runit_start()
 		fi
 		sleep 1 && i=$(expr $i + 1)
 	done
+	if [ $retval -eq 1 ]; then
+		# clean up the link else sv will keep on trying
+		rm "${service_link}"
+	fi
 	eend $retval "Failed to start ${name:-$RC_SVCNAME}"
 }
 
--- a/sh/runit.sh
+++ b/sh/runit.sh
@@ -20,6 +20,14 @@ runit_start()
 	service_link="${RC_SVCDIR}/sv/${service_path##*/}"
 	ebegin "Starting ${name:-$RC_SVCNAME}"
 	ln -snf "${service_path}" "${service_link}"
+	local i=0 retval=1
+	while [ $i -lt 6 ] ; do
+		if sv -w 6 start "${service_link}" > /dev/null 2>&1; then
+			retval=0
+			break
+		fi
+		sleep 1 && i=$(expr $i + 1)
+	done
 	sv start "${service_link}" > /dev/null 2>&1
 	eend $? "Failed to start ${name:-$RC_SVCNAME}"
 }


williamh:
I would make the wait time a variable which can be configured.
Also, do we need the loop/sleep algorithm at all if we set the wait time on sv start?

aadityabagga:
@williamh 

Sleep loop is needed, as runit's directory scan can take upto [5 seconds](http://smarden.org/runit/faq.html#tell) to bring `runsv` up, _after_ which the service is started (wait time is for the service). Thats why the loop counter of **6**.

Thanks.

williamh:
One more question.
Can we kick the scanner so it brings runsv up right away instead of just doing a sleep loop? I'm asking all of these questions because  I would rather find a way to make things come up faster than add a sleep loop to every runit service.

udeved:
Afaik @williamh  it could also be avoided by putting some sleep in the runit run scripts.
But hang on, I'll ping our runit guy.
@konimex

konimex:
It won't work since runit marks the service as started as soon as the sleep command is run.

Also, the loop and `sv -w 6 start` is a bit useless IMO since `sv start` is basically the same as `sv -w 7 up` or `sv -w 7 start`.
Anyway, I'll take a look at the bug, `sv start` shouldn't return anything other than 0 if the service has been symlinked and started.

aadityabagga:
@williamh 

I appreciate your concern. Here is how it play out in practice. Following script is used for checking:
~~~~
success=0; failure=0
for i in {1..10}; do if rc-service netdata-runit start; then success=$((success + 1)); else failure=$((failure + 1)); fi; sleep 1s; rc-service netdata-runit stop; sleep 1s; done
echo "success: $success";echo "failure: $failure"
~~~~

**With patch (edited to echo slept time):**
~~~~
 * Starting netdata-runit ...
slept: 1
slept: 2
slept: 3                                                                                                                                                          [ ok ]
 * Stopping netdata-runit ...                                                                                                                                     [ ok ]
 * Starting netdata-runit ...                                                                                                                                     [ ok ]
 * Stopping netdata-runit ...                                                                                                                                     [ ok ]
 * Starting netdata-runit ...
slept: 1                                                                                                                                                          [ ok ]
 * Stopping netdata-runit ...                                                                                                                                     [ ok ]
 * Starting netdata-runit ...                                                                                                                                     [ ok ]
 * Stopping netdata-runit ...                                                                                                                                     [ ok ]
 * Starting netdata-runit ...
slept: 1                                                                                                                                                          [ ok ]
 * Stopping netdata-runit ...                                                                                                                                     [ ok ]
 * Starting netdata-runit ...                                                                                                                                     [ ok ]
 * Stopping netdata-runit ...                                                                                                                                     [ ok ]
 * Starting netdata-runit ...                                                                                                                                     [ ok ]
 * Stopping netdata-runit ...                                                                                                                                     [ ok ]
 * Starting netdata-runit ...                                                                                                                                     [ ok ]
 * Stopping netdata-runit ...                                                                                                                                     [ ok ]
 * Starting netdata-runit ...
slept: 1                                                                                                                                                          [ ok ]
 * Stopping netdata-runit ...                                                                                                                                     [ ok ]
 * Starting netdata-runit ...                                                                                                                                     [ ok ]
 * Stopping netdata-runit ...                                                                                                                                     [ ok ]
success: 10
failure: 0
~~~~

**Without patch:**
~~~~
 * Starting netdata-runit ...
fail: /run/openrc/sv/netdata: runsv not running
 * Failed to start netdata-runit                                                                                                                                  [ !! ]
 * ERROR: netdata-runit failed to start
 * WARNING: netdata-runit is already stopped
 * Starting netdata-runit ...
ok: run: /run/openrc/sv/netdata: (pid 19332) 1s                                                                                                                   [ ok ]
 * Stopping netdata-runit ...                                                                                                                                     [ ok ]
 * Starting netdata-runit ...
ok: run: /run/openrc/sv/netdata: (pid 19667) 1s                                                                                                                   [ ok ]
 * Stopping netdata-runit ...                                                                                                                                     [ ok ]
 * Starting netdata-runit ...
ok: run: /run/openrc/sv/netdata: (pid 19981) 1s                                                                                                                   [ ok ]
 * Stopping netdata-runit ...                                                                                                                                     [ ok ]
 * Starting netdata-runit ...
ok: run: /run/openrc/sv/netdata: (pid 20295) 1s                                                                                                                   [ ok ]
 * Stopping netdata-runit ...                                                                                                                                     [ ok ]
 * Starting netdata-runit ...
fail: /run/openrc/sv/netdata: runsv not running
 * Failed to start netdata-runit                                                                                                                                  [ !! ]
 * ERROR: netdata-runit failed to start
 * WARNING: netdata-runit is already stopped
 * Starting netdata-runit ...
ok: run: /run/openrc/sv/netdata: (pid 20615) 1s                                                                                                                   [ ok ]
 * Stopping netdata-runit ...                                                                                                                                     [ ok ]
 * Starting netdata-runit ...
ok: run: /run/openrc/sv/netdata: (pid 20953) 0s                                                                                                                   [ ok ]
 * Stopping netdata-runit ...                                                                                                                                     [ ok ]
 * Starting netdata-runit ...
fail: /run/openrc/sv/netdata: runsv not running
 * Failed to start netdata-runit                                                                                                                                  [ !! ]
 * ERROR: netdata-runit failed to start
 * WARNING: netdata-runit is already stopped
 * Starting netdata-runit ...
ok: run: /run/openrc/sv/netdata: (pid 21272) 1s                                                                                                                   [ ok ]
 * Stopping netdata-runit ...                                                                                                                                     [ ok ]
success: 7
failure: 3
~~~~

In theory, I agree that running `sv start ${service}` should start `runsv`, wait 7 seconds for the service to be up, and then report status and exit.

In practice, it seems not to function this way. Maybe we create a race condition by first linking the service, not waiting for it to come up, and running `sv start ${service}` ?

To be sure will need to look at the source as @konimex suggests. In practice the patch works well enough in my opinion. Adding the sleep allows OpenRC to report the correct status, the service takes the same time to be up regardless.

konimex:
>In practice, it seems not to function this way.

IIRC, it's because runit has detected that there's no error in the daemon (i.e. something that would block the daemon from running properly so runit has to restart it until timeout (by default 7 sec, or 6 sec since you specified `-w 6`)), it just returns 0.

So I'd say it's an intended behavior.

aadityabagga:
Can we get this merged? Fixes the incorrect service status and does not lead to long sleep loop in most cases. Check https://github.com/OpenRC/openrc/pull/253#issuecomment-436886189 for demo.

aadityabagga:
Hi all,

Have used a different approach. Instead of adding upto 6 second delay,  we dont report false failed.
Status can be checked via `rc-status`.

https://github.com/OpenRC/openrc/pull/322

PR #255, opened 2018-11-17
openrc-run: Fixing command_user in start-stop-daemon

concatime:
As mentionned in man openrc-run.8,
> “If the daemon does not support changing to a different user id, you can use this to change the user id, **and optionally group id**, before start-stop-daemon(8) or supervise-daemon(8) launches the daemon.”

### DIFF
--- a/sh/start-stop-daemon.sh
+++ b/sh/start-stop-daemon.sh
@@ -55,7 +55,7 @@ ssd_start()
 		${error_logger_arg} \
 		${procname:+--name} $procname \
 		${pidfile:+--pidfile} $pidfile \
-		${command_user+--user} $command_user \
+		${command_user+--chuid} $command_user \
 		${umask+--umask} $umask \
 		$_background $start_stop_daemon_args \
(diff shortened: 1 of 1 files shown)


williamh:
We use --user, not --chuid, so this fix is not correct.

concatime:
But `start_stop_daemon_args='--user nobody:www-data'` does not work, `start_stop_daemon_args='--chuid nobody:www-data'` does.
From `start-stop-daemon --help`:
```
  -u, --user <username|uid>     process owner to check
Options:
  -g, --group <group|gid>       run process as this group
  -c, --chuid <name|uid[:group|gid]>
                                change to this user/group before process
```

concatime:
Up. From the [man](//github.com/OpenRC/openrc/blob/master/man/openrc-run.8#L158) page, `command_user` should also interpret group/gid. But `--user` does not.

williamh:
I will check this out and reply later, but if this is an issue, the fix 
would be to make sure --user works correctly and not switch to --chuid.

PR #257, opened 2018-11-17
sh/functions.sh.in: return a different value for invalid input in yesno()

austin987:
Currently 0 is returned for yes, 1 for no, and 1 for invalid input. This changes invalid input to return 255, so it can be differentiated from no.

@williamh


vapier:
seems like we should use `2` instead of moving into the reserved space above 125

austin987:
Updated to return 2.

PR #261, opened 2018-12-02
Fix/test muslc

austin987:
Thanks to @concatime for pointing out the musl FAQ. This fixes the build with muslc, verified by TravisCI, and adds a TravisCI check to prevent future regressions.

@williamh please let me know if you'd like this handled differently. The BSD .mk include files already define BSD_SOURCE, but given that muslc is Linux only, os-Linux.mk seemed appropriate.


austin987:
helpers.h is already included in libeinfo.c:
https://github.com/OpenRC/openrc/blob/master/src/libeinfo/libeinfo.c#L41

PR #265, opened 2018-12-06
.cirrus.yml: give cirrus-ci a try (freebsd)

gyakovlev:
This will require a bit more work, but it's a good start.
I'll keep pushing to this branch as I figure out how to fix things.

While working on it I found a strange behavior in cirrus-ci, they set `OS=freebsd` (lowercase) while makefile excpects it to be `FreeBSD` so there is an env workaround for now.
https://github.com/cirruslabs/cirrus-ci-docs/issues/100

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,15 @@
+# Cirrus CI integration
+# https://cirrus-ci.org
+
+test_task:
+  freebsd_instance:
+    matrix:
+      image: freebsd-12-0-release-amd64
+      image: freebsd-11-2-release-amd64
+  env:
+    OS: FreeBSD
+  procfs_script: >
+    [ -f /proc/curproc ] || mount -t procfs proc /proc
+  pkg_install_script: pkg install -y bash gawk gmake gsed
+  gsed_hack_script: rm /usr/bin/sed && ln -s /usr/local/bin/gsed /usr/bin/sed
+  test_script: bash test/cirrus.sh
--- a/test/cirrus.sh
+++ b/test/cirrus.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+# Copyright (c) 2007-2018 The OpenRC Authors.
+# See the Authors file at the top-level directory of this distribution and
+# https://github.com/OpenRC/openrc/blob/master/AUTHORS
+#
+# This file is part of OpenRC. It is subject to the license terms in
+# the LICENSE file found in the top-level directory of this
+# distribution and at https://github.com/OpenRC/openrc/blob/master/LICENSE
+# This file may not be copied, modified, propagated, or distributed
+# except according to the terms contained in the LICENSE file.
+
+set -e
+set -u
+set -x
+
+# These are steps to run on Cirrus CI under a jailed FreeBSD system.
+# See $TOP/.cirrus.yml for more info about the Cirrus CI setup.
+
+cpus=$(getconf NPROCESSORS_CONF || echo 1)
+gmake -j"${cpus}" -O
+gmake test


test/cirrus.sh
```diff
@@ -0,0 +1,21 @@
...
+# the LICENSE file found in the top-level directory of this
+# distribution and at https://github.com/OpenRC/openrc/blob/master/LICENSE
+# This file may not be copied, modified, propagated, or distributed
+# except according to the terms contained in the LICENSE file.
+
+set -e
+set -u
+set -x
+
+# These are steps to run on TravisCI under a containerized Ubuntu system.
```
vapier: outdated comment ?

.cirrus.yml
```diff
@@ -0,0 +1,9 @@
+freebsd_instance:
```
vapier: can you add a file header to this file like we have in the travis one ?

.cirrus.yml
```diff
@@ -0,0 +1,9 @@
...
+  image: freebsd-11-2-release-amd64
+
+test_task:
+  procfs_script: >
+    [ -f /proc/curproc ] || mount -t procfs proc /proc
+  pkg_install_script: pkg install -y bash gawk gmake gsed
```
vapier: i think we were striving to support BSD make, and stick to POSIX awk and sed over the GNU versions.  what were you seeing that required the GNU versions of either ?

bash makes sense.

test/cirrus.sh
gyakovlev: yep, I copied travis script originally. will fix.

.cirrus.yml
gyakovlev: will do.

.cirrus.yml
gyakovlev: bsd make complains and I can't build it by just running `make` even on native freebsd vm(non-ci instance). my make-fu is not so good to figure out if it's something trivial or not and I haven't really looked at it.

```make
vagrant@freebsd:~/openrc % make
make: "/home/vagrant/openrc/Makefile" line 14: Cannot open /Makefile.inc
make: "/home/vagrant/openrc/Makefile" line 20: Missing dependency operator
make: "/home/vagrant/openrc/Makefile" line 22: Need an operator
make: "/home/vagrant/openrc/Makefile" line 26: Missing dependency operator
make: "/home/vagrant/openrc/Makefile" line 28: Need an operator
make: "/home/vagrant/openrc/Makefile" line 32: Missing dependency operator
make: "/home/vagrant/openrc/Makefile" line 34: Need an operator
make: "/home/vagrant/openrc/Makefile" line 41: Cannot open /mk/sys.mk
make: "/home/vagrant/openrc/Makefile" line 42: Cannot open /mk/os.mk
make: "/home/vagrant/openrc/Makefile" line 43: Cannot open /mk/subdir.mk
make: "/home/vagrant/openrc/Makefile" line 44: Cannot open /mk/dist.mk
make: "/home/vagrant/openrc/Makefile" line 45: Cannot open /mk/gitver.mk
make: "/home/vagrant/openrc/Makefile" line 48: Missing dependency operator
make: "/home/vagrant/openrc/Makefile" line 50: Missing dependency operator
make: "/home/vagrant/openrc/Makefile" line 52: Need an operator
make: Fatal errors encountered -- cannot continue
make: stopped in /home/vagrant/openrc
```

sed clearly complains about guiisms
check here: https://github.com/OpenRC/openrc/runs/38170947
```
sed: 1: "/^\n*$/{$q1;N;ba
": extra characters at the end of q command
```
awk seems to be fine, I just added it for safety but it's not used here as it has to be called as `gawk`

what's the preferred way of handling bsdsed<->gsed differences?
I see 2 ways:
1* rewrite sed expressions to be agnostic
2* introduce implicit dependency on gsed for tests on bsd and handle sed-gsed somewhere in the scripts, probably using `SED` variable like many projects do.
can't just call `gsed` because gentoo does not install `gsed` symlink or binary for example but some others distros do, but it's inconsistent.

gyakovlev: nope
gnu
```
q [exit-code]
              Immediately quit the sed script without processing any more input, except that if auto-print is not disabled the current pattern space will be printed.  The exit code argument is a GNU extension.
```

bsd
```
[1addr]q
	     Branch to the end of the script and quit without starting a new
	     cycle.
```

so removing the **1** after **q** makes bsdsed happy but probably will never reach `|| echo` part as there will be no non-zero return code.

gyakovlev: opened bsdmake issue here https://github.com/OpenRC/openrc/issues/266

vapier: ok, lets file a bug for the GNU sed thing.  that test logic is way more complicated than it needs to be.  we should just delete the sed entirely.

gyakovlev:
rebased,
freebsd now passes tests.
I'll squash commits for clean history.

the only suspicious thing in the test output is `fcntl(): Bad file descriptor`, can't reproduce that locally on freeBSD vm

PR #268, opened 2018-12-13
scripts: fix shutdown, poweroff, halt, reboot

udeved:
if script args are empty, assign 'now'

Perhaps no ideal solution.


williamh:
These scripts are designed to emulate their sysvinit counterparts.
shutdown requires a time argument, so it was a bug to not require it before.
The others do not, and they pass now by default if one isn't given.
Because of this, I have accepted part of your pull request.

PR #275, opened 2018-12-23
src/rc/supervise-daemon.c: do not pass NULL to strcmp

philhofer:
The following will cause a segfault due to NULL being
passed to `strcmp(3)`

```
$ RC_SVCNAME=foo supervise-daemon
```

Fix the bounds check on `argc` in `main`. If `argc<=1`, then
it is not safe to dereference `argv[1]`.


vapier:
thanks, looks good

PR #278, opened 2018-12-24
Convert internal rc_getline() calls to getline() calls

williamh:
getline has been in posix since POSIX.1-2008, so it should be safe for
us to use it instead of using our wrapper function.

### DIFF
--- a/src/librc/librc-daemon.c
+++ b/src/librc/librc-daemon.c
@@ -141,7 +141,7 @@ rc_find_pids(const char *exec, const char *const *argv, uid_t uid, pid_t pid)
 		fp = fopen("/proc/self/status", "r");
 		if (fp) {
 			while (!feof(fp)) {
-				rc_getline(&line, &len, fp);
+				getline(&line, &len, fp);
 				if (strncmp(line, "envID:\t0", 8) == 0) {
 					openvz_host = true;
 					break;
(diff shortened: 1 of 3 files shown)


williamh:
I would like to call for some testers to make sure this works before I merge it. Please report back here.

williamh:
@dwfreed  The whole reason rc_getline was written was because getline wasn't standard when it was needed. Also, the code already uses both rc_getline and getline. I think I would still rather migrate to getline.

williamh:
@dwfreed  I see that rc_getline is only used in librc. If I do a wrapper, I would want it to be private to the library, not available for general use.

vapier:
the low usage rates implies it's not a big deal to drop even if it slightly simplifies trailing newline semantics. everyone is used to the posix api at this point so it's a bit of a lost battle.

technically we'd need to keep the symbol just to not break abi.

williamh:
@vapier I could use some insight here. Is it ok in this case because of the low usage to break abi and restore it if someone complains? If not, I guess it means that if I keep the symbol I have to keep the function?

dwfreed:
I'm not able to do a proper review right now, but a quick look indicates many of the consumers of rc_getline expect and need the newline dropping semantics. The sheer quantity suggests it would be better to just use rc_getline as a wrapper around getline to drop the newline, otherwise you will have to implement that in several places, which seems error prone and a waste of effort.

dwfreed:
Superseded by #747

PR #285, opened 2019-01-18
zsh-completion: _rc-service enhancement

xdch47:
- support extra_commands of init-script
  - add for support flags / command combination

### DIFF
--- a/zsh-completion/_rc-service
+++ b/zsh-completion/_rc-service
@@ -1,27 +1,32 @@
 #compdef rc-service
 
-if (( CURRENT == 2 )); then
-  _arguments -s \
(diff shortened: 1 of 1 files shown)


williamh:
I would rather you not use sed to parse commands out of the script.
Can you parse the output of "rc-service foo describe" the way the bash completion does?
Also how do you feel about two separate commits (one that adds the functionality that gets the list of services and one for the change that gets the commands from a service)?

xdch47:
Hi, 
I tried to split the commits - the first one is rather a fix, allow to 
complete service and action with flags (e.g. `rc-service -v iptables start`), what did not worked before...
 The second commit adds the extra commands (e.g. `rc-service iptables save`)

PR #287, opened 2019-02-06
net-online: fix ping test timeout

mode89:
ping doesn't wait, if only one probe was requested.


mode89:
Hi @DocMAX!

I made this pull request two years ago, and frankly don't remember why :)
Can you please help by describing your use-case?

thesamesam:
Should be fixed by https://github.com/OpenRC/openrc/commit/e21b01b97e84f81589f52e087d2b2ac53f0bf144.

PR #289, opened 2019-02-11
remove _BSD_SOURCE and _XOPEN_SOURCE feature test macros

williamh:
These are not needed on glibc systems, so this is to find out why they
are needed for musl.


williamh:
@blueness  Hi, this looks like a possible issue with musl, maybe with openrc but I'm not sure. I don't understand why OpenRC needs _BSD_SOURCE and _XOPEN_SOURCE on musl but not gglibc. Can you give me some insight?

Thanks much.

vapier:
ftr, `_DEFAULT_SOURCE` is a glibc-ism and is new to glibc-2.19.  versions older than that did not support `_DEFAULT_SOURCE` at all.  you can consider it (largely) equiv to `_BSD_SOURCE`.  so the question there is more "do you want to support building against <glibc-2.18".  if so, you need to define both.

`_XOPEN_SOURCE` is a [POSIX standard define](http://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_02_01_02).  defining it w/out a version i think is like saying "give me the latest POSIX version".

iiuc, it's a bit redundant with also defining `_POSIX_C_SOURCE`.

but making the argument "we have `_DEFAULT_SOURCE` therefore we don't need `_BSD_SOURCE` or `_XOPEN_SOURCE`" is misleading.  you should define what you want to support precisely (C library versions & portability) and from there we can determine what defines to set up.

williamh:
@vapier As of musl 1.1.5, musl also supports _DEFAULT_SOURCE.
I don't know how I feel about <glibc-2.19 since we don't have it in Gentoo, and I haven't received any issues about not supporting older glibc versions.

williamh:
@vapier  @blueness  Ok, I am pretty confused here. I don't understand why we need so many more feature test macros for musl than we do for glibc. I thought that _DEFAULT_SOURCE on musl was supposed to give the same profile as it does on glibc, but apparently it doesn't. We can't gdrop _DEFAULT_SOURCE because modern glibc is deprecating _BSD_SOURCE in favor of _DEFAULT_SOURCE.

Thoughts?

williamh:
ok, reading the posix definition of _XOPEN_SOURCE, it looks like -D_XOPEN_SOURCE=700 might be what I need instead of the POSIX_C_SOURCE definition. I'm testing that on this p/r shortly.

williamh:
@vapier  Ok, if I put this commit on the main branch, since it builds on current Gentoo glibc and musl (per travis), what does it say we are compatible with in terms of glibc? what about uclibc, or do we care about uclibc?

vapier:
personally i find the date based `_POSIX_C_SOURCE` to be more clear at a glance since `_XOPEN_SOURCE` requires me to open the POSIX spec, find the relevant chapter, and look the constant up in the table.

although at this point, it looks like the meson build has a completely different set of defines, and the changes haven't been synced with the makefiles.

N-R-K:
Looks like this can probably be closed? The gentoo ebuild for openrc doesn't do anything special with musl, so I assume whatever issues this PR was trying to fix has been fixed already.

navi-desu:
this changes only the makefile, and i don't musl needs any special handling anymore, so let's just close it (and if any changes are needed, open a new one for the meson.build)

PR #291, opened 2019-02-14
sysctl: reuse BSD script for BusyBox

makepost:
If asking sysctl for its version on Linux fails, assume it's from
BusyBox and call it like BSD sysctl. When passing a conf line, trim
spaces around `=` with sed, because both BSD and BusyBox expect the
`variable=value` sysctl format. Despite the BusyBox sysctl applet
missing the --system command line switch, OpenRC can use this applet
now, so remove the recommendation to disable its build.

### DIFF
--- a/BUSYBOX.md
+++ b/BUSYBOX.md
@@ -23,8 +23,5 @@ or recognize the nofail option in fstab.
 
 CONFIG_SETFONT -- The setfont applet does not support the -u option from kbd.
 
-CONFIG_BB_SYSCTL -- The sysctl applet does not support the --system command
-line switch.
-
 There is work to get most of these supported by busybox, so this file
 will be updated as things change.
--- a/init.d/sysctl.in
+++ b/init.d/sysctl.in
@@ -24,12 +24,12 @@ BSD_sysctl()
 	for conf in /etc/sysctl.conf /etc/sysctl.d/*.conf; do
 		if [ -r "$conf" ]; then
 			vebegin "applying $conf"
-			while read var comments; do
+			sed 's/ = /=/' "$conf" | while read var comments; do
 				case "$var" in
 				""|"#"*) continue;;
 				esac
 				sysctl -w "$var" >/dev/null || retval=1
-			done < "$conf"
+			done
 			veend $retval
 		fi
 	done
@@ -52,7 +52,12 @@ start()
 	ebegin "Configuring kernel parameters"
 	case "$RC_UNAME" in
 	*BSD|GNU) BSD_sysctl; rc=$? ;;
-	Linux) Linux_sysctl; rc=$? ;;
+	Linux)
+		if sysctl -V > /dev/null 2>&1; then # procps-ng
+			Linux_sysctl; rc=$?
+		else # busybox
+			BSD_sysctl; rc=$?
+		fi ;;
 	esac
 	eend $rc "Unable to configure some kernel parameters"
 }


williamh:
The bsd_sysctl function does not process all of the same directories as the --system switch for procps sysctl, so you would end up with different behaviour on busybox than you would with procps if we do this.
Also, "&>" is a bashism so this should be avoided.

We should either make busybox sysctl have a --system switch that is compatible with procps sysctl or, imo the less preferred way would be to write a separate function in the sysctl service that processes everything the same way procps does using busybox sysctl.

makepost:
Exactly same behavior isn't the goal, what procps does is even 
unexpected given OpenRC's sysctl.d/README. On Gentoo I don't have any of 
the additional paths listed in the procps manual:

```
# ls /run/sysctl.d/*.conf /run/sysctl.d/*.conf 
/usr/local/lib/sysctl.d/*.conf /usr/lib/sysctl.d/*.conf 
/lib/sysctl.d/*.conf /etc/sysctl.conf
ls: cannot access '/run/sysctl.d/*.conf': No such file or directory
ls: cannot access '/run/sysctl.d/*.conf': No such file or directory
ls: cannot access '/usr/local/lib/sysctl.d/*.conf': No such file or 
directory
ls: cannot access '/usr/lib/sysctl.d/*.conf': No such file or directory
ls: cannot access '/lib/sysctl.d/*.conf': No such file or directory
/etc/sysctl.conf
```

Which packages do these directories come from? On a systemd system I do 
have one other match, /usr/lib/sysctl.d/50-default.conf, which comes 
from systemd itself and thus not applicable here.

I'll edit out the &> and rebase.

williamh:
The /*/lib/sysctl.d locations are part of what systemd recommends as locations for packages to install sysctl* files, so we need to check for and not complain about *.conf files in those locations missing.

It is a little more tricky though. My understanding is that the way the files are processed for the --system option  is a little more subtile than just putting all of the files in order and going through them. A file found in a later location in the directories actually fully overrides a file of the same name found earlier in the directories, so if we want to do that in busybox without the --system switch we have to emulate that behavior. That's why I would rather let sysctl deal with the --system switch and why we can't use the bsd_sysctl function for busybox. Feel free to correct me if I am wrong about the --system option.

To check which sysctl you are dealing with, it might be better to capture the output of
`command -v sysctl` in a variable and inspect that.

makepost:
Of multiple files with the same basename, the one in the first dir applies and and the latter are skipped, according to the manual. It's not difficult to script, the question is why? This systemd recommendation about paths and overrides sounds arbitrary, no packages I've got installed except systemd itself have any files in these dirs, and I don't get why keep the BSD method at all if systemd behavior compliance matters.

Will see what distinguishable info the command check provides and try to apply it.

makepost:
Command only prints the path, doesn't make it clear which sysctl is present. With BusyBox, it can be a binary or a symlink to /bin/busybox. I chose the `-V` option for the condition because most BusyBox applets deliberately don't implement it and have it return a failure code.

LW-archlinux:
from man sysctl :

>        --system
>              Load  settings from all system configuration files. Files are read from directories in the following list in given order from top to
>              bottom.  Once a file of a given filename is loaded, any file of the same name in  subsequent directories is ignored.
>              /run/sysctl.d/*.conf
>              /etc/sysctl.d/*.conf
>              /usr/local/lib/sysctl.d/*.conf
>              /usr/lib/sysctl.d/*.conf
>              /lib/sysctl.d/*.conf
>              /etc/sysctl.conf

```
$ pacman -Qo /usr/bin/sysctl
/usr/bin/sysctl is owned by procps-ng 3.3.15-1
$
```

Assuming the man page aligns with the code, the behaviour doesn't seem to be complicated or PID1 specific.
I'm guessing busybox sysctl doesn't come from procps and behaves differently ?

LW-archlinux:
Whoops, I think this is already known to the other posters.

makepost:
Thanks for quoting the procps manual, yes it's what I'm referring to and what I'm asking to clarify. Neither BSD nor Linux systems to which I've got access place their sysctl conf in these additional paths. The expected OpenRC behavior according to the OpenRC doc is to look up just the /etc ones, which is already more than FreeBSD scripts do and that would still be enough. If there's agreeable rationale to match the new procps paths introduced by systemd you can come up with, I'd prefer to update the README in detail alongside implementing this logic. Otherwise should side with the BSD behavior and even remove the use of the --system flag which mismatches what OpenRC README claims OpenRC to be doing.

LW-archlinux:
Did some checking and found this :
debian, fedora , opensuse, gentoo, archlinux have procps-ng
void linux uses uedev + runit and procps-ng
apline linux uses openrc + busybox + musl and procps-ng

This indicates that while systemd devs started this change, it's integration in procps-ng has made it the default sysctl implementation on linux.
Openrc should support it.

Since BSDs and Busybox use a different sysctl implementation and openrc also supports running with them, the question becomes how we should support it.

Atleast 2 approaches seem possible :
- emulate procps-ng sysctl behaviour on busybox / BSDs in openrc code
(It may be possible to get parts of this code upstreamed)

- have openrc only use sysctl functionality shared between procps-ng / busybox / BSDs and split off code specific to procps-ng sysctl.d to a separate package .
This is similar to what was done for tmpfiles.d setup with the opentmpfiles package.

williamh:
We don't really want to force anything on the bsds, they don't have a reason to emulate procps. That is why I don't want to touch the bsd_sysctl function.

For busybox, the ideal would be to get the busybox sysctl command to understand the --system switch. That way, we don't have to make any changes to OpenRC.
There is a bug for this upstream which hasn't been closed or responded to, so it looks like w e just need to find someone to write the patch [1].

[1] https://bugs.busybox.net/show_bug.cgi?id=8556

makepost:
Could you run `ls /run/sysctl.d/*.conf /run/sysctl.d/*.conf /usr/local/lib/sysctl.d/*.conf /usr/lib/sysctl.d/*.conf /lib/sysctl.d/*.conf` wherever you've got access to, and post which packages have files there if any? That'll actually answer whether even on Linux anything needs to be emulated. I've been running this patch for two months on openrc + busybox + musl before posting and haven't run into such a need. IMO it's clean to not hardcode path quirks in a compiled utility, but have a shell script in the init system default to procps and if it's missing then fall back to BSD behavior, which it already has code for and which this patch lets it use.

williamh:
Even if I do that, it only proves that the packages I have installed on my system don't use the feature.
The /run directory is dynamic and files can be written there by packages. The */lib* directories are where packages themselves can install sysctl configurations, and just because I don't have things there doesn't mean other packages that are not installed don't.
The --system switch is a standard on all linux systems that use procps' version of sysctl, so we need to get this switch added to busybox sysctl. Your approach would create a "fix" that only works for OpenRC users. I think this needs to go further up the chain to busybox sysctl.

wltjr:
@williamh I know this is closed, but the desire to have busybox have a ```---system``` parameter seems to have fallen on deaf ears. There has been no activity on the [bugs opened in 2015](https://bugs.busybox.net/show_bug.cgi?id=8556), 3.5 years later. May need to patch downstream and provide patch in bug to upstream. Even then that does not mean it will be accepted and added to busybox. It seems like busybox devs are not interested, not sure a patch will make any difference.

It maybe time to consider alternative approaches as needed. If this comes up again. Either way i would not hold out hopes for busybox getting a ```--system``` parameter like procps-ng anytime soon.

makepost:
@wltjr Following your link, you can join the CC list and write a
comment, devs will see activity. But read above. `--system` is a systemd
quirk, and nobody seems able to explain why it exists or even show which
apps scatter their config all over the disk like it suggests. Convincing
procps-ng developers to deprecate it makes more sense, if you want to
spend any time on this.

PR #292, opened 2019-02-17
src/test/runtests.sh: drop 'readelf'-based tests

trofi:
The 'readelf'-based tests cover a few situations:
1. undefined symbols in shared libraries
2. unexpected exports in shared libraries

Bug #575958 shows that [2.] implementation is too simplistic
in assuming that presence of relocation equals to export presence.

It is incorrect for PLT stubs and local symbols.
Let's just drop these tests.

If one neds to cover [1.] it is better to use LDFLAGS=-Wl,--no-undefined.

Reported-by: Benda Xu
Bug: https://bugs.gentoo.org/575958

### DIFF
--- a/src/test/runtests.sh
+++ b/src/test/runtests.sh
@@ -27,53 +27,6 @@ fail_on_out() {
 	fi
 }
 
-ebegin "Checking exported symbols in libeinfo.so (data)"
-checkit einfo.data $(
-readelf -Ws ${libeinfo_builddir}/libeinfo.so \
-	| awk '$4 == "OBJECT" && $5 == "GLOBAL" && $7 != "UND" {print $NF}' \
-	| LC_ALL=C sort -u
-)
-
-ebegin "Checking exported symbols in libeinfo.so (functions)"
-checkit einfo.funcs $(
-readelf -Ws ${libeinfo_builddir}/libeinfo.so \
-	| awk '$4 == "FUNC" && $5 == "GLOBAL" && $7 != "UND" {print $NF}' \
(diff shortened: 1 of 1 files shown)


vapier:
i don't think we should remove the tests.  these continue to catch mistakes with the API that never really goes away.

the test code already has a filter to try and remove "safe" relocs -- look at the parisc-specific filter.

trofi:
You think all tests should stay?

https://bugs.gentoo.org/575958#c8 has bits of output for failed test. I'm not sure you can just filter out `R_PPC64_REL24` and still protect against external symbols.

I'd say tests against `*.so` files directly are fine and could be tweaked time to time.

My main concern is interpreting `readelf` against `*.o` files:
```
readelf -Wr $(grep -l '#include[[:space:]]"librc\.h"' ${librc_srcdir}/*.c | sed 's:\.c$:.o:')
```

That will need to be extended to distinct LOCAL and HIDDEN symbols to be useful.

williamh:
@trofi I am open to removing these tests. If we come up with something better, we can always add new tests later. Also, it looks like if we drop those tests, we can remove the *.list files from the test directory.

gyakovlev:
if merged this will probably unblock https://github.com/OpenRC/openrc/pull/265

I did not look close enough what's really happening with  readelf tests on FreeBSD, but output differs
https://github.com/OpenRC/openrc/pull/265/checks?check_run_id=58018033

williamh:
I had a pretty detailed IRC conversation with @trofi  over the weekend and it looks like the best option for now is to remove these tests.

Issue #293, opened 2019-02-20
wrong file permissions in /run/openrc, rc-status segfaults, missing checks.


gyakovlev:
Reported this before, just hit that again.

https://bugs.gentoo.org/659906


Basically, for some reason  some files  in /run end up only root readable, namely depconfig, deptree, softlevel.

```bash
# ls -lA /run/openrc/
total 56
drwxrwxr-x 14 root root   380 Feb 19 12:50 .
drwxr-xr-x 23 root root  1060 Feb 19 12:50 ..
drwxr-xr-x 24 root root   480 Feb 19 12:50 daemons
-rw-------  1 root root    29 Feb 19 12:49 depconfig
-rw-------  1 root root 30465 Feb 19 12:49 deptree
drwxr-xr-x  2 root root    40 Feb 19 12:50 exclusive
drwxr-xr-x  2 root root    40 Feb 19 12:50 failed
drwxr-xr-x  2 root root    40 Feb 19 12:49 hotplugged
drwxr-xr-x  2 root root    40 Feb 19 12:49 inactive
-rw-r--r--  1 root root  4676 Feb 19 12:49 nettreeeth0
-rw-r--r--  1 root root  4676 Feb 19 12:50 nettreewg0
drwxr-xr-x 20 root root   400 Feb 19 12:50 options
drwxr-xr-x  2 root root    40 Feb 19 12:49 scheduled
-rw-------  1 root root     7 Feb 19 12:49 softlevel
drwxr-xr-x  2 root root  1200 Feb 19 12:50 started
drwxr-xr-x  2 root root    40 Feb 19 12:50 starting
drwxr-xr-x  2 root root    40 Feb 19 12:49 stopping
drwxr-xr-x  2 root root    40 Feb 19 12:49 tmp
drwxr-xr-x  2 root root    40 Feb 19 12:49 wasinactive
```

```sh
mount | grep run
tmpfs on /run type tmpfs (rw,nosuid,nodev,noexec,mode=755)
```


I do not have /run in fstab or change umask at all.
I use dracut as initrd generator, if that matters.

as a result running rc-status as user is not possible,
and it results in segmentation fault, as openrc tries to operate on empty deptree, but can't load it and does not check that.


```sh
done.
(gdb) run
Starting program: /bin/rc-status 
Runlevel: sysinit
 dmesg                                                                                                                                                                                                                            [  started  ]
 udev                                                                                                                                                                                                                             [  started  ]
[...]

PR #294, opened 2019-02-21
librc/librc-depend.c: fix NULL pointer dereference

gyakovlev:
In some cases deptree or depinfo can be NULL, check
before dereferencing.

Bug: https://bugs.gentoo.org/659906
Issue: https://github.com/OpenRC/openrc/issues/293

It's just a partial fix for mentioned issues, as while it no longer segfaults, I'm still trying to find why files ended up being only root readable in the first place and how to properly handle in openrc code instead of NULLin everything and assuming we are at sysinit runlevel.


gyakovlev:
merged in https://github.com/OpenRC/openrc/commit/7478c104fc479067d71d191dccfa4cbf541311d8
closing

PR #301, opened 2019-02-26
remove hidden-visibility.h

williamh:
@trofi
I believe we spoke about this on IRC, and if I recall correctly, it doesn't
benefit us very much in OpenRC. If this is correct, I would like to remove it.
What are your thoughts?


trofi:
Looks good.

PR #302, opened 2019-03-02
Kill SVCNAME once and for all

dwfreed:


dwfreed:
175 Gentoo packages contain files in their filesdir that reference SVCNAME...

PR #311, opened 2019-06-10
clarify supervise-daemon-guide

5paceToast:
1. The given default for respawn_max is wrong.
2. The example for respawn_period is nonsensical. Try to clarify it.

### DIFF
--- a/supervise-daemon-guide.md
+++ b/supervise-daemon-guide.md
@@ -92,7 +92,8 @@ respawn_max=x
 ```
 
 This is the maximum number of times to respawn a supervised process
-during the given respawn period. The default is unlimited.
+during the given respawn period.
+The default is 10. 0 means unlimited.
 
 ``` sh
 respawn_period=seconds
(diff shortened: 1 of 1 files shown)


supervise-daemon-guide.md
```diff
@@ -101,6 +102,10 @@ respawn_period=seconds
 This works in conjunction with respawn_max and respawn_delay above to
 decide if a process should not be respawned for some reason.
 
-For example, if respawn_period is 60, respawn_max is 2 and respawn_delay
-is 3 and a process dies more than 4 times, the process will not be
-respawned and the supervisor will terminate.
+For example, if a respawn period is 10, respawn_max is 2 the process
+would need to die 3 times within 10 seconds to no longer be respawned.
```
aadityabagga: Just a minor change for clarity:

For example, if respawn period is 10 and respawn_max is 2, the process
would need to die 3 times within 10 seconds to no longer be respawned.

PR #317, opened 2019-08-03
net-online.in: add ! on symbolic check in getinterface()

### LINKED ISSUE #316: net-online get_interfaces() function makes an incorrect assumption in /sys/class/net
```
get_interfaces()
{
	local ifname iftype
	for ifname in /sys/class/net/*; do
		[ -h "${ifname}" ] && continue // <--- The problem is here.
		read iftype < ${ifname}/type
		[ "$iftype" = "1" ] && printf "%s " ${ifname##*/}
	done
}
```

As far as I know, all interfaces in /sys/class/net are symlinks, so it will always continue if it finds any. This leads to no interfaces being returned, and the script becomes stuck until timing out.

thefallenrat:
`get_interfaces()` should skip to the next iteration of the loop
only if it's not symbiolic links, not the other way around

This commit fixes the problem by adding `!` operator on test command
so that it will only skip to the next loop if the file aren't
symbolic links

fixes #316

References:

commit f42ec82f21f3760b829507344ad0ae761e1d59aa
https://www.tldp.org/LDP/abs/html/loopcontrol.html

### DIFF
--- a/init.d/net-online.in
+++ b/init.d/net-online.in
@@ -23,7 +23,7 @@ get_interfaces()
 {
 	local ifname iftype
 	for ifname in /sys/class/net/*; do
-		[ -h "${ifname}" ] && continue
+		[ ! -h "${ifname}" ] && continue
 		read iftype < ${ifname}/type
 		[ "$iftype" = "1" ] && printf "%s " ${ifname##*/}
 	done


williamh:
Fixed in #391.

PR #318, opened 2019-08-07
Add support for the INIT_HALT environment variable

floppym:
Please note that this change is untested; I'm looking for some feedback before I invest the effort in actually testing it.

The shutdown command provided by sysvinit sets the INIT_HALT environment
variable before switching to runlevel 0.

shutdown -hH -> INIT_HALT=HALT
shutdown -hP -> INIT_HALT=POWEROFF

Add a new "halt" init script to interpret this variable and invoke "halt"
or "poweroff" as appropriate.

This restores the ability to actually halt the system instead of always
turning off the power.

Reference:
http://git.savannah.nongnu.org/cgit/sysvinit.git/tree/doc/Changelog?h=2.95#n30
http://git.savannah.nongnu.org/cgit/sysvinit.git/tree/src/shutdown.c?h=2.95#n457
http://git.savannah.nongnu.org/cgit/sysvinit.git/tree/src/shutdown.c?h=2.95#n554

Signed-off-by: Mike Gilbert <floppym@gentoo.org>

### DIFF
--- a/init.d/Makefile
+++ b/init.d/Makefile
@@ -22,7 +22,7 @@ SRCS-FreeBSD+=	adjkerntz.in devd.in dumpon.in encswap.in ipfw.in \
 		mixer.in nscd.in powerd.in syscons.in
 
 SRCS-Linux=	agetty.in binfmt.in devfs.in cgroups.in dmesg.in hwclock.in \
-	consolefont.in keymaps.in killprocs.in modules.in \
+	consolefont.in halt.in keymaps.in killprocs.in modules.in \
 	mount-ro.in mtab.in numlock.in procfs.in net-online.in save-keymaps.in \
 	save-termencoding.in sysfs.in termencoding.in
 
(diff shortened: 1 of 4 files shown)


williamh:
I like it, and if it works I think we should use it. 
Adding this to the shutdown runlevel by default assumes that everyone using OpenRC on Linux uses sysvinit, and this isn't necessarily the case. 
I think A better place to put something like this would be in the support/sysvinit directory along with updating the README stored there.

williamh:
The more I think about this, it is a bug, but I'm not sure about whether the approach here works either. The final halt is run by /etc/inittab after OpenRC shuts down its services.

floppym:
There's no way to inspect an environment variable in `inittab`, so we would need to invoke a external script. Given that it already invokes `openrc shutdown`, I figured adding it to the end of the `shutdown` runlevel would suffice.

floppym:
> The more I think about this, it is a bug, but I'm not sure about whether the approach here works either.

I'm pretty sure this approach would work. There's no requirement that the final halt has to be invoked directly from inittab.

floppym:
Closing this in favor of #319.

PR #322, opened 2019-08-30
remove broken service startup detection for runit services

abchk1234:
The current check used to fail if runsv was not up.
It can take upto 5 seconds for runsv to be invoked on the service.
To report correct status, we can wait for 5 seconds or
check it via `rc-status`

See https://github.com/OpenRC/openrc/pull/253 for the sleep approach and its discussion.

### DIFF
--- a/sh/runit.sh
+++ b/sh/runit.sh
@@ -20,8 +20,6 @@ runit_start()
 	service_link="${RC_SVCDIR}/sv/${service_path##*/}"
 	ebegin "Starting ${name:-$RC_SVCNAME}"
 	ln -snf "${service_path}" "${service_link}"
-	sv start "${service_link}" > /dev/null 2>&1
-	eend $? "Failed to start ${name:-$RC_SVCNAME}"
 }
 
 runit_stop()


aadityabagga:
This patch works correctly when the service works correctly.
In case the runit service does not work as expected `rc-status` will show the service as started while checking `rc-service status <service>` will reveal the true status, ie:

~~~~
# rc-service netdata-runit status
down: /run/openrc/sv/netdata: 1s, normally up, want up
~~~~

In my opinion this is better than the status quo.

williamh:
So is this the better change to merge than #253?

aadityabagga:
> So is this the better change to merge than #253?

Hi @williamh 

From best to worst IMO:

#253 > #322 (this one) > current code

#253 has the best status reporting, but it needs `sleep` for runit timeout.
This one avoids the sleep, but leaves service state detection to `rc-service status`.

If you apply this patch #253 is not needed.

Thanks,
aaditya

williamh:
I have merge #253, so I am closing this.

abchk1234:
Thanks!

PR #323, opened 2019-09-19
service net-online: fix get_interfaces() and add minimum_up

### LINKED ISSUE #316: net-online get_interfaces() function makes an incorrect assumption in /sys/class/net

dweomer:
This change incorporates a fork that I made for k3OS that helps
solve a number of network availability race-conditions.

See rancher/k3os#184.

~This fixes #316 by removing the `test -h $ifname` check altogether.
Not only was the check backwards, resulting in undetected interfaces,
but on k3OS at least, everything in /sys/class/net is a symlink, making
the corrected check redundant.~

This sidesteps "file not found" errors when attempting to read
`carrier` and `operstate` for explicitly specified interfaces that do not
exist.

Lastly, a new `conf.d` variable, `minimum_up`, allows for net-online to
return successfully when at least that number of interfaces are in the
"up" state. The default empty values results in the old behavior of
expecting all specified (or detected) interfaces to be "up". This is useful
when connecting edge devices to the internet by either ethernet or wifi
but not necessarily both (and not necessarily the same device as last boot).

~Fixes #316~
Supercedes #317

### DIFF
--- a/conf.d/net-online
+++ b/conf.d/net-online
@@ -3,6 +3,11 @@
 # default is all interfaces that support ethernet.
 #interfaces=""
 
+# The minimum_up setting controls how many interfaces the net-online
+# service expects to be "up" when deciding whether the network is active.
+# The default is the number of detected/specified interfaces.
+#minimum_up=
+
 # This setting controls whether a ping test is included in the test for
 # network connectivity after all interfaces are active.
 #include_ping_test=no
--- a/init.d/net-online.in
+++ b/init.d/net-online.in
@@ -38,40 +38,41 @@ start ()
 	rc=0
 	interfaces=${interfaces:-$(get_interfaces)}
 	timeout=${timeout:-120}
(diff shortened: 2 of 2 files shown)


dweomer:
Also, I've attempted to normalize the whitespace (using tabs).

williamh:
@dweomer  Please rebase this on master and resolve the conflicts. If you can do that, I'll take a look. Thanks.

PR #324, opened 2019-09-20
Make swap use lvm

graywolf:
If swap is in logical volume, it might not be present yet. So swap
service should use and after lvm.


graywolf:
Are there any issues or blockers on this one?

PR #330, opened 2019-10-23
Run shellcheck in travis

austin987:
@williamh

### DIFF
--- a/ci/travis.sh
+++ b/ci/travis.sh
@@ -16,6 +16,14 @@ set -x
 # These are steps to run on TravisCI under a containerized Ubuntu system.
 # See $TOP/.travis.yml for more info about the TravisCI setup.
 
+# Run shellcheck, but don't fail (yet):
+files_to_check="$(git ls-files | xargs file | grep -e 'POSIX shell script' | cut -d : -f1)"
+
+for shellscript in $files_to_check; do
+    echo "Checking ${shellscript} with shellcheck:"
+    shellcheck -s sh "${shellscript}" || true
+done
+
 cpus=$(getconf _NPROCESSORS_CONF || echo 1)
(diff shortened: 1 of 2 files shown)


ci/travis.sh
```diff
@@ -16,6 +16,14 @@ set -x
 # These are steps to run on TravisCI under a containerized Ubuntu system.
 # See $TOP/.travis.yml for more info about the TravisCI setup.
 
+# Run shellcheck, but don't fail (yet):
+files_to_check="$(git ls-files | xargs file | grep -e 'POSIX shell script' | cut -d : -f1)"
```
vapier: need to use -z with ls-files and -0 with xargs

`file` is not a great heuristic.  you're missing bash scripts for example, as well as all the init scripts.  prob should do something more like inspect the first line for the shebang, and then forcibly check all conf.d/ files, and either add #!/bin/sh to the remaining .sh files, or just always check *.sh & *.sh.in.

ci/travis.sh
```diff
@@ -16,6 +16,14 @@ set -x
...
+
+for shellscript in $files_to_check; do
+    echo "Checking ${shellscript} with shellcheck:"
```
vapier: indent with tabs, not spaces

ci/travis.sh
```diff
@@ -16,6 +16,14 @@ set -x
 # These are steps to run on TravisCI under a containerized Ubuntu system.
 # See $TOP/.travis.yml for more info about the TravisCI setup.
 
+# Run shellcheck, but don't fail (yet):
+files_to_check="$(git ls-files | xargs file | grep -e 'POSIX shell script' | cut -d : -f1)"
+
+for shellscript in $files_to_check; do
```
vapier: missing braces on this var

PR #331, opened 2019-10-23
.cirrus.yml: update runner to FreeBSD-11.3

gyakovlev:
simple update on 11 branch image.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -5,7 +5,7 @@ test_task:
   freebsd_instance:
     matrix:
       image: freebsd-12-0-release-amd64
-      image: freebsd-11-2-release-amd64
+      image: freebsd-11-3-release-amd64
   env:
     OS: FreeBSD
   procfs_script: >


vapier:
the CI failed on the new 11-3 image.  want to take a look ?

gyakovlev:
@vapier looks like CI is the problem itself, agent is not responding. Already tried to re-launch it yesterday, trying again.

williamh:
It looks like this failed, can you check it?

Thanks.

gyakovlev:
it keeps timing out without even starting, not sure anything can be done except getting lucky with free quota.

vapier:
if their CI can't handle any real load, then trying to run against but always failing is going to just add noise and confusion to all our PRs right?

gyakovlev:
what's strange is that 11.2 and 12.0 check works fine, something with scheduling 11.3 image on their gcloud.

I'll close this for now and revisit later, checking 11.2 is still ok, and it did not create any confusion before.

vapier:
thanks. happy to upgrade if we can rely on it :).

PR #350, opened 2020-02-11
Update moused stop function

vimanuelt:
Add stop function
Add keywords: "-stop" and "-shutdown"
Add "provide mouse"

### DIFF
--- a/init.d/moused.in
+++ b/init.d/moused.in
@@ -67,3 +67,25 @@ start()
 
 	eend $retval "Failed to start moused"
 }
+
+stop()
+{
+        if [ -z "$moused_device" ]; then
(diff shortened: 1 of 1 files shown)


vimanuelt:
Updated moused stop function

williamh:
Hey, first off, sorry it took me this long to respond.
You don't need OPENRC_SILENT; EINFO_QUIET handles this function and is part of the libeinfo library.
Also, there is only one service n openrc for the mouse (moused), so you don't really need
"provide mouse".
Let me know what you think.

Thanks,

William

vimanuelt:
I agree with your assessment. Will make corrections as noted.

PR #351, opened 2020-02-14
Add support for starting services in a specified VRF.

kaniini:
The venerable iproute2 utility has recently introduced support
for executing programs in specific VRFs which are virtualized
routing tables.  These are typically used to isolate different
networking planes from each other, for security or flexibility
reasons.

Services which use the normal supervisor/start-stop-daemon
pattern can be configured by setting the vrf variable in the
/etc/conf.d tree for the service.

This allows for things like configuring the sshd service to
run in a management VRF, which is useful for high assurance
environments where the management plane is intended to be
isolated.

Signed-off-by: Ariadne Conill <ariadne@dereferenced.org>

### DIFF
--- a/src/rc/rc-status.c
+++ b/src/rc/rc-status.c
@@ -20,6 +20,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+#include <inttypes.h>
 
 #include "einfo.h"
 #include "queue.h"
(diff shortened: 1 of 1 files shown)


dwfreed:
The variable used for enabling this should start with "rc_" to avoid potential clashing with runscript-specific variables.

Also note that this only applies when using the "declarative" style for runscripts (ie, settings some variables and not defining a start function).  This will not apply to runscripts that use their own start function.

chutz (changes requested):

williamh:
Everything in the provided reviews is correct. The response to the comment about this not applying to scripts that have their own start/stop functions is correct, but if they have their own start/stop functions we assume they know what they are doing.

I would also add a couple of things.
Since this is  linux specific, you would also want to check $RC_UNAME to make sure we are on Linux.

Also, please add documentation for this variable in etc/rc.conf.

Thanks much.

William

PR #354, opened 2020-02-29
modules: support BusyBox moddprobe command

casantos:
Check if modprobe is provided by kmod or BusyBox. The BusyBox one does
not recognize "--first-time". Blacklist support is not always available
and its option is "-b", not "--use-blacklist". The kmod one also accepts
"-b", so the short option works in both cases.

Signed-off-by: Carlos Santos <unixmania@gmail.com>

### DIFF
--- a/init.d/modules.in
+++ b/init.d/modules.in
@@ -11,6 +11,16 @@
 
 description="Loads a user defined list of kernel modules."
 
+if [ "$RC_UNAME" = Linux ]; then
+	# Check if modprobe is provided by kmod or BusyBox. The BusyBox one
+	# does not recognize "--first-time". Blacklist support is not always
+	# available and its option is "-b", not "--use-blacklist". The kmod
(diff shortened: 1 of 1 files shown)


williamh:
There are several issues using the busybox user space utilities with OpenRC (see the BUSYBOX.md file in the repository).
There are bugs open on the busybox site about most of these, but there has been no response there.
The tracker is https://bugs.gentoo.org/529086.
I would prefer that things like this get fixed in busybox because the only way around it from the openrc side is to do checks similar to what you do in this pr or maybe use "command -v" on every command where there might be an issue and have different code branches to support busybox behavior.
I would like to hear comments on this pr about whether we should do the checking in OpenRC or advocate for the fixes to be done in busybox.

casantos:
I'm not involved anymore with the project for which this PR was made, so feel free to drop it.

lu-zero:
Thank you nonetheless.

PR #364, opened 2020-04-19
rc-status: Remove noise from ini-formatted output

mrueg:
Otherwise this would create the following output:

```
rc-status -f ini
 * Caching service dependencies ...        [ ok ]
[default]
dbus =  started
NetworkManager =  started
syslog-ng =  started
...

```


lu-zero (approved):
It sounds good to me.

PR #365, opened 2020-04-19
.cirrus.yml: Update FreeBSD releases

mrueg:
This might fix CI.


gyakovlev:
probably not getting quota on google cloud. maybe it's time just to use latest 12.x and drop 11.x.

Issue #312, opened 2019-07-02
IPv6 static routes


elrido:
I have an environment with a hand full of VMs with static IPv4 and IPv6 addresses. Until recently this has been a very straight forward setup with just a single gateway, so I just had to configure that one router as the default gateway on the network interfaces.

Now I have enabled IPv6 support on a docker VM, so I want to route a certain IPv6 subnet to that host. Because I use static IP configuration, I can't use radvd to advertise this new route for the other hosts to pick up. Hence I turned to the `/etc/conf.d/staticroute` file and using the `/etc/init.d/staticroute` script.

As far as I understand the logic in [staticroute](https://github.com/OpenRC/openrc/blob/54780a45826f789f1291742eb2c99be369bfe64c/init.d/staticroute.in#L91), if I populate the staticroute environment variable in `/etc/conf.d/staticroute` with "net 192.168.0.0 netmask 255.255.255.0 gw 10.73.1.1" (one of the provided example lines) the init script will actually run this command on start:
```bash
route add -net 192.168.0.0 netmask 255.255.255.0 gw 10.73.1.1
```
For my IPv6 case I therefore added "2001:ffff:ffff:1:4900::/79 gw 2001:ffff:ffff:1:4800::23" as the staticroute, which gets turned into:
```bash
route -A inet6 add -2001:ffff:ffff:1:4900::/79 gw 2001:ffff:ffff:1:4800::23
```
This fails to be interpreted by `route`, because of the dash that gets added.

My workaround on my systems is to change the [line 91 in /etc/init.d/staticroute](https://github.com/OpenRC/openrc/blob/54780a45826f789f1291742eb2c99be369bfe64c/init.d/staticroute.in#L91) to:
```bash
				route $family $xtra $2 $args
```
The implication is that I would now have to change any IPv4 routes to start with that dash that I removed from the script.

Given that this is probably a rare edge case and if changed, would break existing IPv4 routes configuration, I am not sure if this should be changed globally. Please let me know if you would accept a PR to document this in the commented examples in the /etc/conf.d/staticroute configuration file or feel free to close this issue. Alternatively I could provide a patch that prevents the dash to get added only for the IPv6 address family.

user318:
Hi,

If I understand right, you shoud use "net" for IPv6 too, so you need for your example:
`net 2001:ffff:ffff:1:4900::/79 gw 2001:ffff:ffff:1:4800::23`
And your config is missing "net" in the beginning.

user318:
[...]

PR #366, opened 2020-05-09
staticroute to ommit erroneous dash args prefix when inet6

axelu:
This is to address issue #312 where the route command is invoked by init.d/staticroute with an erroneous dash in case of inet6.

`route` command syntax for
Adding routes to IPv6 subnets:
`route add -net networkaddress netmask netmask interface`
Adding routes to IPv6 subnets:
`route add -A inet6 networkaddress/prefixlength interface`

The issue is, that the init.d/staticroute script **always** injects the dash in front of the arguments, so in the case of IPv6 it becomes `-networkaddress/prefixlength` which results in an error.
Tracing the actually executed route code in the net-tools package has this also in regards to the expected syntax:
`inet_route [-vF] add {-host|-net} Target[/prefix] [gw Gw] [metric M]`
vs
`inet6_route [-vF] add Target [gw Gw] [metric M] [[dev] If]`

In addition to the inet6 issue, one can also see that actually also declaring the target in the inet aka IPv4 case explicitly as a network or a host is optional, so the dash would also cause problems with that.  When given an address in dotted quad notation, route attempts to guess whether it is a network or a hostname in case net or host is not speficied.

All that being said, `route` is deprecated and the `ip route` command from the networking tools of the iproute2 suite should be used anyway, which can be done by using the staticiproute key in the respective config file to declare a static route, which will cause init.d/staticroute to invoke the `ip route` command instead of the `route` command, and for that the issue does not exist.

### DIFF
--- a/init.d/staticroute.in
+++ b/init.d/staticroute.in
@@ -57,7 +57,7 @@ dump_args()
 
 do_routes()
 {
-	local xtra= family=
+	local xtra= family= argprefix=
 	[ "$RC_UNAME" != Linux ] && xtra=-q
 
 	ebegin "$1 static routes"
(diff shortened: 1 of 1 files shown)


lu-zero (approved):
It seems correct, I guess the set can be squashed on landing.

PR #369, opened 2020-05-14
Accept RC_DEBUG setting from a service script

lego12239:
Sometimes it's need to debug a certain service script on a system startup.
With old-style sysvinit scripts we just add -x to the first line of our
service script. openrc-run doesn't allow us to do this, but we still can
place "set -x" somewhere at the beginning of our script.

This patch allow us to use RC_DEBUG instead of "set -x" for debug
purposes.


lego12239:
I doesn't know is that really need. We can just use "set -x" inside a service script. But if need - there is the patch :-).

PR #376, opened 2020-08-12
Change ping test host to www.w3.org

TheEvilSkeleton:


EchedeyLR:
I think www.w3.org would be more universal

TheEvilSkeleton:
Sure, I'll change

TheEvilSkeleton:
Done http://www.w3.org/

ArcticTheRogue:
Maybe also change the PR name? (if possible)

TheEvilSkeleton:
> Maybe also change the PR name? (if possible)

Thank you for notifying me :)

ArcticTheRogue:
No problem

EchedeyLR:
https://github.com/OpenRC/openrc/blob/72df51e17ba0e1a0f94451b4bbfb338288c4625c/conf.d/net-online#L12 must be edited too

vapier:
this does not seem like a good idea.  google.com has a lot more resources & world wide distribution & guaranteed uptime/availability to handle this kind of load compared to the w3.org website or gentoo.org.

what are you trying to "fix" ?

resynth1943:
In this ping function, it seems to me that the user's IP is sent to Google.

If a person is aiming to avoid Google, this may expose them to Google. For example, Google may very well cross-reference IP's from its different services, and build a profile on them.

---

Also, the w3c website was a very good choice. Thank you, @TheMainOne. That is going tk be very reliable.

vapier:
the user is transmitting a few ICMP packets that will be seen by the local LAN admin, the ISP, and any network in between.  if w3.org is hosted on a cloud service (e.g. Google or Cloudflare), then you're still exposing yourself.

the fact that the service received a few ICMP packets is meaningless.  what exactly could that even add to a "profile" that someone is maintaining ?  as for the idea that this is even happening in the first place, that's pure speculation leading to tin foil hat territory.

if you're paranoid enough about these things, change the default on your system.  that's not justification for moving the load to a random website that is not set up to handle or service this sort of load.

resynth1943:
> if w3.org is hosted on a cloud service (e.g. Google or Cloudflare), then you're still exposing yourself.

From what I can see, that's not true, and seems to be a strawman.

> as for the idea that this is even happening in the first place, that's pure speculation leading to tin foil hat territory.

[Well](https://en.wikipedia.org/wiki/Google_and_privacy_issues), [I'm](https://www.reuters.com/article/us-alphabet-google-privacy-lawsuit-idUSKBN23933H) [understandably](https://www.forbes.com/sites/daveywinder/2019/06/23/google-confirms-creepy-new-privacy-problem/) [suspicious](https://www.zdnet.com/article/goodbye-google-why-and-how-to-take-back-your-privacy/) [of](https://www.cnn.com/2019/07/22/tech/google-street-view-privacy-lawsuit-settlement/index.html) [Google](https://en.wikipedia.org/wiki/Google_litigation)

You should be too, but whether you are is of your own volition.

> that's not justification for moving the load to a random website that is not set up to handle or service this sort of load.

It doesn't necessarily have to be this website. This is still an early pull request, so there's always room for change.

resynth1943:
Sidenote: the Gentoo wiki [recommends pinging gentoo.org](https://wiki.gentoo.org/wiki/Handbook:AMD64/Installation/Networking).

vapier:
the person wanting to make a change needs to provide justification for it.  at this point, there is none beyond paranoia, and it is objectively worst as i already described.

find a default that has the same properties i highlighted and we can consider it.

EchedeyLR:
> random website that is not set up to handle or service this sort of load.

You know W3C webpage, right?

cromerc:
W3C is well known, but is nothing like google. Google has servers all over the world. It is never "down", which is why it's one of the best targets to ping to check if connection works. The only other good target I could even think of would be cloudflare since they run a worldwide cdn network and are almost always up just like google.

vapier:
> You know W3C webpage, right?

as a w3c member having made (minor) contributions to standards, yes, i think i'm "familiar" with them.
but you're ignoring my technical points.

EchedeyLR:
I was sure that W3C had a similar state over the servers. Other alternative would be deb.debian.org which points to a FastCDN or whatever is called and depend on Debian mirrors all over the world.

EchedeyLR:
Mmmm, I was right, I checked to be sure and W3C webpage is using Gandi.

cromerc:
Gandi has 3 data centers in France, 1 in Luxembourg and 4 in the United States... that is not anywhere near similar to what cloudflare or google have.

Using debian mirrors doesn't change anything. They collect data/info when people use their services. If you have a look here is  what debian collects: https://wiki.debian.org/Statistics

Here is an example: http://ftp.ch.debian.org/webalizer/usage_202008.html

So they have ip's of everyone that uses it, where they are from, what you download, etc.

The only way you are truly going to protect your privacy is disable the ping altogether, or point it to your own personal web server instead of a 3rd party one.

resynth1943:
> Gandi has 3 data centers in France, 1 in Luxembourg and 4 in the United States... that is not anywhere near similar to what cloudflare or google have.

I'm 100% sure we can find something that can match Google, in terms of stability, at least. And perhaps privacy. I'll do some research.
(and please no silly discussion locking)

resynth1943:
Would it be wise to use the servers of a domain registrar (or a wheel of them, even)? They, I assume, would have very good uptime.

cromerc:
Just change it in your conf.d file on your machines.

It isn't like the option to choose the server that is pinged doesn't exist.

google is just the default, it's not forced on you or anyone else.

resynth1943:
> It isn't like the option to choose the server that is pinged doesn't exist.

I never disputed that.

> google is just the default, it's not forced on you or anyone else.

Or that.

Nevertheless, I think it may be wise to try and decide on an alternative that satisfies both crowds: privacy-focused, and mainstream.

EchedeyLR:
> Gandi has 3 data centers in France, 1 in Luxembourg and 4 in the United States... that is not anywhere near similar to what cloudflare or google have.
> 
> Using debian mirrors doesn't change anything. They collect data/info when people use their services. If you have a look here is what debian collects: https://wiki.debian.org/Statistics
> 
> Here is an example: http://ftp.ch.debian.org/webalizer/usage_202008.html
> 
> So they have ip's of everyone that uses it, where they are from, what you download, etc.
> 
> The only way you are truly going to protect your privacy is disable the ping altogether, or point it to your own personal web server instead of a 3rd party one.

I know about that in Debian, which is different to Google analisis and use of the info. I contribute sending info in some pcs through popularity-contest package. You compare Google enterprise management with Debian organization/community management and use.

EchedeyLR:
> Gandi has 3 data centers in France, 1 in Luxembourg and 4 in the United States... that is not anywhere near similar to what cloudflare or google have.

I don't understand certain point and the technical stuff referenced in:
> this does not seem like a good idea. google.com has a lot more resources & world wide distribution & guaranteed uptime/availability to handle this kind of load compared to the w3.org website or gentoo.org.

What is the requirement for the ping functionality in order to be equivalent, in resources of what you need?
I doubt that you need the Google/Cloudflare wide network for that but I don't know.

What I am trying to propose are alternatives capable of what you seemed to need, guaranteed availability, good support over the world and good resources.

cromerc:
The point of the ping is that is it succeeds, then all other services that require internet to be up before running are able to start. If the ping fails, those services won't start. That is why it has to be up 99.99% of the time. If mysite.org is down, then the services that require internet don't start...

EchedeyLR:
Gandi ensures 99.95% of availability on their cloud servers so Debian one would be reilable in that sense.

resynth1943:
What about GitHub?

vapier:
availability is only one aspect i highlighted.  regional distribution is another.  only having presence in Europe & the US is heavily skewed towards those small regions and ignores more than half the world's area & population.  being "up" in Luxembourg doesn't help if Taipei can't reach you.  or has a lot of packet latency or loss getting there.

plus, if you want to pin to a network, you shouldn't be using random customer sites like w3.org as a proxy for it.  i've never heard of them before so i can't say whether they have a central point one could pin to.

resynth1943:
> as a w3c member having made (minor) contributions to standards, yes, i think i'm "familiar" with them.
>  i've never heard of them before

Thonkers.

> availability is only one aspect i highlighted. regional distribution is another. only having presence in Europe & the US is heavily skewed towards those small regions and ignores more than half the world's area & population. being "up" in Luxembourg doesn't help if Taipei can't reach you. or has a lot of packet latency or loss getting there.

Fair. I've read into how the internet backbone works, and sometimes it does go down.

resynth1943:
What about pinging the set NTP server? How would that work? Surely that's a little better, right?

EchedeyLR:
>  being "up" in Luxembourg doesn't help if Taipei can't reach you. or has a lot of packet latency or loss getting there.

That should not be an issue if Gandi is using OVH network or has direct connection to it. I didn't include it when speaking about the avaliability because I considered obvious given my point before about Debian "deb.debian.org" site in FastlyCDN.

vapier:
you might be familiar with "Gandi", but that doesn't mean it's commonly recognized by people, nor does it mean people randomly go sniffing their internal network details to see how it's hosted.  please don't make assumptions -- state whatever data you have to support your suggestions, and if no data is presented, then the assumption is that there is none.

if your point is that w3.org is OK because it happens to use Gandi which happens to use OVH network (whatever that is) which is supposed to have good world-wide distribution, then i think that shows even more that w3.org is not a good default.

> NTP server

NTP servers are required to respond to UDP requests on port 123, but there's no requirement i'm aware of that they respond to ICMP, or they have really good uptime.  last i looked, the mirrors tend to be selected via round robin DNS, have regional hostnames that you have to manually pick, and the lower stratum servers are tend to be run by volunteers.  you might randomly get a good hit, and other times not so much.  NTP is meant to be an ongoing service to keep clocks in sync (by slowly stepping them), so if requests now & then get dropped, it's not a big deal -- it's designed to be more resilient.

Google (as have other large cloud companies like Cloudflare) has invested a lot in making things work well by default so users don't have to go fiddling with esoteric settings like hostnames or DNS settings or the like, assuming the device they're using even allows them to control such things.

EchedeyLR:
The other person already told which is the minimum avaliability and something hosted in Gandi finished not being reilable enough.

The point now would be using deb.debian.org which is sponsored by FastlyCDN (https://www.fastly.com/network-map)

PR #388, opened 2020-12-05
user-guide: clarify note for Runlevels/rc-update usage

craftyguy:
I found the original note a little confusing, since using rc-update will
add it to a runlevel so it *is* auto-started when the system reaches
that runlevel again, but I don't think that was the intended meaning of
'auto-start', so hopefully this makes it a little more clear.


lu-zero (approved):
Thank you, it sounds good to me.

PR #391, opened 2020-12-13
net-online: fix process of symlinks in sysfs

gportay:
The test `[ -h "${ifname}" ] && continue` skips the symlinks while it is
the opposite that is the expected: ignoring files that are not symlinks.

Fixes commit f42ec82f21f3760b829507344ad0ae761e1d59aa.

### DIFF
--- a/init.d/net-online.in
+++ b/init.d/net-online.in
@@ -23,7 +23,7 @@ get_interfaces()
 {
 	local ifname iftype
 	for ifname in /sys/class/net/*; do
-		[ -h "${ifname}" ] && continue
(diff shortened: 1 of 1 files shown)


gportay:
Unless I am mistaken, we are expecting the exact opposite.

E5ten:
Would probably be nicer to change the `&&` to `||` instead of having an inverted test

gportay:
Oops, I am sorry for that noise.

williamh:
Fixes #316 and fixes #317

PR #394, opened 2020-12-23
Changed /etc to /etc/conf.d for pf.conf location.

vimanuelt:
Changed /etc to /etc/conf.d for pf.conf location. This complies with OpenRC conventional purpose for /etc/conf.d directory.

### DIFF
--- a/init.d/pf.in
+++ b/init.d/pf.in
@@ -10,7 +10,7 @@
 # except according to the terms contained in the LICENSE file.
 
 name="Packet Filter"
-: ${pf_conf:=${pf_rules:-/etc/pf.conf}}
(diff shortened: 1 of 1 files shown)


williamh:
`/etc/conf.d/pf.conf` is not the same file as `/etc/pf.conf`.
The version in `/etc/conf.d` is for setting the variables used in `/etc/init.d/pf`, -- `pf_rules` and `pf_args`.
`/etc/pf.conf` is the default setting for pf_rules, which is the default packet filter rules file.
Because of that, it doesn't make sense to move `/etc/pf.conf` to `/etc/conf.d`
Please let me know if you need more information.

Thanks.

William

PR #397, opened 2020-12-24
Fixes #396 cgroup_get_pids() do not return pid of subshell

### LINKED ISSUE #396: cgroup cleanup hangs waiting for its own subshell
I'm consistently getting hangs while stopping services with cgroup cleanup enabled. Through debugging I have determined that cgroup_get_pids() finds 2 pids each iteration of the loop. One is the parent openrc shell ($$) and the other is the subshell that cgroup_get_pids() is running in due to process substitution. This means that cgroup_cleanup() spins until the timeout as cgroup_get_pids() always returns a pid.

Replacing cgroup_get_pids() with this fixes the issue:
```bash
cgroup_get_pids()
{
        local cgroup_procs p pids self
        cgroup_procs="$(cgroup2_find_path)"
        read -r self _ < /proc/self/stat
        [ -n "${cgroup_procs}" ] &&
                cgroup_procs="${cgroup_procs}/${RC_SVCNAME}/cgroup.procs" ||
                cgroup_procs="/sys/fs/cgroup/openrc/${RC_SVCNAME}/tasks"
        [ -f "${cgroup_procs}" ] || return 0
        while read -r p; do
                [ "$p" -eq $$ ] || [ "$p" -eq "$self" ] || pids="${pids} ${p}"
        done < "${cgroup_procs}"
        printf "%s" "${pids}"
        return 0
}
```

I'll try to get a proper PR up soon.

matthew-sharp:


matthew-sharp:
I just realised this adds a bashism. Is this project bash specific or do I need to replace this with a portable implementation?

E5ten:
As far as I know the only bashism that the project allows is `local` so I think this would need to be changed to be POSIX.

matthew-sharp:
I've replaced it with a POSIX version. Not as clean though.

williamh:
Does it make sense to replace `subpid=$(exec sh -c 'echo "$PPID"')` with `subpid=$$`?
I played with this on the command line and they both seem to return the same value.
Please correct me if I am wrong.

williamh:
Or, can we get rid of the subpid variable altogether and use `$PPID` directly in the while loop?

matthew-sharp:
Regarding your first comment, that's true most of the time. However because cgroup_get_pids is called via process substitution, $$ refers to the main shell and not the subshell  that cgroup_get_pids is running in. If $$ and $(exec sh -c 'echo "$PPID"') were the same, then this bug would not exist in the first place (also $BASHPID would have no reason to exist).

Regarding  your second comment, I believe that would be possible (would need to test). I just thought it better to start one subshell and capture the PID rather than do that once per iteration of the loop.

matthew-sharp:
Here's a small script to illustrate the difference.

`#!/bin/bash

echo -n 'Parent Shell $$: '
echo $$

echo -n 'Subshell $$: '
subdol=$(echo $$)
echo $subdol

echo -n 'Subshell $PPID: '
subppid=$( echo $(exec sh -c 'echo "$PPID"' ))
echo $subppid`

williamh:
I'm wondering if I should avoid process substitution in this situation.
I'm going to do a quick rewrite on another pr, would you be willing to check it out?

matthew-sharp:
I had a look at your PR, and I think it's the better fix.

PR #401, opened 2021-02-02
start-stop-daemon: fix typo in manpage

sqozz:
Until now there was a small typo in the description of `-m, --make-pidfile` regarding the background option:

```
     -m, --make-pidfile
             Saves the pid of the daemon in the file specified by the -p, --pidfile option. Only useful when used with daemons that run
             in the foreground and forced into the background with the --b, --background option.
```

while in reality the abbreviation of `--background` is `-b` and not `--b`. This small PR fixes that :)


vapier:
thx!

PR #409, opened 2021-02-25
rc-cgroup.sh: avoid process substitution for cgroup_get_pids

### LINKED ISSUE #396: cgroup cleanup hangs waiting for its own subshell

williamh:
This should make cgroup_cleanup work successfully since cgroup_get_pids
no longer uses a subshell.

This fixes #396.

### DIFF
--- a/sh/rc-cgroup.sh
+++ b/sh/rc-cgroup.sh
@@ -24,18 +24,26 @@ cgroup_find_path()
 	printf "%s" "${result}"
 }
 
+# This extracts all pids in a cgroup and puts them in the cgroup_pids
+# variable.
+# It is done this way to avoid subshells so we don't have to worry about
+# locating the pid of the subshell in the cgroup.
+# https://github.com/openrc/openrc/issues/396
 cgroup_get_pids()
 {
-	local cgroup_procs p pids
+	local cgroup_procs p
+	cgroup_pids=
 	cgroup_procs="$(cgroup2_find_path)"
-	[ -n "${cgroup_procs}" ] &&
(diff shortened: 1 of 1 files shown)


matthew-sharp:
I think this a better solution than my PR, trying to track PIDs when there are subshells being started is messy. Normally I wouldn't be a fan of calling a function to set a global variable, but I think it's justified in this case.

One potential suggestion is should there be a comment explaining the rationale for not using subshells?

PR #417, opened 2021-04-03
Ping loop finishing too early

DocMAX:
Ping loop finishing too early. Adding sleep to loop.


thesamesam:
Should be fixed by https://github.com/OpenRC/openrc/commit/e21b01b97e84f81589f52e087d2b2ac53f0bf144.

williamh:
See #487.

williamh:
Correction, see #480.

Issue #420, opened 2021-05-20
service-script-guide.md does not contain any instances of the string "env"


ForestJohnson:
Even if setting environment variables for a service to run under is not directly supported, there should be some mention of whether or not it is supported (and how to do it) in this file, especially because it is a top search engine hit when searching for "openrc init script set environment variables"

ForestJohnson:
Nevermind, I think I might be lost.   I have no idea what the thing I am looking for is called. It looks like this:

```
#!/sbin/openrc-run

supervisor="supervise-daemon"
command="/usr/local/bin/concourse web"
user="concourse"
group="concourse"
directory="/opt/concourse"
```

What is the above thing called?  Is this a service script ?

ForestJohnson:
Ok apparently this _is_ a service script, I just didnt realize it because the section of the docs that would have explained this ("The Default Functions") was too far down the page for me to catch it, buried under other very verbose sections.

I have submitted a PR that contains the word env and it links to the correct location explaining how to set environment variables (`conf.d`)

PR #422, opened 2021-05-21
link to the section of the docs explaining env vars

ForestJohnson:
This way if someone lands on this page and searches for "env" they will not be sorely disappointed.

Fixes https://github.com/OpenRC/openrc/issues/420

### DIFF
--- a/service-script-guide.md
+++ b/service-script-guide.md
@@ -123,7 +123,11 @@ command_args=
 pidfile=
 ```
 
-Thus the 'smallest' service scripts can be half a dozen lines long
+Thus the 'smallest' service scripts can be half a dozen lines long.
+
+## Use conf.d to specify default environment variables
+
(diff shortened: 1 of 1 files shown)


williamh:
The values typically set in conf.d/* are not environment variables; they are just shell variables that affect the startup script.

If you want to set environment variables for a daemon you are running with start-stop-daemon, you should set them in `start_stop_daemon_args` with the --env switch.

ForestJohnson:
terminology issue, when I said "environment variables"   I meant "shell variables that affect the startup script"

thesamesam:
Would you mind fixing the commit too?

ForestJohnson:
ok i updated it, point is the 3 letters "env" need to be present in that file because thats what folks are gonna search for, at least that's what I would expect where I'm from

PR #423, opened 2021-06-13
capabilities: Add support for Linux capabilities(7)

### LINKED ISSUE #314: Set daemon capabilities(7)
Hi,
I would like to be able to easily set a daemon capabilities in a way similar to ``command_user``. This would allow for example to have a service listening on a privileged port (<1024) with never having it running as root.

LinkTed:
This adds capabilities for `start-stop-daemon` and `supervise-daemon` by adding `--capabilities` option. As a result, the user can specify the ambient, inheritable and bounding set by define `capabilities` in the service script.

This fixes #314.


LinkTed:
I did not know that there are [cap_iab(3)](https://man.archlinux.org/man/cap_iab.3.en) functions, so I change the implementations. Now, it is much simpler.

LinkTed:
I added capabilities support for the supervised daemon, but I did not how to test it.

lanodan:
> I added capabilities support for the supervised daemon, but I did not how to test it.

One way you can test it is to run a service binding to <1024 as non-root (`command_user`) and adding the `CAP_NET_BIND_SERVICE` capability to it.

LinkTed:
I test it by calling the  `supervise-daemon` directly as root:
```bash
RC_SVCNAME=test ./supervise-daemon test --start --capabilities ^cap_chown,^cap_net_bind_service /usr/bin/sleep 30
```

It worked:
```bash
$ cat /proc/15521/status
[...]
SigPnd:	0000000000000000
ShdPnd:	0000000000000000
SigBlk:	0000000000000000
SigIgn:	0000000000000000
SigCgt:	0000000000000000
CapInh:	0000000000000401
CapPrm:	000001ffffffffff
CapEff:	000001ffffffffff
CapBnd:	000001ffffffffff
CapAmb:	0000000000000401
[...]
```

LinkTed:
Can we merge this?

jirutka:
@williamh, this is a very useful feature, can you please do a code review and consider merging it?

src/rc/start-stop-daemon.c
```diff
@@ -66,8 +70,13 @@ static struct pam_conv conv = { NULL, NULL};
 
 const char *applet = NULL;
 const char *extraopts = NULL;
+#ifdef HAVE_CAP
+const char *getoptstring = "I:KN:PR:Sa:bc:d:e:g:ik:mn:op:s:tu:r:w:x:1:2:3:4:A:" \
```
vapier, line 74: please do not duplicate the entire string.  this will be a nightmare to maintain.

instead, always define the flags.  add a runtime check to the flag parser that errors out if it's used when HAVE_CAP is not defined.

vapier (changes requested):

LinkTed:
I added the meson support.

meson.build
```diff
@@ -14,9 +14,9 @@ fs = import('fs')
 audit_dep = dependency('audit', required : get_option('audit'))
 if audit_dep.found()
   cc_audit_flags = '-DHAVE_AUDIT'
-  else
+else
   cc_audit_flags = []
-  endif
+endif
```
jirutka, line 19: This change is not related to the topic of this commit/PR, i.e. shouldn’t be here.

sh/start-stop-daemon.sh
```diff
@@ -38,10 +38,6 @@ ssd_start()
 		service_inactive && _inactive=true
 		mark_service_inactive
 	fi
-	[ -n "$output_logger" ] &&
-		output_logger_arg="--stdout-logger \"$output_logger\""
-	[ -n "$error_logger" ] &&
-		error_logger_arg="--stderr-logger \"$error_logger\""
```
jirutka, line 44: This change is not related to the topic of this commit/PR, i.e. shouldn’t be here.

src/rc/start-stop-daemon.c
```diff
@@ -66,7 +70,7 @@ static struct pam_conv conv = { NULL, NULL};
 
 const char *applet = NULL;
 const char *extraopts = NULL;
-const char *getoptstring = "I:KN:PR:Sa:bc:d:e:g:ik:mn:op:s:tu:r:w:x:1:2:3:4:" \
+const char *getoptstring = "I:KN:PR:Sa:bc:d:e:g:ik:mn:op:s:tu:r:w:x:1:2:3:4:B" \
```
vapier, line 73: please keep sorted

src/rc/start-stop-daemon.c
```diff
@@ -96,6 +100,7 @@ const struct option longopts[] = {
 	{ "stderr",       1, NULL, '2'},
 	{ "stdout-logger",1, NULL, '3'},
 	{ "stderr-logger",1, NULL, '4'},
+	{ "capabilities", 1, NULL, 'B'},
```
vapier, line 103: why `-B` ?  is that a common short flag in tools that i'm not aware of ?  seems like `-C` would make more sense since "capabilities" starts with "c" (and `-c` is already taken).

src/rc/start-stop-daemon.c
```diff
@@ -517,6 +526,17 @@ int main(int argc, char **argv)
 			stderr_process = optarg;
 			break;
 
+#ifdef HAVE_CAP
```
vapier, line 529: you can merge the redundant logic

```c
case 'B':
#ifdef HAVE_CAP
   ....
#else
  errorx(...)
#endif
  break;
```

src/rc/start-stop-daemon.c
```diff
@@ -517,6 +526,17 @@ int main(int argc, char **argv)
...
+		case 'B':
+			cap_iab = cap_iab_from_text(optarg);
+			if (cap_iab == NULL)
+				eerrorx("Could not parse iab: %s", strerror(errno));
+			break;
+#else
+		case 'B':
+			eerrorx("Was not compile with capability support: MKCAP=no");
```
vapier, line 537: `MKCAP=no` is specific to the make based system.  just say generically "Capability support not enabled".

vapier (changes requested):

src/rc/start-stop-daemon.c
LinkTed: `C` is also already taken, by `_usage.h`. Which letter should I use?

LinkTed: I used `B` again.

LinkTed:
I resolved all suggestions. Next round. :smile:

src/rc/start-stop-daemon.c
vapier: thanks, i was only looking at local letters

let's not allocate a short option for it at all.  use 0x80 in the long opts array so you can look it up the table.

LinkTed: What do mean? I am not an expert about getopt framework? How I make it only long option?

E5ten: I think it just means replacing all the relevant instances of `'B'` with `0x80`, because it's out of the range of ascii and so won't get passed on the command line, but can still be checked in the switch.

vapier: sorry for not being clear. what @E5ten suggested is good.

LinkTed: Okay, I figured it out. @E5ten Thanks for your help.

LinkTed:
@vapier Can you do a review again?

vapier (changes requested):
sorry, i wrote this comment yesterday, but looks like it wasn't actually published.  the GH phone app is not that good :/.

LinkTed:
> sorry, i wrote this comment yesterday, but looks like it wasn't actually published. the GH phone app is not that good :/.

Done

vapier (approved):
i'll give it a day or two to see if William wants to chime in before merging

LinkTed:
> i'll give it a day or two to see if William wants to chime in before merging

Okay, I will wait then. :smile:

LinkTed:
> i'll give it a day or two to see if William wants to chime in before merging

Should we wait longer?

vapier:
i had the thread snoozed for tonight

thanks for the improvements!

PR #431, opened 2021-07-23
openrc-init: ignore an empty string in argv[1]

floppym:
Bug: https://bugs.gentoo.org/803536

### DIFF
--- a/src/rc/openrc-init.c
+++ b/src/rc/openrc-init.c
@@ -263,7 +263,7 @@ int main(int argc, char **argv)
 
 	printf("OpenRC init version %s starting\n", VERSION);
 
-	if (argc > 1)
+	if (argc > 1 && argv[1][0] != '\0')
 		default_runlevel = argv[1];
 	else
 		default_runlevel = NULL;


vapier:
but like, why ?  it seems like an error to run `init ""`.

floppym:
> but like, why ? it seems like an error to run init "".

I agree, but it was requested in https://bugs.gentoo.org/803536#c4.

It's worth noting that sysvinit just ignores any argument it doesn't recognize, including an empty string.

floppym:
openrc-init behaves a bit differently in that it treats argv[1] as the default runlevel, and performs no validation.

So maybe a better solution would be to check if the specified runlevel exists instead of checking for an empty string specifically.

williamh:
I'll work on this, but I think a better solution even would be to fall back to the default runlevel if the passed in runlevel doesn't exist.

williamh:
@vapier  @floppym  What do you think?

vapier:
> I agree, but it was requested in https://bugs.gentoo.org/803536#c4.

that doesn't make it a good idea ... people request all sorts of things.
commit messages should also have complete explanation & logic in them rather than a loose link to a website.

> I'll work on this, but I think a better solution even would be to fall back to the default runlevel if the passed in runlevel doesn't exist.

if we're running as pid 1, that seems reasonable -- emit a warning, and fallback to the default.

if we aren't running as pid 1, we should show an error & exit non-zero.

floppym:
> commit messages should also have complete explanation & logic in them rather than a loose link to a website.

The reason for the change seems obvious to me. If William wanted some further explanation, I assume he would have asked for it.

floppym:
> that doesn't make it a good idea ... people request all sorts of things.

Let me rephrase: I agree it is an error to invoke openrc-init with an empty string as the first argument. However the request to guard against such a user error also seems reasonable to me.

williamh:
@floppym  I need to check rc_runlevel_exists. The bug you linked says that an empty string passed to that function returns true. If that's the case, that seems to be the underlying issue.

PR #436, opened 2021-08-17
supervise-daemon: implement SSD_IONICELEVEL

whitslack:
`supervise-daemon` was apparently overlooked when support for the `SSD_IONICELEVEL` environment variable was added. This commit brings `supervise-daemon` up to parity with `start-stop-daemon` with respect to this environment variable.


vapier:
sorry for the delay ... my computer is still not resurrected :/

PR #438, opened 2021-08-23
Move .so files from /usr to /

SameExpert:
This will move `libeinfo.so.1` and `librc.so.1` from `/usr` to `/` so that `openrc-init` can be used before mounting `/usr`.

### DIFF
--- a/mk/sys.mk
+++ b/mk/sys.mk
@@ -50,7 +50,7 @@ INCMODE?=		0644
 _LIBNAME_SH=		case `readlink /lib` in /lib64|lib64) echo "lib64";; *) echo "lib";; esac
 _LIBNAME:=		$(shell ${_LIBNAME_SH})
 LIBNAME?=		${_LIBNAME}
-LIBDIR?=		${UPREFIX}/${LIBNAME}
+LIBDIR?=		${PREFIX}/${LIBNAME}
 LIBMODE?=		0644
 
 LIBEXECDIR?=		${PREFIX}/libexec/rc


williamh:
I'm curious why you need this? On Linux you should be mounting /usr from your initramfs before you call openrc-init, or making sure all of the libraries openrc-init needs are in your initramfs if you are calling it there.

williamh:
Or, another option is to override LIBDIR on the make command line.

SameExpert:
> I'm curious why you need this? On Linux you should be mounting /usr from your initramfs before you call openrc-init, or making sure all of the libraries openrc-init needs are in your initramfs if you are calling it there.

I'm having a seperate `@usr` btrfs subvolume, using `openrc-init` as init system and want it to mount it on `/usr`.

I cannot have initramfs because Alpine's init script doesn't allow initramfs without gzip or xz compression method and my kernel is very minimalistic and doesn't have them set in its image.

I use openrc from Alpine Linux's repos directly. Currently I `mv`ed these two _.so_ files out from `/usr` and everything is working fine, it mounts the fstab's entries properly. By default, as an init system it should be at least mount at startup.

williamh:
Hi, can you please check out the 0.44 release? I want to migrate to using a Meson based build system instead of the custom Makefiles we have used in the past.

Can you let me know if you can get the meson build to do what you need?

Thanks,

William

SameExpert:
> Can you let me know if you can get the meson build to do what you need?

I don't know if I did it correctly, but `meson builddir` gives this error:
~~~
meson.build:86:0: ERROR: Unknown statement.
~~~

[full log](https://github.com/OpenRC/openrc/files/7159080/meson-log.txt)

williamh:
This should be fixed in 0.44.5. The split-usr option doesn't do anything in this release so you shouldn't need it.
Please report back after testing 0.44.5.

PR #443, opened 2021-09-08
meson: allow rc override

udeved:
* allow to change the rc dir name of the rc_libexecdir path
*  introduce a librcdir option for override with value  'rc'


aduskett:
This looks good to me as well.

PR #448, opened 2021-09-10
rc-mount: make timeout invocation compatible with busybox timeout

kaniini:
Busybox timeout does not support the `-k` flag. As such, invoking fuser
from do_unmount never worked without this patch. This went unnoticed as
standard error is redirected to /dev/null. This patch fixes this by
simply removing the incompatible `-k` flag.

[Ariadne: the `-k` is redundant anyway, since we are sending the KILL
 signal to begin with.]

### DIFF
--- a/sh/rc-mount.sh
+++ b/sh/rc-mount.sh
@@ -48,7 +48,7 @@ do_unmount()
 		retry=4 # Effectively TERM, sleep 1, TERM, sleep 1, KILL, sleep 1
 		while ! LC_ALL=C $cmd "$mnt" 2>/dev/null; do
 			if command -v fuser >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


vapier:
this is fine because it's redundant as you noted.  thanks.

as a general thing, we've been avoiding rewriting code to workaround busybox limitations because it makes the code largely impossible to verify and keep working, and in some cases, make things even harder to write.  instead, the preference is to fix busybox itself, or in the case of busybox config options, enable the extra ones rather than omit features.  anything that is "standard" or "reasonable" in utilities is fair game, and busybox missing the functionality isn't generally sufficient to have OpenRC avoid it.

kaniini:
haha, i think everyone hates busybox's anachronisms.  i only sent this one upstream because it's redundant ;)

PR #450, opened 2021-09-10
call /sbin/mkmntdirs in localmount OpenRC service

kaniini:
[Ariadne: proposed for inclusion upstream per IRC discussion with WilliamH.]

### DIFF
--- a/init.d/localmount.in
+++ b/init.d/localmount.in
@@ -21,6 +21,8 @@ depend()
 
 start()
 {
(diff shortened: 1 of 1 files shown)


vapier:
there's no explanation as to what "mkmntdirs" is, and it's not POSIX or OpenRC or anything else standard that i can see.

kaniini:
Oh, oops.  I think that’s specific to Alpine.  Let me check.

kaniini:
Yes, mkmntdirs is an Alpine specific utility.  Closing.

PR #454, opened 2021-09-15
Use cgroup.kill in cgroup_cleanup where possible

aidanharris:
The old code is moved to a cgroup_fallback_cleanup function and only
called if cgroup2_kill fails.


williamh:
Hi, how is this feature enabled -- is it a kernel config setting or added in some kernel version?

williamh:
I was informed that this is a feature starting in Linux 5.14.x.

PR #465, opened 2021-10-09
add github action to perform shellcheck on scripts

williamh:


ci/shellcheck.sh
```diff
@@ -0,0 +1,30 @@
...
+		sh/*.in) ;;
+		sh/*.sh) ;;
+		support/init.d.examples/*.in) ;;
+		support/openvpn/*.sh) ;;
+		support/sysvinit/halt.sh) ;;
+		test/*.sh) ;;
+		*) continue ;;
+	esac
+	echo "Checking ${shellscript} with shellcheck:"
+	shellcheck -s sh "${shellscript}" || true
```
vapier, line 28: this ignores all errors, so this job will never fail.  since it will always be green (in the actions tab, and in PRs), no one will ever look at it.  so like, what's the point ? :)

i think you want to set `ret=0` at the top of the file, change this `true` to `ret=1`, and then `exit ${ret}` at the bottom.

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,13 @@
...
+on: [push, pull_request]
+
+jobs:
+
+  shellcheck:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - run: sudo apt-get update -q
+    - run: sudo apt-get install -q -y build-essential libpam-dev meson
```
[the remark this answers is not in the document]
williamh: You are probably right, I'm still learning how to do this. ;-) I'll make that change in the next update.

ci/shellcheck.sh
vapier: none of the CI checks block merges or pushes, so i think it'd be OK to just land this as-is.  but i'm fine with an `exit 0` at the end of the script with a `TODO` explaining why it's ignoring the `$ret` if you want to go that route.

vapier (commented):
should prob resolve all open conversations ...

PR #466, opened 2021-10-09
separate github actions based on OS

williamh:
@thesamesam What do you think about splitting the work flow based on the OS?


thesamesam (approved):
Looks good to me. At some point we can factor the Ubuntu one to share common steps too using the 'matrix' stuff.

PR #478, opened 2021-11-22
SELinux fixes (build system again and fix warning)

thesamesam:


stefson:
works as intended for my setup

PR #488, opened 2021-12-24
Small format improvements

LinkTed:
Fix indentation in the meson.build file.
Use consistent args passing in start-stop-daemon.


vapier:
thx!

PR #489, opened 2021-12-30
rearrange tree

williamh:
This PR does several things to streamline the source tree.

- drop old build system
- move shared include files to common directory
- move shared source files to common directory
- move executable-specific code to separate directories
- move scripts to separate directories under src

All of this should not result in any user facing changes. Please review.

@vapier @thesamesam


.gitignore
```diff
@@ -1,16 +0,0 @@
-*.gz
```
vapier, line 1: this file isn't specific to a build system.  you might want to trim a few entries for the GNU make setup, but deleting it entirely doesn't make sense.

meson.build
```diff
@@ -181,7 +182,7 @@ cc_warning_flags = cc.get_supported_arguments(cc_warning_flags_test)
 cc_flags = [cc_debug_flags, cc_os_flags, cc_warning_flags]
 add_project_arguments(cc_flags, language : 'c')
 
-incdir = include_directories('src/includes')
+incdir = include_directories('src/common')
```
vapier, line 185: i don't think collapsing includes/ into common/ is a good pattern.  having a dir dedicated to headers for APIs that are meant to be reused across the codebase helps enforce good boundaries.  when the dir mixes such "exported" APIs with internal headers, it's hard for other dirs to know what's "internal", and what should be kept that way.

basically, the existence of a header file in a dir should not be a signal by itself that it's OK for reuse outside of the dir.

src/common/meson.build
```diff
@@ -1,6 +1,35 @@
+rc_misc_c = files([
+  'rc-misc.c',
```
vapier, line 2: since these files are in common/ now, do they still really need the `rc-` prefix on the filename ?

meson.build
```diff
@@ -107,6 +107,7 @@ rc_bindir = rc_libexecdir / 'bin'
 rc_sbindir = rc_libexecdir / 'sbin'
 rc_shdir = rc_libexecdir / 'sh'
 sbindir = rootprefix / get_option('sbindir')
+pamdir =	get_option('sysconfdir') / 'pam.d'
```
vapier, line 110: we don't use tabs for indentation elsewhere in here

also, this doesn't seem like it's related to this change (moving programs to dedicated dirs).  seems like this is a bugfix in general that should be in a sep commit for fixing pam.d install ?

src/checkpath/meson.build
```diff
@@ -0,0 +1,9 @@
+executable('checkpath',
```
vapier, line 1: i'm having a hard time seeing why per-program subdirs is an improvement

meson.build
```diff
@@ -201,7 +201,6 @@ subdir('init.d')
 subdir('local.d')
 subdir('man')
 subdir('pkgconfig')
-subdir('scripts')
```
vapier, line 203: same comment as the compiled code ... having a hard time seeing this as improvement and not just a lot of busywork

thesamesam: oo

williamh:
@vapier I'm having a11y issues with the individual review comments, so I'll respond here.
- The first comment in your review is fixed; I addressed this on the master branch.
- I don't understand the second one. All code in "common" is shared across the code base.
- No, the files in common do not need the rc_prefix. I'll take care of that in another pr.
- The fourth one is addressed the same way the first one was, I moved that to master.
- I agree that the fifth and sixth ones are somewhat subjective, but I feel that it makes the source cleaner to have one directory of common files used across the code base and separate directories for the executable-specific code.

.gitignore
williamh: @vapier  This has been updated on master.

williamh:
@vapier  @thesamesam  I would like to merge this farely soon, so can you take a look and comment?

vapier:
prob want to go through all open comment threads and close them out

i don't see anything wrong with the changes other than the existing high level feedback i left already

PR #496, opened 2022-01-29
init.d/bootmisc.in: prevent error due to nonexistant file

dermotbradley:
During boot if the "previous_dmesg" setting is enabled in
/etc/conf.d/bootmisc then during the 1st boot of a machine the
bootmisc init.d script will attempt to move a nonexistant dmesg
file, so generating an error on the console.

Modify the script to only move an existing file.


vapier:
thx bud

PR #497, opened 2022-02-03
Add Posix Scheduling Support to start-stop-daemon

sad-goldfish:


meson.build
```diff
@@ -59,6 +59,12 @@ else
   cc_pam_flags = []
 endif
 
+if get_option('sched')
```
vapier, line 62: does this need to be an option ?  `sched.h` is in POSIX:
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sched.h.html

src/rc/start-stop-daemon.c
```diff
@@ -1005,7 +1024,15 @@ int main(int argc, char **argv)
 
 		for (i = getdtablesize() - 1; i >= 3; --i)
 			close(i);
-
+#ifdef HAVE_SCHED
+		if (scheduler != -1){
```
vapier, line 1028: missing space before the `{`

src/rc/start-stop-daemon.c
```diff
@@ -1005,7 +1024,15 @@ int main(int argc, char **argv)
...
+			sched = malloc(sizeof(sched));
```
vapier, line 1029: why does this need to be allocated on the heap ?  just declare it as a local stack variable.

src/rc/start-stop-daemon.c
```diff
@@ -1005,7 +1024,15 @@ int main(int argc, char **argv)
...
+			sched->sched_priority = sched_prio;
```
vapier, line 1030: you only initialize this field which means the rest are random garbage.  when declaring it on the stack, use a struct designated initializer to fix that.

src/rc/start-stop-daemon.c
```diff
@@ -1005,7 +1024,15 @@ int main(int argc, char **argv)
...
+			if (sched_setscheduler(mypid,scheduler,sched))
```
vapier, line 1031: missing spaces after the commas

src/rc/start-stop-daemon.c
```diff
@@ -1005,7 +1024,15 @@ int main(int argc, char **argv)
...
+				eerror("Failed to set scheduler: %s",strerror(errno));
```
vapier, line 1032: missing space after the comma

src/rc/start-stop-daemon.c
```diff
@@ -549,6 +560,14 @@ int main(int argc, char **argv)
 			stderr_process = optarg;
 			break;
 
+		case 0x81: /* --scheduler "Linux Process scheduler index" */
+#ifdef HAVE_SCHED
+			sscanf(optarg, "%d:%d", &scheduler, &sched_prio);
```
vapier, line 565: this makes the underlying constants part of the CLI which doesn't work as those values are not defined by POSIX.  i think you need to have the CLI specify symbolic names (e.g. `fifo`), and then manually convert that to `SCHED_FIFO` and such.

vapier (changes requested):
needs to update the man page too

sad-goldfish:
All done @vapier

sad-goldfish:
This way we maintain a Posix interface for setting scheduler while also supporting unknown schedulers. (E.g SCHED_ISO on a patched Linux Kernel). Though I do not know of any current implementations of  SCHED_SPORADIC so I have not implemented this.

man/start-stop-daemon.8
```diff
@@ -130,6 +130,12 @@ Data can be from 0 to 7 inclusive.
 Modifies the scheduling priority of the daemon.
 .It Fl -oom-score-adj Ar adj
 Modifies the OOM score adjustment of the daemon.
+.It Fl -scheduler Ar policy
+Sets the scheduling policy of the daemon. Possible values are other, fifo and rr on POSIX systems and, additionally, batch and idle on Linux. If
```
vapier, line 134: line is very long.  please wrap to 80 cols.

src/rc/start-stop-daemon.c
```diff
@@ -103,6 +108,8 @@ const struct option longopts[] = {
 	{ "stdout-logger",1, NULL, '3'},
 	{ "stderr-logger",1, NULL, '4'},
 	{ "progress",     0, NULL, 'P'},
+	{ "scheduler",    1, NULL, 0x81},
```
vapier, line 111: i just updated the code to use enums so we don't have to maintain constants anymore

src/rc/start-stop-daemon.c
```diff
@@ -313,6 +322,8 @@ int main(int argc, char **argv)
 	mode_t numask = 022;
 	char **margv;
 	unsigned int start_wait = 0;
+	char *scheduler = NULL;
```
vapier, line 325: should use `const char *`

src/rc/start-stop-daemon.c
```diff
@@ -59,6 +59,11 @@ static struct pam_conv conv = { NULL, NULL};
 #include <sys/capability.h>
 #endif
 
+#ifdef __linux__
+#include <linux/sched.h>
```
vapier, line 63: seems this is only included for `SCHED_BATCH` ?  but glibc provides this for us, albeit only when `_GNU_SOURCE` is defined.  we try to avoid defining `_GNU_SOURCE`, but we should linux/ includes more.

maybe at the top do:

```
#ifdef __linux__
/* For extra SCHED_* defines. */
# define _GNU_SOURCE
#endif
```

src/rc/start-stop-daemon.c
```diff
@@ -1006,6 +1025,36 @@ int main(int argc, char **argv)
...
+		if (scheduler != NULL) {
+			int scheduler_index;
+			struct sched_param sched =  {.sched_priority = sched_prio};
+			if (strcmp(scheduler, "fifo") == 0)
+				scheduler_index = SCHED_FIFO;
+			else if (strcmp(scheduler, "rr") == 0)
+				scheduler_index = SCHED_RR;
+			else if (strcmp(scheduler, "other") == 0)
+				scheduler_index = SCHED_OTHER;
+#ifdef __linux__
```
vapier, line 1037: would prob be better to use `#ifdef SCHED_BATCH` for each subsection rather than tie it to the OS

src/rc/start-stop-daemon.c
```diff
@@ -1006,6 +1025,36 @@ int main(int argc, char **argv)
...
+#endif
+			else if (sscanf(scheduler, "%d", &scheduler_index) != 1)
+				eerrorx("Unknown scheduler: %s", scheduler);
+
+			if (sched_prio == -1) {
+				sched.sched_priority = sched_get_priority_min(scheduler_index);
+			}
+
+			if (sched_setscheduler(mypid, scheduler_index, &sched))
+				eerror("Failed to set scheduler: %s", strerror(errno));
```
vapier, line 1051: shouldn't this be `eerrorx` so it exits ?

src/rc/start-stop-daemon.c
```diff
@@ -1006,6 +1025,36 @@ int main(int argc, char **argv)
...
+#ifdef __linux__
+			else if (strcmp(scheduler, "batch") == 0)
+				scheduler_index = SCHED_BATCH;
+			else if (strcmp(scheduler, "idle") == 0)
+				scheduler_index = SCHED_IDLE;
+#endif
+			else if (sscanf(scheduler, "%d", &scheduler_index) != 1)
+				eerrorx("Unknown scheduler: %s", scheduler);
+
+			if (sched_prio == -1) {
```
vapier, line 1046: don't need the braces

src/rc/start-stop-daemon.c
```diff
@@ -1006,6 +1025,36 @@ int main(int argc, char **argv)
...
+				sched.sched_priority = sched_get_priority_min(scheduler_index);
+			}
+
+			if (sched_setscheduler(mypid, scheduler_index, &sched))
+				eerror("Failed to set scheduler: %s", strerror(errno));
+		} else if (sched_prio != -1) {
+			const struct sched_param sched =  {.sched_priority = sched_prio};
+			if (sched_setparam(mypid, &sched))
+				eerror("Failed to set scheduler parameters: %s", strerror(errno));
```
vapier, line 1055: same here wrt `eerrorx`

vapier (changes requested):

sad-goldfish:
All done @vapier

src/rc/start-stop-daemon.c
```diff
@@ -81,6 +88,8 @@ enum {
   LONGOPT_OOM_SCORE_ADJ,
   LONGOPT_NO_NEW_PRIVS,
   LONGOPT_SECBITS,
+  LONGOPT_SCHEDULER,
```
vapier, line 91: `SC` before `SE` ;)

vapier:
thx again bud

PR #500, opened 2022-03-05
Add newline for `want`

catleeball:
The comment block for `want` seemed to be unintentionally part of the `use` block. Added a newline so `want` will have its own section.


vapier:
thx!

PR #502, opened 2022-03-08
Add clangd support for development

LinkTed:
Add a clangd config file to the default build directory, so that clangd can find the `compile_commands.json`.


LinkTed:
I also want to add a `.clang-format` file in another PR to set the style code. In addition to that, it would make sense to add a CI jobs, which checks the style of the code by clang format.

PR #503, opened 2022-03-11
Add clang format

LinkTed:
I create a clang format file with options, which I think are useful. This would make the development faster because the style code checking is automated. It seems that the style was not consisted through the files. The code is already formatted according to the clang format file. So, we can see the result of the style code definition. I am open to change the clang format file, if something does not look nice in your opinion. I also add a GitHub action to check the code format for PRs.


src/checkpath/checkpath.c
```diff
@@ -45,32 +45,20 @@ typedef enum {
...
-};
-const char * const longopts_help[] = {
-	"Create a directory if not exists",
-	"Create/empty directory",
-	"Create a file if not exists",
-	"Truncate file",
-	"Create a named pipe (FIFO) if not exists",
-	"Mode to check",
-	"Owner to check (user:group)",
+const struct option longopts[] = {{"directory", 0, NULL, 'd'},
```
vapier, line 50: this is not an improvement :P

.clang-format
```diff
@@ -0,0 +1,84 @@
+# .clang-format
```
vapier, line 1: i thought there was a BasedOnStyle for Linux code, but maybe i'm remembering wrong

i think we largely seem to follow the Linux kernel style, so should we use that to start with ?
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/.clang-format

obviously have to tweak the ForEachMacros, but the rest ?

vapier (changes requested):

.clang-format
LinkTed: Okay, I will use this as a base.

LinkTed: I use the same option for the C language as in https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/.clang-format now.

PR #508, opened 2022-03-28
Seedrng cleanup

williamh:
This is most of the cleanup I want to do. All of these commits will be squashed into 1 before I merge.
The one thing left after this is reworking the urandom init script.
I'm thinking about using urandom-bsd for the bsd side and seedrng for linux, but having them both provide urandom for backward compatibility.

@thesamesam  @zx2c4

### DIFF
--- a/src/rc/seedrng.c
+++ b/src/rc/seedrng.c
@@ -23,6 +23,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <fcntl.h>
+#include <getopt.h>
 #include <poll.h>
 #include <unistd.h>
 #include <time.h>
@@ -37,15 +38,46 @@
 #include "rc.h"
 #include "einfo.h"
 #include "helpers.h"
+#include "_usage.h"
 
 #ifndef GRND_INSECURE
 #define GRND_INSECURE 0x0004 /* Apparently some headers don't ship with this yet. */
 #endif
 
-static const char *SEED_DIR;
-static const char *LOCK_FILE;
-static char *CREDITABLE_SEED;
-static char *NON_CREDITABLE_SEED;
+/* Use long option value that is out of range for 8 bit getopt values.
+ * The exact enum value is internal and can freely change, so we keep the
+ * options sorted.
+ */
+enum long_opts {
+  /* This has to come first so following values stay in the 0x100+ range. */
+  LONGOPT_BASE = 0x100,
+  LONGOPT_LOCK_FILE,
+  LONGOPT_SEED_DIR,
+  LONGOPT_SKIP_CREDIT,
+};
+
+const char *applet = NULL;
+const char *extraopts =NULL;
+const char getoptstring[] = getoptstring_COMMON;
+const struct option longopts[] = {
+	{ "lock-file", 1, NULL, LONGOPT_LOCK_FILE },
+	{ "seed-dir", 1, NULL, LONGOPT_SEED_DIR },
+	{ "skip-credit", 0, NULL, LONGOPT_SKIP_CREDIT },
+	longopts_COMMON
+};
+const char * const longopts_help[] = {
+	"lock file",
+	"directory for seed files",
+	"skip credit",
+	longopts_help_COMMON
+};
+const char *usagestring = NULL;
+
+static char *lock_file = NULL;
+static char *seed_dir = NULL;
+static char *creditable_seed = NULL;
+static char *non_creditable_seed = NULL;
+static bool skip_credit = false;
 
 enum blake2s_lengths {
 	BLAKE2S_BLOCK_LEN = 64,
@@ -245,7 +277,8 @@ static size_t determine_optimal_seed_len(void)
 	int fd = open("/proc/sys/kernel/random/poolsize", O_RDONLY);
 
 	if (fd < 0 || read(fd, poolsize_str, sizeof(poolsize_str) - 1) < 0) {
-		ewarn("Unable to determine pool size, falling back to %u bits: %s", MIN_SEED_LEN * 8, strerror(errno));
+		ewarn("%s: Unable to determine pool size, falling back to %u bits: %s",
+			applet, MIN_SEED_LEN * 8, strerror(errno));
 		ret = MIN_SEED_LEN;
 	} else
 		ret = DIV_ROUND_UP(strtoul(poolsize_str, NULL, 10), 8);
@@ -329,20 +362,20 @@ static int seed_from_file_if_exists(const char *filename, bool credit, struct bl
 		return 0;
 	else if (fd < 0) {
 		ret = -errno;
-		eerror("Unable to open seed file: %s", strerror(errno));
+		eerror("%s: Unable to open seed file: %s", applet, strerror(errno));
 		return ret;
 	}
-	dfd = open(SEED_DIR, O_DIRECTORY | O_RDONLY);
+	dfd = open(seed_dir, O_DIRECTORY | O_RDONLY);
 	if (dfd < 0) {
 		ret = -errno;
 		close(fd);
-		eerror("Unable to open seed directory: %s", strerror(errno));
+		eerror("%s: Unable to open seed directory: %s", applet, strerror(errno));
 		return ret;
 	}
 	seed_len = read(fd, seed, sizeof(seed));
 	if (seed_len < 0) {
 		ret = -errno;
-		eerror("Unable to read seed file: %s", strerror(errno));
+		eerror("%s: Unable to read seed file: %s", applet, strerror(errno));
 	}
 	close(fd);
 	if (ret) {
@@ -351,7 +384,8 @@ static int seed_from_file_if_exists(const char *filename, bool credit, struct bl
 	}
 	if ((unlink(filename) < 0 || fsync(dfd) < 0) && seed_len) {
 		ret = -errno;
-		eerror("Unable to remove seed after reading, so not seeding: %s", strerror(errno));
+		eerror("%s: Unable to remove seed after reading, so not seeding: %s",
+			applet, strerror(errno));
 	}
 	close(dfd);
 	if (ret)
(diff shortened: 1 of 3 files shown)


src/rc/seedrng.c
```diff
@@ -392,10 +433,27 @@ int main(int argc _unused, char *argv[] _unused)
...
+		switch (opt) {
+			case LONGOPT_LOCK_FILE:
+			if (!lock_file)
+				lock_file = xstrdup(optarg);
+			break;
+		case LONGOPT_SEED_DIR:
+			if (!seed_dir)
+				seed_dir = xstrdup(optarg);
+			break;
+		case_RC_COMMON_GETOPT
```
mattst88, line 454: This looks like a typo. Should be

````c
case RC_COMMON_GETOPT:
````

Comment (mattst88) [reply src/rc/seedrng.c:454]: With... presumably something in the body. Or maybe this line just needs to be deleted. Dunno.

Comment (dwfreed) [reply src/rc/seedrng.c:454]: It's a macro: https://github.com/OpenRC/openrc/blob/301161a7a4bda27c7ebd54c537dfec0cbc926365/src/rc/_usage.h#L39-L45

**Review by thesamesam — APPROVED** [review]

File: src/rc/seedrng.c [hunk 2]
```diff
@@ -371,18 +399,31 @@ static int seed_from_file_if_exists(const char *filename, bool credit, struct bl
...
-	SEED_DIR = getenv("SEEDRNG_SEED_DIR");
-	if (!SEED_DIR || !*SEED_DIR)
-		SEED_DIR = "/var/lib/seedrng";
-	LOCK_FILE = getenv("SEEDRNG_LOCK_FILE");
-	if (!LOCK_FILE || !*LOCK_FILE)
-		LOCK_FILE = "/var/run/seedrng.lock";
-	xasprintf(&CREDITABLE_SEED, "%s/seed.credit", SEED_DIR);
-	xasprintf(&NON_CREDITABLE_SEED, "%s/seed.no-credit", SEED_DIR);
+	if (!lock_file) {
+		lock_file = xstrdup(getenv("SEEDRNG_LOCK_FILE"));
```
zx2c4, line 403: Why do you strdup this? What's the point of that heap allocation?

src/rc/seedrng.c
```diff
@@ -371,18 +399,31 @@ static int seed_from_file_if_exists(const char *filename, bool credit, struct bl
...
+		if (!lock_file || !*lock_file)
+			xasprintf(&lock_file, "%s", "/var/run/seedrng.lock");
```
zx2c4, line 405: Here, why do you use "%s", when you could use strdup on the string literal? But more importantly, why allocate at all? Just keep the type as `const char *`, and assign either `getenv()` to it, or the string constant to it.

src/rc/seedrng.c
```diff
@@ -371,18 +399,31 @@ static int seed_from_file_if_exists(const char *filename, bool credit, struct bl
...
-	if (!SEED_DIR || !*SEED_DIR)
-		SEED_DIR = "/var/lib/seedrng";
-	LOCK_FILE = getenv("SEEDRNG_LOCK_FILE");
-	if (!LOCK_FILE || !*LOCK_FILE)
-		LOCK_FILE = "/var/run/seedrng.lock";
-	xasprintf(&CREDITABLE_SEED, "%s/seed.credit", SEED_DIR);
-	xasprintf(&NON_CREDITABLE_SEED, "%s/seed.no-credit", SEED_DIR);
+	if (!lock_file) {
+		lock_file = xstrdup(getenv("SEEDRNG_LOCK_FILE"));
+		if (!lock_file || !*lock_file)
```
zx2c4, line 404: In the event that xstrdup returns a size 0 string (but not a null string), !*lock_file is true, but then you leak memory when xasprintf does its thing.

zx2c4:
While you're cleaning up various things, you might want to import https://git.zx2c4.com/seedrng/commit/?id=05fedc64c18555f391dbf766630ab7f6f42d22d2

src/rc/seedrng.c
```diff
@@ -392,10 +433,27 @@ int main(int argc _unused, char *argv[] _unused)
...
 
+	atexit(cleanup);
+	applet = basename_c(argv[0]);
+	if (geteuid())
+		eerrorx("%s: superuser access is required", applet);
 	umask(0077);
-	if (getuid())
-		eerrorx("This rc helper program requires root");
 
+	while ((opt = getopt_long(argc, argv, getoptstring,
```
zx2c4, line 442: Sounds like it'd make more sense to either do:
- changes via environment variable; or
- changes via getopt_long,
but not both at the same time. Just pick one.

src/rc/seedrng.c
```diff
@@ -404,17 +462,17 @@ int main(int argc _unused, char *argv[] _unused)
...
+	lock = open(lock_file, O_WRONLY | O_CREAT, 0000);
 	if (lock < 0 || flock(lock, LOCK_EX) < 0)
 		eerrorx("Unable to open lock file: %s", strerror(errno));
 
-	ret = seed_from_file_if_exists(NON_CREDITABLE_SEED, false, &hash);
+	ret = seed_from_file_if_exists(non_creditable_seed, false, &hash);
 	if (ret < 0)
 		program_ret |= 1 << 1;
-	ret = seed_from_file_if_exists(CREDITABLE_SEED, !rc_yesno(getenv("SEEDRNG_SKIP_CREDIT")), &hash);
+	ret = seed_from_file_if_exists(creditable_seed, !rc_yesno(getenv("SEEDRNG_SKIP_CREDIT")), &hash);
```
zx2c4, line 475: If you choose to keep the option parsing and get rid of the getenv, maybe also turn SEEDRNG_SKIP_CREDIT into an option too?

Or get rid of the option parsing, and do it all with environment variables.

zx2c4 (changes requested):
I made a few suggestions throughout the code. As well, you might want to import https://git.zx2c4.com/seedrng/commit/?id=05fedc64c18555f391dbf766630ab7f6f42d22d2 on top of these changes, since, by looking at your `atexit` stuff, you seem intent on having a good cleanup sequence at the end.

zx2c4:
> I'm thinking about using urandom-bsd for the bsd side and seedrng for linux, but having them both provide urandom for backward compatibility.

That'd be a clean solution, and then symlinking them to urandom. Or have them both be called urandom, like they are now, but install different files for bsd and for linux. I think that'd be cleanest. But you run into problems with the Makefile doing this, while it's easier to do with Meson. Maybe you can figure out the build system stuff there, and then that'd be the cleanest approach I think.

zx2c4:
> As well, you might want to import https://git.zx2c4.com/seedrng/commit/?id=05fedc64c18555f391dbf766630ab7f6f42d22d2 on top of these changes, since, by looking at your `atexit` stuff, you seem intent on having a good cleanup sequence at the end.

I went ahead and did this in an easy to merge PR here, so as not to clutter this set of cleanups: https://github.com/OpenRC/openrc/pull/509

zx2c4:
Hi @williamh - it looks like you took the content of https://github.com/OpenRC/openrc/pull/509 , added it to your own commit, and changed the authorship details, and then got some details wrong. That doesn't seem right. Can you just merge #509 instead?

src/rc/seedrng.c
```diff
@@ -404,17 +473,19 @@ int main(int argc _unused, char *argv[] _unused)
...
 
-	if (mkdir(SEED_DIR, 0700) < 0 && errno != EEXIST)
-		eerrorx("Unable to create \"%s\" directory: %s", SEED_DIR, strerror(errno));
+	if (mkdir(seed_dir, 0700) < 0 && errno != EEXIST)
+		eerrorx("Unable to create \"%s\" directory: %s", seed_dir, strerror(errno));
 
-	lock = open(LOCK_FILE, O_WRONLY | O_CREAT, 0000);
+	lock = open(lock_file, O_WRONLY | O_CREAT, 0000);
 	if (lock < 0 || flock(lock, LOCK_EX) < 0)
 		eerrorx("Unable to open lock file: %s", strerror(errno));
```
zx2c4, line 481: https://github.com/OpenRC/openrc/pull/509 handles the case of closing lock where lock >= 0, but flock() < 0, if you care about closing everything before the OS does it for you.

src/rc/seedrng.c
zx2c4: @williamh It looks like you've added another option, but are keeping the environment variable thing. Can you tell me why you want to have _two ways_ of configuring the same utility? Just choose one. It doesn't matter which one. But having two ways to do it will lead to hard to debug problems for somebody eventually.

src/rc/seedrng.c
zx2c4: "It looks like" is what I said.

Anyway, please do read #509; it does this more correctly (e.g. handles the case where somebody closes all fds and then exec()s), and fixes the flock thing too.

(Note that none of this actually matters, because it's a short lived utility where the OS cleans up all the resources anyway. But I can understand the preference for cleanliness in exit paths.)

thesamesam:
> Hi @williamh - it looks like you took the content of #509 , added it to your own commit, and changed the authorship details, and then got some details wrong. That doesn't seem right. Can you just merge #509 instead?

Your PR didn't exist when he pushed the changes at the point of this comment?

zx2c4:
> 

My PR did exist then. It was submitted 17 hours ago. (And the commit it imports is from Sunday.)

thesamesam:
> My PR did exist then. It was submitted 17 hours ago. (And the commit it imports is from Sunday.) Anyway, William has since merged #509.

You made the review comments then shortly afterwards submitted the PR. William did not push since before then until just now.

src/rc/seedrng.c
```diff
@@ -392,10 +441,30 @@ int main(int argc _unused, char *argv[] _unused)
 	struct timespec realtime = { 0 }, boottime = { 0 };
 	struct blake2s_state hash;
 
+	atexit(cleanup);
+	applet = basename_c(argv[0]);
+	if (geteuid())
+		eerrorx("%s: superuser access is required", applet);
```
zx2c4, line 447: The `"%s: ...` prefix looks nice and seems reasonable. But I noticed that the other eerror and ewarn statements don't have the `%s: ` prefix in there. I don't know if some types of messages need this and others don't, but maybe this is inconsistent. Up to you, so just FYI.

zx2c4:
> > My PR did exist then. It was submitted 17 hours ago. (And the commit it imports is from Sunday.) Anyway, William has since merged #509.
> 
> You made the review comments then shortly afterwards submitted the PR. William did not push since (a fair bit) before then until just now.
> 
> I don't get what you mean about Sunday given he merged your earlier PR as-is? I don't see any evidence that William pushed after you provided your additional coeanup changes. I was testing this yesterday and saw exactly when he pushed and you opened yours.

You are mistaken. William pushed a fair bit prior. Then I made these comments. Then in the last 30 minutes he merged #509 and then force-pushed this repo.

zx2c4 (changes requested):
Latest cleanups look good. Only two comments:

- the `%s: ...` thing in eerror/ewarn consistency thing I mentioned.
- Having both command line switches and environment variables seems awkward and confusing, and I think it'd make sense to choose one rather than keeping them both.

thesamesam:
Sure.

zx2c4:
> Sure, happy to move on, although my emails support my recollection.

Your "although" means you're not happy to move on.

We have the same set of emails doubtlessly. I'll take some screenshots:

This is from 17 hours ago:

image

This is from even further before that:

image

William's "version" of 509 is from hours ago:

image

Which is supported by the stamp on https://github.com/OpenRC/openrc/commit/e8473b198c0cc9449ac01979d170292784885d2f

Do you want to post opposing screenshots to dispute the above, or do you want to stop arguing over bullshit?

src/rc/seedrng.c
```diff
@@ -404,8 +473,8 @@ int main(int argc _unused, char *argv[] _unused)
 	blake2s_update(&hash, &realtime, sizeof(realtime));
 	blake2s_update(&hash, &boottime, sizeof(boottime));
 
-	if (mkdir(SEED_DIR, 0700) < 0 && errno != EEXIST)
-		eerrorx("Unable to create \"%s\" directory: %s", SEED_DIR, strerror(errno));
+	if (mkdir(seed_dir, 0700) < 0 && errno != EEXIST)
+		eerrorx("Unable to create \"%s\" directory: %s", seed_dir, strerror(errno));
 
 	lock = open(LOCK_FILE, O_WRONLY | O_CREAT, 0000);
```
zx2c4, line 479: You forgot to change `LOCK_FILE` here into `lock_file`, and so the code doesn't compile.

src/rc/seedrng.c
williamh: I tend to agree with this, but there are several places in OpenRC here we do override environment variable settings with command line options, so I added the capability to have some form of consistency with those. If I change that, I'll probably just go with command line parsing.

zx2c4: Going only with command line options seems like a good choice.

Is there a reason or a utility in keeping the environment variables around? Keep in mind that this utility lives in /rc, so it's not like it's generally callable by most things. Pretty isolated tool.

zx2c4: Also, I assume you want to replace /etc/init.d/urandom's usage of the environment variables with the command line switches as part of this patchset?

williamh: If we need to get this into a 0.44.x release, I'll not bother urandom too much before that (I'll rework it in a separate pr) so I can backport easily to 0.44.x and not surprise users.
The previous commit which adds seedrng, along with the squashed one from this pr  is what I would backport to 0.44.x. After that I want to turn urandom into a virtual service (using provide in the depend() function) which will be provided by /etc/init.d/seedrng on Linux and /etc/init.d/urandom-bsd on the bsds.
This may require people to add the appropriate service to their runlevels if they are upgrading.
If we don't need seedrng in 0.44.x, I can keep going in this pr.

zx2c4: Probably seedrng should go into the next release, yea.

zx2c4 (approved):
Sounds like you want to leave the env var removal for later which is fine. So these remaining changes look good to me. Let's merge it!

PR #511, opened 2022-03-31
seedrng: remove environment variables

zx2c4:
Now that we have the command line arguments, there's no need for the environment variables.


zx2c4:
CC @williamh

PR #514, opened 2022-04-09
binfmt: ignore empty lines (#492)

### LINKED ISSUE #492: binfmt not ignoring empty lines
<https://www.freedesktop.org/software/systemd/man/binfmt.d.html#Configuration%20Format>

> Empty lines and lines beginning with ";" and "#" are ignored. Note that this means you may not use those symbols as the delimiter in binary format rules.

But we didn't strip empty lines here:

<https://github.com/OpenRC/openrc/blob/301161a7a4bda27c7ebd54c537dfec0cbc926365/sh/binfmt.sh.in#L18-L44>

<details>
<summary>Example error output</summary>

```shell
binfmt                       | * Loading custom binary format handlers ...
binfmt                       |/usr/lib/openrc/sh/binfmt.sh: line 33: /proc/sys/fs/binfmt_misc/: Is a directory
binfmt                       |/usr/lib/openrc/sh/binfmt.sh: line 35: echo: write error: Invalid argument
binfmt                       |binfmt: invalid entry on line 3 of `/usr/lib/binfmt.d//wine.conf'
```

</details>

mochaaP:
This Pull Request closes #492.

According to <https://freedesktop.org/software/systemd/man/binfmt.d.html#Configuration%20Format>:

> *Empty lines* and lines beginning with ";" and "#" are ignored.
> Note that this means you may not use those symbols as the delimiter in binary format rules.

### DIFF
--- a/sh/binfmt.sh.in
+++ b/sh/binfmt.sh.in
@@ -27,6 +27,7 @@ apply_file() {
 		case $line in
 			\#*) continue ;;
 			\;*) continue ;;
+			'')  continue ;;
 		esac
 
 		local reg=${line#*:}


vapier:
thx bud

Issue #513, opened 2022-04-07
integrate tests into the meson build


williamh:
The tests should be integrated into the meson build system and expanded. OpenRC has never had unit tests and I want to add them. I think it will be easier to do so under Meson than it would have been with the old makefiles.
This issue is for integrating the original tests, then they will be expanded later.

PR #515, opened 2022-04-10
rewrite tests

williamh:
- rewrite tests to work with meson
- add tests to ci


williamh:
This will fix #513  once it is merged.

williamh:
I need some folks from Debian/Ubuntu/Alpine to take a look and tell me what I'm missing.
The tests work on Gentoo, even when I rewrite the code to use all pre 0.56 meson code.
However, as you can see on this pr, the tests break CI for your distros.
Please let me know what I should do.

eli-schwartz:
Not a user of any of those, however I may have some suggestions.

It's a weird error message, but that being said -- this script uses /bin/bash instead of /bin/sh and Alpine prefers not to have bash in the base setup, while the BSDs typically put bash in /usr/local/bin due to hysterical raisins.

williamh:
Hi @eli-schwartz , That was a typo actually, all of the test scripts are meant to be sh scripts. I made that change, and Alpine is the only one that is failing now. Can you take a look?

Thanks much,

William

.github/workflows/ci-alpine-linux.yml
```diff
@@ -23,6 +23,9 @@ jobs:
       - run: meson setup builddir/
         env:
           CC: gcc
-      - run: ninja -C builddir
+      - run: meson compile -C builddir
+        env:
+          CC: gcc
+      - run: meson test -C builddir
         env:
           CC: gcc
```
dwfreed, line 31:
```suggestion
      - run: meson test -C builddir
        id: run-tests
        env:
          CC: gcc
      - run: cat builddir/meson-logs/testlog.txt
        if: ${{ steps.run-tests.outcome == 'failure' }}
```

TheOneric:
It appears this test already failed at 0efc1b133e4182bd53cde78153bd8b5cc2e99448 in Alpine.
Looking at the [POSIX sed specification](https://pubs.opengroup.org/onlinepubs/9699919799/) the issue appears to be that `q` does not take an exit code argument in POSIX, but can in GNU sed. Removing the `1` gets rid of the command parsing error, but of course also means it no longer errors on trailing empty lines. Curiously FreeBSD’s sed does not  document an argument for `q` either in its [man page](https://www.freebsd.org/cgi/man.cgi?query=sed&apropos=0&sektion=0&manpath=FreeBSD+13.0-RELEASE&arch=default&format=html). Is the CI using GNU sed from ports instead of the system’s default sed?
If there’s no reason to avoid awk, the following should do the trick while using only POSIX features iinm:
```diff
--- a/test/check-trailing-newlines.sh
+++ b/test/check-trailing-newlines.sh
@@ -6,7 +6,7 @@ top_srcdir=${SOURCE_ROOT:-..}
 ebegin "Checking trailing newlines in code"
 out=$(cd ${top_srcdir};
        for f in `find */ -name '*.[ch]'` ; do
-               sed -n -e :a -e '/^\n*$/{$q1;N;ba' -e '}' $f || echo $f
+               awk '/^$/ {ret=1; next} 1 {ret=0} END {exit ret}' $f || echo $f
        done)
 [ -z "${out}" ]
 eend $? "Trailing newlines need to be deleted:"$'\n'"${out}"

```

PR #517, opened 2022-04-17
start-stop-daemon: fix: pam_start before chrooting

sijanec:
pam_start(3) must be called before chroot(2) for chroots that do not have PAM-related files present.

related bug report: http://bugs.gentoo.org/show_bug.cgi?id=292632

Example with patched openrc (on Gentoo):
```
rc-service sear.c start
 * /var/run/sear.c: creating directory
 * Starting sear.c ...
```
Service starts successfully. See below to see how is start-stop-daemon being run when this is executed.

Example with unpatched openrc (on Gentoo):
```
strace --follow-forks -o /tmp/out.txt start-stop-daemon --start --exec sear.c --chroot /var/run/sear.c --pidfile /pid --user searc:searc --make-pidfile
 * start-stop-daemon: pam error: Critical error - immediate abort
 * start-stop-daemon: failed to start `sear.c'
```
strace output uploaded to <http://1507103400/krneki/strace.txt>.

Note that on Gentoo openrc is a bit modified. I hope this is still an upstream issue and I'm not falsely submitting a PR, although that may be the case.


williamh:
It looks like supervise-daemon has the same issue, can you fix it there also?

sijanec:
> It looks like supervise-daemon has the same issue, can you fix it there also?
I did the same thing, but I did not test anything.

Another thing: stderr_process and redirect_stderr is also called after chrooting. It might be useful to execute the logger process before chrooting, to avoid including programs for logging and to avoid leaving logs in the chroot.

That would break setups that expect the current behavior, although I doubt that many people went through the hassle of setting up a logger in a chroot (?).

For example using `--stderr-logger logger` does not work in my case of a minimal chroot.

Perhaps four additional arguments could be added, --{stderr,stdout}{-logger,}-before-chroot for this.

williamh:
This commit has too many unrelated changes.
Please rework it so that each commit has a single logical change -- for example, adding an option and documenting it on the man page. Also, doing the pam processing before chroot is called could be a separate commit.

Thanks,

William

createyourpersonalaccount:
@williamh 
Is there still interest for this patch? I could try to rework it into a series of commits as you have asked.

q3cpma:
Well, that'd be nice in any form. Not to act as an entitled nagger, but for the fun anecdote, I lost a full afternoon trying to chroot the transmission-daemon init script (basically re-implementing parts of bwrap myself by doing RO bind mounts for runtime dependencies) only to get stuck on the same PAM error.

Think I'll continue to launch the daemon manually and bwrap it myself (though bwrap can't forward signals at the moment, so I'll have to use RPC to shut it down).

createyourpersonalaccount:
@q3cpma I don't think I'm going to work on this right now, I don't use OpenRC currently.

q3cpma:
> @q3cpma I don't think I'm going to work on this right now, I don't use OpenRC currently.

Don't worry, I wasn't expecting anything of the sort.

N-R-K:
I'm interested in trying to extract the pam related change into a commit and getting it merged.

@q3cpma Do you have a small reproducer for the problem so I can check for myself? Or maybe you'd be willing to test it if I create a new branch?

N-R-K:
I have a PR fixing the issue #839.

It additionally fixes `initgroups()` happening inside the chroot while nearly everything else (e.g gid lookup) happens in root.

I did some basic tests in a VM. If anyone is up for additional testing, that'd be nice.

q3cpma:
On Mon Mar 17, 2025 at 5:54 AM CET, NRK wrote:
> N-R-K left a comment (OpenRC/openrc#517)
>
> I'm interested in trying to extract the pam related change into a commit and getting it merged.
>
> @q3cpma Do you have a small reproducer for the problem so I can check for myself? Or maybe you'd be willing to test it if I create a new branch?

Sorry for the lack of reply from my part. I don't have my example service file anymore and plan on going further than chroot (bwrap) anyway. Hope you don't have a hard time finding an example.

PR #519, opened 2022-04-20
seedrng: incorporate various improvements

zx2c4:
There have been a number of subtle improvements and cleanups to seedrng,
including using openat and locking the directory fd instead of a
separate lock file. Also various stylistic cleanups.

### DIFF
--- a/conf.d/seedrng
+++ b/conf.d/seedrng
@@ -1,17 +1,14 @@
-# The convention in this file is to show the default settings
-# commented out.
+# The convention in this file is to show the default settings commented out.
 #
-# Sometimes you want to have urandom start before "localmount"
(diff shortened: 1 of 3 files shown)


zx2c4:
CC @williamh @thesamesam

thesamesam (approved):
Thank you!

williamh:
@thesamesam  As I said on IRC, I'm curious what @zx2c4  reason was for changing the cleanup/atexit setup to gotos.

zx2c4:
While not strictly necessary, I added in a `free(seed_dir)` to the end of main and made the function flow go into that. Of course we could omit the entire cleanup sequence, since the OS does that for us anyway. But since we're already closing those fds, might as well free the seed_dir var too.

PR #520, opened 2022-04-21
bring coverity action under the team

williamh:


williamh:
@vapier:

I need to know a couple of things about the coverity action.

- What do I need to do to make this pr work?  I would rather have all of
  the pieces of coverity set up under the team if we keep it.
- Is there something I can install on my box that will provide the
  same scan output? I'm not particularly fond of going to the web to get it.

vapier:
what are you trying to fix ?

williamh:
I've figured most of this out, so I won't merge the PR. I just need to take a look now at how to view the secrets.

vapier:
by (GH) design, you can't view the secrets once they've been added to GH:
https://github.com/OpenRC/openrc/settings/secrets/actions

but those values are trivial to lookup and/or change as documented in the action:
https://github.com/marketplace/actions/unofficial-coverity-scan#usage

forking the action itself like you've proposed here just doesn't make sense.  it's a generic action which anyone can use in any project by tweaking the action's settings.  which is why i'm trying to understand what the problem is you're running into.

PR #523, opened 2022-04-27
Coverity needs secrets

jsoref:
Every time I fetch a new `master` from this repository, the coverity check fails like this:
https://github.com/jsoref/openrc/runs/6190299997?check_suite_focus=true

It is possible to configure a workflow to look before it leaps, like this:
https://github.com/jsoref/openrc/actions/runs/2232136346

For a bit of information about how this works, here's the documentation:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-secrets

The version of the flow I'm using is the longer-winded version. It is possible to do the work in the same job, but there are 5 steps that should be skipped, and it's a lot cleaner to skip a job w/ 5 steps than to add conditions to skip each of 5 steps.

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -9,8 +9,26 @@ on:
     branches: [master]
 
 jobs:
+  secrets:
+    runs-on: ubuntu-latest
+    outputs:
+      HAS_SCAN_EMAIL: ${{ steps.check.output.HAS_SCAN_EMAIL }}
+      HAS_SCAN_TOKEN: ${{ steps.check.output.HAS_SCAN_TOKEN }}
+    steps:
+      - run: >
+          echo "::set-output name=HAS_SCAN_EMAIL::${{ env.COVERITY_SCAN_EMAIL != '' }}";
+          echo "::set-output name=HAS_SCAN_TOKEN::${{ env.COVERITY_SCAN_TOKEN != '' }}";
+        id: check
+        env:
+          COVERITY_SCAN_EMAIL: ${{ secrets.COVERITY_SCAN_EMAIL }}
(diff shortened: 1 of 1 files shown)


vapier:
if there is a misconfiguration or typo, then the action will silently & incorrectly pass in OpenRC/openrc which is not ideal.  would really prefer the workflow would be disabled on forks by default, and let people opt back in if they set up their own coverity instance to push to.  GH does that for scheduled workflows, but i think that's the only thing they support currently.

this is also a lot of boilerplate to maintain, and kind of sucks to spin up an entire container just to run some shell code.  the link you cite shows how to do this within an existing job without having to spawn an extra container/host -- can't the `if` be applied to the `uses` step ?

prob should move this to the action itself as an option.  something where the caller can declare that certain github repos must have things configured, and all others are allowed to exit early in case the token isn't set.

jsoref:
I couldn't tell how useful the `meson` step was.

Yes, you can use the `if` in the `uses` step.

I suppose I could see about sending the action a PR. I try not to do a depth-first approach to problems, otherwise I'd never finish.

jsoref:
I didn't intend to close the PR (I blame GitHub).

Fwiw, the action _imagines_ both fields are `required`:
https://github.com/vapier/coverity-scan-action/blob/9bb274b991c374589e112e223df0d08da550cfda/action.yml#L11-L17

which means they should be willing to accept a thing to change how they respond to being missing. Sadly, they're likely to insist on failing, which would not address my problem.

vapier:
> Fwiw, the action imagines both fields are required:

not sure what you mean by imagines.  they are both required in order for coverity to work.

> Sadly, they're likely to insist on failing, which would not address my problem.

i think i have a pretty good idea how they'll react.

PR #527, opened 2022-05-18
Block running service in chroot.

sulincix:
I implement chroot check like this
https://github.com/systemd/systemd/blob/main/src/basic/virt.c#L861

openrc-run must block if chrooted.

### DIFF
--- a/src/openrc/rc.c
+++ b/src/openrc/rc.c
@@ -527,7 +527,7 @@ runlevel_config(const char *service, const char *level)
 	char *conf, *dir;
 	bool retval;
 
-	dir = dirname(init);
+	dir = dirname(dirname(init));
 	xasprintf(&conf, "%s/conf.d/%s.%s", dir, service, level);
 	retval = exists(conf);
 	free(conf);


sulincix:
https://github.com/systemd/systemd/blob/main/src/systemctl/systemctl.c#L1127

dwfreed:
There are plenty of valid reasons to run openrc within a chroot, so I'm inclined to reject this.  (openrc doesn't even need to run as root or with any particular capabilities so long as you don't intend to use the privileges afforded by the missing capabilities.)

src/openrc-run/openrc-run.c
```diff
@@ -1232,6 +1243,11 @@ int main(int argc, char **argv)
 	/* Ok, we are ready to go, so setup selinux if applicable */
 	selinux_setup(argv);
 
+	if(running_in_chroot()){
```
vapier, line 1246: bad style -- missing whitespace around the various parts

vapier (changes requested):
agreed that blanket rejecting chroots is wrong.  i don't think systemd does this either ... seems to me that they default to *not* rejecting it, but you can configure it to do so:
https://github.com/systemd/systemd/blob/v252/src/basic/virt.c#L870

if you want to add an rc.conf setting to let people opt-in to stubbing out chroots, i'm amenable to that.

PR #531, opened 2022-06-09
fix quoting of loggers in start-stop-daemon

nekopsykose:
previously broken in 6034866d1c74d5a23eb9f3e0ebf40c9d278aac93
caused *_logger options to be passed unquoted, so
`error_logger="logger -t .."` would pass -t to s-s-d and fail to start
the service.

### DIFF
--- a/sh/start-stop-daemon.sh
+++ b/sh/start-stop-daemon.sh
@@ -47,8 +47,8 @@ ssd_start()
 		${directory:+--chdir} $directory \
 		${output_log+--stdout} $output_log \
 		${error_log+--stderr} $error_log \
-		${output_logger:+--stdout-logger} "$output_logger" \
-		${error_logger:+--stderr-logger} "$error_logger" \
+		${output_logger:+--stdout-logger \"$output_logger\"} \
+		${error_logger:+--stderr-logger \"$error_logger\"} \
 		${capabilities+--capabilities} "$capabilities" \
 		${secbits:+--secbits} "$secbits" \
 		${no_new_privs:+--no-new-privs} \


kaey:
actually this is not a complete fix either. if `$error_logger` contains `"` it will not work. there are also problems with other variables, for example `$directory` is not double quoted and thus must not contain spaces in current implementation which is a unnecessarily limiting.

I think the only way to fully solve quoting issues here using posix shell is to use intermediate function like this

```
run() {
start-stop-daemon ...... ${directory:+--chdir} "$directory" ....... -- "$@"
}

eval run $command_args $command_args_background
```

Though `$start_stop_daemon_args` would also need to be examined

vapier:
kind of feels like the only way out of this pit is to stop using shell

nekopsykose:
> actually this is not a complete fix either  

ah, indeed it's not. though at least this matches the previous behaviour (i think?); before it was broken in the same ways.
+1 for a more complete non-shell solution, though i'm probably not going to be writing it :)

l29ah:
omg @linkted very consistent wow
had to spend over half an hour figuring why my init script doesn't work and produce funny input after rebooting

LinkTed:
@l29ah Wow it seems you are writing only bug free code. Good for you.

LinkTed:
@nekopsykose sorry for this bug. I did not consider this case. Again, sorry for this inconvenience.

l29ah:
> @l29ah Wow it seems you are writing only bug free code. Good for you.

i'm not suggesting writing bug-free code, ofc; i'm suggesting reverting the commit and releasing a point release 24-ish days ago when it was noticed
wait... nevermind

LinkTed:
So, when you are writing:

> omg @LinkTed very consistent wow


You are suggesting to revert the commit and release a new version? Really? And why did you tag me then?


Maybe next time you have the courage to apologize.

williamh:
I am about to be unavailable for a few days, but I will look at this on 7/14.

jirutka:
@williamh, this is a trivial fix for a bug introduced in the latest release. This bug breaks basically all init scripts that use `output_logger` or `error_logger`. What are you waiting for, what is unclear about this? This should have been already merged and released!

I’m sorry, but I’m really losing confidence in this project.

PR #532, opened 2022-06-16
start-stop-daemon: use vfork to avoid races

Arusekk:
While running `rc-service start docker` on Gentoo,
I found that the command does not start the service 90% of the time,
with an enigmatic 'service crashed' message.

The root cause of this is apparently rc-service spawning a pty,
running start-stop-daemon inside that pty, and exitting,
before start-stop-daemon child process calls setsid(),
which results in the child process being killed with SIGHUP (SI_KERNEL).

Theoretically this bug was present ever since the file was created in
5af58b45146a ("Rewrite the core parts in C. We now provide...")
(or even before that), but it should have been only a minor issue before
45bd125dccdc ("Use a pty for prefixed output instead of pipes for...").
Not sure why nobody has had the issue so far (it has been present for
almost 15 years).

As here setsid() is the last call before execve(), the most natural
locking mechanism is vfork(), as it gives back control to parent
process only after execve() or process termination.
So this way the bug can be fixed by adding a single letter. :-)

Another way to ensure this would be using an O_CLOEXEC file descriptor
or some custom lock, which would need to be released not before setsid().

Fixes: 5af58b45146a ("Rewrite the core parts in C. We now provide...")

### DIFF
--- a/src/start-stop-daemon/start-stop-daemon.c
+++ b/src/start-stop-daemon/start-stop-daemon.c
@@ -864,8 +864,8 @@ int main(int argc, char **argv)
 	if (background)
 		signal_setup(SIGCHLD, handle_signal);
 
-	if ((pid = fork()) == -1)
-		eerrorx("%s: fork: %s", applet, strerror(errno));
+	if ((pid = vfork()) == -1)
+		eerrorx("%s: vfork: %s", applet, strerror(errno));
 
 	/* Child process - lets go! */
 	if (pid == 0) {


Arusekk:
While you are all correct, I did not think about the state changes, because I only saw some error handling and printing (I overlooked environ mangling, but I see that now).  The changes should not 'corrupt' the parent though, since the parent is not running until the child process releases its memory, and the parent is not doing much meaningful things anyway. A probably more proper solution (on Linux at least, not sure whether OpenRC supports other kernels) would be to use clone(CLONE_VFORK) but without CLONE_VM and other problematic flags, but it would be a massive change here, primarily because clone() requires passing a function pointer, allocating some stack etc.

The signal mask will not work, too: if we sigprocmask() the SIGHUP away, it will be queued on arrival, and delivered as soon as the mask is restored in the child, thereby killing the process anyway.  If we do not unmask the signals, the daemon will be left unable to be signaled with SIGHUP (which many daemons implement as a way to communicate reloading their config for example).

Creating an O_CLOEXEC pipe would work probably work here, by closing the write end and blocking on reading from the read end in the parent, closing the read end and leaving the write end open (until process termination or execve() closes it automatically) in the child.  This would be the most portable, while keeping no need of having to unlock the lock on every single error path.  A similar file descriptor locking mechanism is implemented for example in xss-lock.

vapier:
the child is opening file descriptors and creating FILE handles.  it's modifying the current environment which is shared.  i'm not auditing the C library for every function we use, directly or indirectly, to understand how its internal state is maintained, or what the compiler does with the stack.  the answer is "use vfork only as documented", and this clearly violates the strong statements in the vfork documentation.

i haven't looked at the bug/runtime situation you're describing in depth, so i don't have any immediate feedback as to a possible solution.  if the parent needs to wait for the child to exit or detach itself from the current session, and a sep program (ssd) is detaching from the current session, i don't think a pipe lock would work since ssd is unaware of that lock.  i can't think of any APIs off the top of my head the parent could use to monitor the child state short of polling it and checking its session id.

floppym:
> if the parent needs to wait for the child to exit or detach itself from the current session, and a sep program (ssd) is detaching from the current session, i don't think a pipe lock would work since ssd is unaware of that lock.

In this case, start-stop-daemon acts as both the parent process, and the child process which calls `setsid`. The pipe-based solution should work fine.

dwfreed:
Superseded by #557

PR #536, opened 2022-07-20
start-stop-daemon.sh: make command_user work properly

FreelancePoliceman:
From the openrc-run manpage, the command_user variable has this description: "If the daemon does not support changing to a user id, you can use this to change the user id, and optionally group id,  before start-stop-daemon(8) or supervise-daemon(8) launches the daemon."

My understanding is that start-stop-daemon's --user flag is not what is intended here, but rather --chuid. Prior to this change, I could not make this flag function as intended (i.e. scripts would run as root no matter what username/group combination was entered), and this edit seems to fix it.

### DIFF
--- a/sh/start-stop-daemon.sh
+++ b/sh/start-stop-daemon.sh
@@ -54,7 +54,7 @@ ssd_start()
 		${no_new_privs:+--no-new-privs} \
 		${procname:+--name} $procname \
 		${pidfile:+--pidfile} $pidfile \
-		${command_user+--user} $command_user \
+		${command_user+--chuid} $command_user \
 		${umask+--umask} $umask \
 		$_background $start_stop_daemon_args \
 		-- $command_args $command_args_background


williamh:
IF you look at start-stop-daemon.c, you will see that --chuid and --user activate the same code. --chuid is going to be removed, so this change is not needed.
What exactly is the issue you are having?

Thanks,

William

FreelancePoliceman:
> IF you look at start-stop-daemon.c, you will see that --chuid and --user activate the same code. --chuid is going to be removed, so this change is not needed. What exactly is the issue you are having?
> 
> Thanks,
> 
> William

You are completely correct. I apologize for the trouble. I am using Devuan, which takes its OpenRC package from Debian, and apparently this package has made a few changes for compatibility with sysvinit. I assume then that this is a Debian/Devuan issue; I made a few tests using another distibution running openrc, and did not encounter problems. 

This is likely completely irrelevant to this project, but I will describe my problem in case anyone searching for this has encountered similar problems. I have an init script for a torrenting program with the flag `command_user="torrent:torrent"` set. Attempting to start this script yields the error: "start-stop-daemon: user 'torrent:torrent' not found." (Both this user and group exist.) Trying  `command_user="torrent"` does begin the script, but it runs as root, with ": torrent" then passed to the command as an argument. Varying the program run, or the user and group passed to command_user, yield identical results. Changing --user to --chuid, however, makes these scripts function as intended.  

Again, I'm sorry to bother you with what is almost certainly a downstream issue.

PR #537, opened 2022-07-20
openrc: fix help messages

xdavidwu:
The two lines seem to both belong to --override, but made into seperate
array elements accidentally, making options after --override and their
help mismatch. This fixes it.


vapier:
thx bud

PR #541, opened 2022-08-21
start-stop-daemon, supervise-daemon: use closefrom()/close_range()

whitslack:
On systems with a very large `RLIMIT_NOFILE`, calling `close()` in a loop from 3 to `getdtablesize()` effects an enormous number of system calls. There are better alternatives. Both BSD and Linux have the `closefrom()` system call that closes all file descriptors with indices not less than a specified minimum. Have `start-stop-daemon` call `closefrom()` on systems where it's implemented, falling back to the old loop elsewhere.

Likewise, calling `fcntl(i, F_SETFD, FD_CLOEXEC)` in a loop from 3 to `getdtablesize()` raises a similar performance concern. Linux 5.11 and onward has a `close_range()` system call with a `CLOSE_RANGE_CLOEXEC` flag that sets the `FD_CLOEXEC` flag on all file descriptors in a specified range. Have `supervise-daemon` utilize this feature on systems where it's implemented, falling back to the old loop elsewhere.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -192,6 +192,14 @@ if cc.compiles(malloc_attribute_test, name : 'malloc attribute with arguments')
   add_project_arguments('-DHAVE_MALLOC_EXTENDED_ATTRIBUTE', language: 'c')
 endif
 
+if cc.has_function('closefrom', prefix: '#define _GNU_SOURCE\n#include <unistd.h>')
+  add_project_arguments('-DHAVE_CLOSEFROM', language: 'c')
+endif
+if cc.has_function('close_range', prefix: '#define _GNU_SOURCE\n#include <unistd.h>') and \
+    cc.has_header_symbol('unistd.h', 'CLOSE_RANGE_CLOEXEC', prefix: '#define _GNU_SOURCE')
+  add_project_arguments('-DHAVE_CLOSE_RANGE_CLOEXEC', language: 'c')
+endif
+
 incdir = include_directories('src/shared')
 einfo_incdir = include_directories('src/libeinfo')
 rc_incdir = include_directories('src/librc')
--- a/src/start-stop-daemon/start-stop-daemon.c
+++ b/src/start-stop-daemon/start-stop-daemon.c
@@ -1105,8 +1105,7 @@ int main(int argc, char **argv)
 			dup2(stderr_fd, STDERR_FILENO);
 
 		for (i = getdtablesize() - 1; i >= 3; --i)
-			if (i != pipefd[1])
-				close(i);
+			close(i);
 
 		if (scheduler != NULL) {
 			int scheduler_index;
--- a/src/start-stop-daemon/start-stop-daemon.c
+++ b/src/start-stop-daemon/start-stop-daemon.c
@@ -1104,8 +1104,12 @@ int main(int argc, char **argv)
 				|| rc_yesno(getenv("EINFO_QUIET")))
 			dup2(stderr_fd, STDERR_FILENO);
 
+#ifdef HAVE_CLOSEFROM
+		closefrom(3);
+#else
 		for (i = getdtablesize() - 1; i >= 3; --i)
 			close(i);
+#endif
 
 		if (scheduler != NULL) {
 			int scheduler_index;
--- a/src/supervise-daemon/supervise-daemon.c
+++ b/src/supervise-daemon/supervise-daemon.c
@@ -22,6 +22,11 @@
 #define ONE_SECOND    1000000000
 #define ONE_MS           1000000
 
+#ifdef HAVE_CLOSE_RANGE_CLOEXEC
+/* For close_range() */
+# define _GNU_SOURCE
+#endif
+
 #include <sys/types.h>
 #include <sys/ioctl.h>
 #include <sys/resource.h>
@@ -569,8 +574,11 @@ static void child_process(char *exec, char **argv)
 	if (redirect_stderr || rc_yesno(getenv("EINFO_QUIET")))
 		dup2(stderr_fd, STDERR_FILENO);
 
-	for (i = getdtablesize() - 1; i >= 3; --i)
-		fcntl(i, F_SETFD, FD_CLOEXEC);
+#ifdef HAVE_CLOSE_RANGE_CLOEXEC
+	if (close_range(3, UINT_MAX, CLOSE_RANGE_CLOEXEC) < 0)
+#endif
+		for (i = getdtablesize() - 1; i >= 3; --i)
+			fcntl(i, F_SETFD, FD_CLOEXEC);
 	cmdline = make_cmdline(argv);
 	syslog(LOG_INFO, "Child command line: %s", cmdline);
 	free(cmdline);


floppym (changes requested):
This needs to be reworked to avoid closing the pipe that was added in #557.

whitslack:
> This needs to be reworked to avoid closing the pipe that was added in #557.

Hmm, #557 didn't need to exempt `pipefd[1]` from the `close` loop in the first place. The pipe FDs are opened with `O_CLOEXEC`, so they will be closed automatically upon `exec`. The only code between the `close` loop and the `execvp` call is to set the scheduler and scheduler parameters, and that code doesn't touch the pipe. Indeed, the pipe only needs to remain open in the child process until after the `setsid()` call, which occurs well before the `close` loop. I'll rebase this PR to resolve the conflict, but I'll do so by eliminating the needless carve-out of `pipefd[1]` from the fallback `close` loop and retaining the `closefrom` call as is.

meson.build
```diff
@@ -192,6 +192,15 @@ if cc.compiles(malloc_attribute_test, name : 'malloc attribute with arguments')
   add_project_arguments('-DHAVE_MALLOC_EXTENDED_ATTRIBUTE', language: 'c')
 endif
 
+if cc.has_function('closefrom', prefix: '#define _GNU_SOURCE\n#include <unistd.h>')
+  add_project_arguments('-DHAVE_CLOSEFROM', language: 'c')
+endif
+if os == 'Linux' and \
```
vapier, line 198: `_GNU_SOURCE` is not specific to Linux, so binding it to that OS is incorrect.  just probe the function all the time.

meson.build
```diff
@@ -192,6 +192,15 @@ if cc.compiles(malloc_attribute_test, name : 'malloc attribute with arguments')
...
+    cc.has_header_symbol('linux/close_range.h', 'CLOSE_RANGE_CLOEXEC') and \
```
vapier, line 199: use the unistd.h API only.  that's from the C library, not the kernel UAPI.

vapier (changes requested):
rather than open-coding the logic inline, these feel like it'd be better in librc with a common API so the portability details don't get splattered.  i grok that there's currently only one user of each, but imo it's better to keep these things isolated from a maintenance pov.

meson.build
whitslack: 🤔 I didn't? Here `unistd.h` is checked for a declaration of `closefrom` regardless of OS. It *is* necessary to define a source-level compatibility macro since the declaration of `closefrom` is inside an `#ifdef __USE_MISC` guard. I define `_GNU_SOURCE` here since that's what's already defined in the source code.

meson.build
whitslack: I don't understand this comment. Read the `close_range(2)` man page. The header to include to get the definition of `CLOSE_RANGE_CLOEXEC` is `<linux/close_range.h>`.

thesamesam: It's available in `unistd.h` as well, mimicking e.g. FreeBSD. It looks like man-pages (the project) needs to be updated.

vapier: C library APIs provide all the defines/values you need.  if the C library provides `close_range()`, then it provides the various flags/bits too.

```sh
$ gcc -E -dD -D_GNU_SOURCE - <<<'#include <unistd.h>' | grep CLOSE_RANGE_CLOEXEC
#define CLOSE_RANGE_CLOEXEC (1U << 2)
$ qlist sys-libs/glibc | xargs grep CLOSE_RANGE_CLOEXEC
/usr/include/bits/unistd_ext.h:#ifndef CLOSE_RANGE_CLOEXEC
/usr/include/bits/unistd_ext.h:# define CLOSE_RANGE_CLOEXEC (1U << 2)
```

meson.build
vapier: yes, i commented on the line that's wrong -- checking the OS

meson.build
whitslack: I think these APIs hadn't yet made it into Glibc when I opened this PR. I've amended it now. Thanks.

whitslack:
> rather than open-coding the logic inline, these feel like it'd be better in librc with a common API so the portability details don't get splattered. i grok that there's currently only one user of each, but imo it's better to keep these things isolated from a maintenance pov.

Wouldn't a *lot* of the child process setup logic need to be moved into librc if we follow this line of thinking? (Not saying you're wrong, but it's a lot more work that should be done under a separate PR.)

meson.build
vapier: fwiw, afaict, the APIs were in glibc-2.34 released in Aug 2021.

it does look like the close_range(2) man page is outdated as Sam noted, and closefrom(2) is missing entirely.  Basically only docs for the kernel interfaces have been written ... none for the C library yet.

vapier:
> Wouldn't a lot of the child process setup logic need to be moved into librc if we follow this line of thinking? (Not saying you're wrong, but it's a lot more work that should be done under a separate PR.)

i'm looking at the discrete API/function level, not higher level logic.  basically librc functions as OpenRC's portability/[OSAL](https://en.wikipedia.org/wiki/Operating_system_abstraction_layer).

glancing at these 2 files, i see ioprio_set should prob be hoisted out.  other than optional functionality (e.g. pam or caps), i don't see other real ifdef usage.  what did you have in mind ?

whitslack:
> i'm looking at the discrete API/function level, not higher level logic.

Ah, okay. I do think it would make sense to move all child process spawning and setup code to a library function with a struct passed by pointer to specify parameters. This would reduce code duplication between SSD and SD and also help to prevent their getting out of sync, which keeps happening because (I guess) not many people care about SD (including myself, tbh).

If you're only talking about using librc as a syscall-level abstraction layer, then I'd also suggest moving the process niceness setting code with its Linux-specific autogroup-niceness wart.

vapier:
i agree with your assessment of the current duplication leading to desync problems.  if you wanted to do another PR to try an clean that mess up, it'd be appreciated, but i certainly wouldn't ask you to do it.

whitslack:
Hmm, given that SSD and SD have evolved in parallel but independently, I think there are a few non-trivial differences in how they set up their child processes, such that unifying them would require making some design decisions that I'd think would be more appropriate for their original author to make. I could see myself unintentionally breaking some obscure use case in one tool or the other by factoring out similar but not identical(!) code from each into a common library function.

williamh:
@whitslack  I'm definitely open to refactoring s-s-d and s-d. I just based s-d on Roy's original work on s-s-d which was based in turn on Debian's s-s-d. :-)
If you want to take a crack at this before I do, feel free to make a pr and see what you can come up with.
I'm not sure how much we need to put  in librc vs adding sources to src/shared and just linking to the objects when they are needed.

whitslack:
> If you want to take a crack at this

@williamh: Nope. I'm definitely not itching to attempt to deduplicate the child process startup code from SSD and SD. :sweat_smile: The only thing I came here to do was to eliminate some blatant syscall spam.

meson.build
thesamesam: I've filed https://bugzilla.kernel.org/show_bug.cgi?id=216951 as well.

PR #549, opened 2022-09-07
Avoid warning on grep 3.8 in hwclock

### LINKED ISSUE #548: hwclock init script logging grep warning
With the latest `grep` (3.8), the `hwclock` init script throws a warning on start in the logic to detect `--noadjfile` argument support:

https://github.com/OpenRC/openrc/blob/112b69860f2cb6059c49404abbfb4e3f22603cfc/init.d/hwclock.in#L74-L75

```
hwclock                   | * Setting system clock using the hardware clock [UTC] ...
hwclock                   |grep: warning: stray \ before -
hwclock                   |grep: warning: stray \ before -
```

This seems to have been introduced in `grep` in this commit: https://git.savannah.gnu.org/cgit/grep.git/commit/?id=e7f8e8eb1fd41b308ee10741bbd8068acc1847c2

I think replacing the affected logic with 

```
grep -q -- "--noadjfile"
```

should work.

particleflux:
Starting with grep version 3.8, the hwclock init script logs warnings about stray backslashes:

> hwclock | * Setting system clock using the hardware clock [UTC] ... 
> hwclock |grep: warning: stray \ before -
> hwclock |grep: warning: stray \ before -

This is caused by the check for existence of the `--noadjfile` argument in function `get_noadjfile()`.

Replacing the affected logic with an explicit argument denoting the pattern as such resolves the issue.

Fixes #548

### DIFF
--- a/init.d/hwclock.in
+++ b/init.d/hwclock.in
@@ -72,7 +72,7 @@ get_noadjfile()
 {
 	if ! yesno $clock_adjfile; then
 		# Some implementations don't handle adjustments
-		if LC_ALL=C hwclock --help 2>&1 | grep -q "\-\-noadjfile"; then
+		if LC_ALL=C hwclock --help 2>&1 | grep -q -e "--noadjfile"; then
 			echo --noadjfile
 		fi
 	fi


vapier:
thx bud

safiire:
Thanks this was bugging me :)

Issue #430, opened 2021-07-15
swclock saves time in tmpfs


cnoss74:
swclock saves time in /var/run/openrc/shutdowntime, hard coded. /var/run is a symlink to /run, which is on tmpfs. No wonder it doesn't work. Maybe make the file an optional argument?

williamh:
@cnoss74 The file can't be optional, because this file is used to save the shutdown time. You are correct however that this can't be saved on tmpfs.
The only other spot I can think of to save this file is in /var/lib/openrc/shutdowntime. However, that would require /var to be mounted in the initramfs or be on the same filesystem as / if you use swclock.
@vapier  Are these reasonable requirements in this situation?

vapier:
swclock needs to have state somewhere. a path under /var makes the most sense.

williamh:
I'm looking at this again along with the above proposed PR. I'm not comfortable with the PR because it changes the time on an executable.
My only concern about using /var is that will require us to wait until /var is mounted to fix the time.
Any thoughts anyone?

dwfreed:
Fixed by #565

PR #560, opened 2022-11-19
swclock: Use the same file for both saving and loading

snh:
If no path is specified on the command-line when calling `swclock`, it defaults to saving and loading the last shutdown time from a file named `shutdowntime`:
https://github.com/OpenRC/openrc/blob/9380347f042f7d294317f4420b648422817eb75a/src/swclock/swclock.c#L37

I noticed `swclock` wasn't operating as I'd expect on Alpine Linux and went investigating to find that `/etc/init.d/swclock` was calling `swclock` with no path defined when saving the time:

https://github.com/OpenRC/openrc/blob/9380347f042f7d294317f4420b648422817eb75a/init.d/swclock.in#L34

But providing the path of `@SBINDIR@/openrc-run` when loading the time:

https://github.com/OpenRC/openrc/blob/9380347f042f7d294317f4420b648422817eb75a/init.d/swclock.in#L26

Sure enough I found a file named `/run/openrc/shutdowntime` was being created, though on a `tmpfs`, and regardless, not being used when loading the time.

This PR updates `/etc/init.d/swclock` so the same file is used for both - `@SBINDIR@/openrc-run`. I chose this over using the default, as I've noted that systems such as Alpine Linux are placing `shutdowntime` in a location backed by `tmpfs`. 

What I am changing hasn't changed in many many years, so either this has been broken for a long time, or I'm misunderstanding how it works, which is entirely possible, apologies if it is the later, I thought a PR rather than Issue would be more worthwhile in-case I am on the right track! 😄 

Possibly closes https://github.com/OpenRC/openrc/issues/430.

### DIFF
--- a/init.d/swclock.in
+++ b/init.d/swclock.in
@@ -31,6 +31,6 @@ start()
 stop()
 {
 	ebegin "Saving the shutdown time"
-	swclock --save
+	swclock --save @SBINDIR@/openrc-run
 	eend $?
 }


williamh:
This is fixed in #565.

PR #572, opened 2022-12-24
Minor rc-status fixes

agvxov:


src/rc-status/rc-status.c
```diff
@@ -444,6 +446,7 @@ int main(int argc, char **argv)
 		 * be added to the list
 		 */
 		unsetenv("RC_SVCNAME");
+		/**/
```
vapier, line 449: sorry, but what is this comment line supposed to be consistent with ?

agvxov: well, i guess that one is not technically out of consistency. however the original comment above is for the unset* line only, with the prints monolithically under. a new line would have also be sufficient, but i found this easier to read while breaking no commenting convention that anyone has set beforehand.

agvxov: ignoring the comments, may i know for what reason a minor edit to make the program consistently parsable is not being accepted?

thesamesam: It's not been rejected. It just hasn't been accepted (yet). It might still be.

PRs tend to be dealt with as a batch (as a single unit) rather than cherry-picking commits.

vapier: nothing uses `/**/` style in OpenRC to break up code, and we're not going to start using it, and certainly not under the claim that it's "improving consistency" (it's actually the exact opposite of that).

if you want to add a blank line, i don't care.

agvxov: @thesamesam thank you for the clarification.
@vapier alright. then i will modify it to a blank line. to do so, is my only option to revert, correct and re-request or perhaps its possible to accomplish within this request?

thesamesam: It's fine to fix it and force push, reusing the same PR.

vapier: yeah, just rewrite the history and force push as Sam said.

if you want to do one-change-per-PR, especially for unrelated changes, that tends to be easier to review+merge.  it's a little more work on your end.

agvxov: done.

vapier:
cheers mate

Issue #432, opened 2021-07-25
"User services" in OpenRC


NyaomiDEV:
It would be nice if users could have and handle their own services and runlevels.

Nowadays a lot of applications rely on services run as the current user (for example sound servers like PipeWire and PulseAudio are required to run multiple applications that play audio in a tradiitonal desktop configuration).

For comparison, Systemd provides users with a mechanism to have their services (units) run and supervised upon login. This is, in turn, used by many desktop environments and also daemons (PulseAudio, PipeWire, etc.)

Some other init systems, like Runit, have no concept of user services, but they can be easily "tricked" to effectively handle them ([Example](https://hristos.co/blog/emacs-daemon-runit-user-service/)).

OpenRC does not have any concept of user services and it doesn't seem to be easily adaptable to handle them in a tricky way; the best an user can do is effectively modify some system services so that they are run under their user and group and then they can write a bunch of `sudo rc-service <service> start` lines on their `.profile` so that the services actually get started and supervised. This works but it is tricky and it requires the user to set a `NOPASSWD` rule for `rc-service` on their sudoers file, which makes for a security problem.

Hence, it is probably wiser to let OpenRC have a concept of user services, in a way or another; the proposed solution would be:
- to write the init scripts in exactly the same way except that they will need to be declared as user services somewhere in the scripts themselves, maybe with a `user_service` (0 or 1, false or true) variable; and that variable, in turn, would conflict with the `command_user` one;
- to have a separate folder to load init scripts from, namely `~/.config/init.d/`, where there can be user services only;
- to have a separate directory to effectively handle the OpenRC inner workings, like `/run/user/<uid>/openrc` that would more or less be of the same function and scope as the current `/run/openrc` directory;
- to have commands like `rc-service` and `rc-update` understand if they are being run as root or as user, to check whether there is a 'user' instance of OpenRC running, and to handle that one instead if they are indeed run as user.

I know this may be sloppy as a proposed solution, but I hope to find more people willing to discuss this further.

mazunki:
**Better solution**

While I agree, user services is something I am looking for, I have some notes:

What I've done on my machine is to create a runtime for my user, and add all the services i want for my user into it, and manually setting `command_user` to my user.group for the services, and hardcoding some paths. I don't know of a current way to make this cross-user compatible.

Then the only command I need to run is `doas openrc mazunki`, and it will run the services for me (including PipeWire).

**Suggestion**

[...]

PR #573, opened 2022-12-29
User Services

navi-desu:
Hello! This is the start of a draft for supporting user services.

The current "design" works by basically mimicking the layout of the system services.
This commit is the start of the proof of concept, based on what was talked on #432.

I'm opening this draft PR early to ask for some comments and a bit of guidance (since is this my first patch contributing to any project!), and to track progress of the feature.

The current implementation for paths is not that great IMO but I failed to think of a better one.
This also only includes librc, some other utilities are accessing the define'd variables and thus are broken in this patch, but I was able to get rc-update and rc-status, and to an extend, rc-service too, working as things is (granted I created the files in .config manually).
Once a better implementation is decided (or if this one is deemed ideal) I'll patch every file needed and run the tests properly.

The current implementation checks for the callers `euid` in order to toggle to user mode. If this is not desirable, we could instead introduce subcommands or flags in the commands themselves (`rc-update user add <service>` for example)

I am also still not familiar with openrc-run, so more study on that and the possible syntax for user service files is needed.

In resume, the idea is to have user-specific runtimes, limited to the users home folder, and tracked by the user (either via PAM, by running the `openrc` command manually to set the runlevel or other method).

If this feature is wanted, I'm willing to keep developing it and maintaining it for as long as necessary.

Signed-off-by: anna <anna@navirc.com>


src/shared/misc.c
```diff
@@ -43,6 +43,106 @@
...
+		snprintf(deptree_skewed_path, PATH_MAX, "%s/%s",
+				rc_svcdir_get(), RC_DEPTREE_SKEWED);
+	}
+	return deptree_skewed_path;
+}
+
+char *
+rc_krunlevel_path_get() {
+	if (!krunlevel_path) {
+		krunlevel_path = xmalloc(sizeof(char) * PATH_MAX);
```
andy5995, line 81: Hi @anna-cli. One major problem I see is that memory is allocated each time the function is called, but never freed. Or did I miss something?

anna-cli:
> Hi @anna-cli. One major problem I see is that memory is allocated each time the function is called, but never freed. Or did I miss something?

Hello @andy5995! It shouldn't allocate more than once per function (since the
pointers are global, and there's the check before allocation). But you
made me realise that is probably better to just declare the string as a
static char array inside the function!

src/shared/misc.c
anna-cli: > Hi [@anna-cli](https://github.com/anna-cli). One major problem I see is that memory is allocated each time the function is called, but never freed. Or did I miss something?

Hello [@andy5995](https://github.com/andy5995)! It shouldn't allocate more than once per function (since the
pointers are global, and there's the check before allocation). But you
made me realise that is probably better to just declare the string as a
static char array inside the function!

src/librc/librc-depend.c
```diff
@@ -22,7 +22,17 @@
...
 
-#define RC_DEPCONFIG    RC_SVCDIR "/depconfig"
+#define RC_DEPCONFIG    "/depconfig"
+
+static char *configpath = NULL;
+static char*
+depconfigpath_get() {
+	if (!configpath) {
+		configpath = xmalloc(sizeof(char) * PATH_MAX);
+		snprintf(configpath, PATH_MAX, "%s/%s", rc_svcdir_get(), RC_DEPCONFIG);
```
N-R-K, line 32: `PATH_MAX` is bit of a [nonsensical constant](https://eklitzke.org/path-max-is-tricky), openrc uses it already is a decent amount of places, but I'd prefer not to introduce it to new code.

src/librc/librc-depend.c
```diff
@@ -22,7 +22,17 @@
 
 #define GENDEP          RC_LIBEXECDIR "/sh/gendepends.sh"
 
-#define RC_DEPCONFIG    RC_SVCDIR "/depconfig"
+#define RC_DEPCONFIG    "/depconfig"
+
+static char *configpath = NULL;
+static char*
+depconfigpath_get() {
```
N-R-K, line 29: I'm not a fan of "global + get function". And global variables, especially in libraries, are just hazards waiting to happen. I'd much rather just let the caller provide the buffer + size, and make the function fill it in as needed.

Also [don't use empty argument list](https://nrk.neocities.org/articles/c-stop-writing-old-function-decel.html) for declaring a function that takes no argument. Explicitly use `void` (e.g `f(void)`) because it has different meaning in C.

src/librc/librc-depend.c
```diff
@@ -682,14 +692,17 @@ rc_deptree_update_needed(time_t *newest, char *file)
 	time_t mtime;
 
 	/* Create base directories if needed */
-	for (i = 0; depdirs[i]; i++)
-		if (mkdir(depdirs[i], 0755) != 0 && errno != EEXIST)
-			fprintf(stderr, "mkdir `%s': %s\n", depdirs[i], strerror(errno));
+	char depdir[sizeof(char) * PATH_MAX];
```
N-R-K, line 695: Aside from `sizeof(char)` being 1, this is still pretty nonsensical. Arrays are declared in amount of elements, not bytes. So `int v[5]` will allocate enough space for 5 ints.

src/librc/librc-depend.c
```diff
@@ -718,7 +731,9 @@ rc_deptree_update_needed(time_t *newest, char *file)
 
 	/* Some init scripts dependencies change depending on config files
 	 * outside of baselayout, like syslog-ng, so we check those too. */
-	config = rc_config_list(RC_DEPCONFIG);
+	char *depconf = depconfigpath_get();
+	config = rc_config_list(depconf);
+	free(depconf);
```
N-R-K, line 736: If I'm reading this correctly, this will lead to use-after-free, because `free` does not set the pointer to NULL (even if it did, it will wouldn't affect the `configpath` global, which will still be dangling). And so the next `depconfigpath_get()` will end up returning a dangling pointer.

I guess this goes on to show how global variables are hazards waiting to happen.

src/shared/misc.c
```diff
@@ -43,6 +43,94 @@
 
 extern char **environ;
 
+char *
+rc_deptree_cache_path_get(void) {
+	static char path[PATH_MAX] = { 0 };
```
N-R-K, line 48: Returning mutable `static` arrays are just as bad as globals. Either let the caller supply the buffer (better, IMO) or return a malloc-ed pointer (worse, complicates memory management).

One other possibility in this case is having some sort of `init` function which sets these up, and then letting the user just (read-only) access those directly.

N-R-K (changes requested):
Looking at https://github.com/OpenRC/openrc/issues/432 I don't see any of the maintainers making any comment on whether this feature is desired or not.

But just glancing over the diff, I see a decent amount of things which I do not like. I figured I'd leave my comments in here.

But keep in mind that I'm not a maintainer here, so for some of my "larger design related" comments, you probably should wait for a maintainer to comment before acting on it.

src/librc/librc-depend.c
navi-desu: This one I had "fixed" locally by using static path[] on depconfigpath_get(), just didn't push yet bc waiting for a more functional batch of patches.

src/librc/librc-depend.c
navi-desu: Replaced all the globals with statics, the only global that it's still there is a "static bool is_userprefix", as an attempt to generalize this patch set for more than just user services, but to work with any "prefix" file structure.

src/shared/misc.c
navi-desu: The static was supposed to be immutable (aside from a set function), but I see how that can't really be enforced. A buffer works better indeed, the only annoyance would be having to re-create the string every time, for a thing that should be set basically only once.

N-R-K: Something like this would work:

```c
struct dirs { const char *x_dir; const char *y_dir; ... };

struct dirs dirs_create(void); // allocates the strings inside the struct and returns it
void dirs_destroy(struct dirs *p); // `free`s all the stuff
```

And then the caller can call `dirs_create()` at startup/init and use that struct for the entire program and call `dirs_destroy()` before exiting.

navi-desu: While that would work, there's a lot of disconnect places calling for rc_.*_get() functions, so those strings would still be re-generated quite a lot... If I'm not mistaken, I see three solutions:

* Just create a new one in many places anyway, then we could either use a struct like that, or have buffers as arguments
* Keep with static/globals, with the risks of callers modifying them by accident
* Rework quite a *lot* of functions to accept a struct dir* as a parameter

I don't quite like the third option, so would the overhead of regenerating the same strings over and over a few times, be worth the lessen chance of an accidental bug? (Or am I not seeing a forth option?)

src/librc/librc-depend.c
navi-desu: What would be better than using PATH_MAX? I knew about it's issues, but didn't about a better solution...

N-R-K: Look into some of the `xasprintf` usage. It will allocate as much as needed.

src/librc/librc-depend.c
```diff
@@ -22,7 +22,17 @@
...
 #define GENDEP          RC_LIBEXECDIR "/sh/gendepends.sh"
 
-#define RC_DEPCONFIG    RC_SVCDIR "/depconfig"
+#define RC_DEPCONFIG    "/depconfig"
+
+static char *configpath = NULL;
+static char*
+depconfigpath_get() {
+	if (!configpath) {
+		configpath = xmalloc(sizeof(char) * PATH_MAX);
```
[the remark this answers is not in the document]
N-R-K: `sizeof(char)` has always been defined as 1. So I can't think of any valid reason to use it, other than perhaps not knowing about it or writing it out of habit.

src/librc/librc-depend.c
navi-desu: This one should be fixed now.

src/librc/librc-depend.c
navi-desu: Fixed.

src/shared/misc.c
navi-desu: Ok, I rewrote the whole thing, and taking this into consideration!

navi-desu:
Hello again! Last two days I rewrote the whole patchset, and I think it is ready for review! (although there's probably a few mistakes left still).

navi-desu:
@dwfreed @williamh @vapier (CC @thesamesam), Hello! sam recommended I tag one of you, could you check this PR? If it is a desired feature, and if the code overall looks okay!

I am currently daily driving the the patches and looking for bugs and if everything is working too.

navi-desu:
Easy way to test on gentoo:
* Add `EGIT_OVERRIDE_REPO_OPENRC_OPENRC="https://github.com/navi-desu/openrc"` to your make.conf
* `emerge -1 "=sys-apps/openrc-9999"`
* Some example scripts are available in [here](https://git.vlhl.dev/user-services/tree/), they should go in ${XDG_CONFIG_DIR}/init.d (defaults to ~/.config/init.d), config files go in ~/.config/conf.d, and the runlevels are created at ~/.config/runlevels. (System-wide services, installed by the package managers will go in /etc/init.d/user.d and /etc/conf.d/user.d)
* Then just use normal rc-* commands with the `--user` flag (or `-U`). `openrc --user default` to start the default runlevel (created automatically)

NyaomiDEV:
This is great! However, have you had any problems with services using D-Bus at all? How would you handle environment exports like that?

navi-desu:
> This is great! However, have you had any problems with services using D-Bus at all? How would you handle environment exports like that? I am asking since stuff like PipeWire requires it to properly function.

for the scripts i wrote so far, only wireplumber needs dbus, a hack to make it work is setting 'rc_env_allow="DBUS_SESSION_BUS_ADDRESS"' in .config/openrc/rc.conf, tho this won't really work when the runlevel is started via PAM (which would be the ideal way to handle that imo).

one thing i'm thinking is making a user init script for a dbus user session, that could be a step to allow the services to see dbus, tho i'm still not sure how to get the var from that into the env of other scripts

NyaomiDEV:
> one thing i'm thinking is making a user init script for a dbus user session, that could be a step to allow the services to see dbus, tho i'm still not sure how to get the var from that into the env of other scripts

That's the problem. You should preemptively export `DBUS_SESSION_BUS_ADDRESS` upon user login to a known value (hint: `unix:path=$XDG_RUNTIME_DIR/bus`) and then make `dbus.user` use that environment variable as the place to place the socket (also, being a PAM module I think it'll run early enough that desktop environments can see it, tho my brain is a bit rusty so it's just better to test that hypothesis.)

amano-kenji:
I call
```
dinitctl setenv DISPLAY XAUTHORITY WAYLAND_DISPLAY
```
in sway configuration to make dinit aware of GUI environment variables. You can also add pipewire environment variables to it.

navi-desu:
> dinitctl setenv DISPLAY XAUTHORITY WAYLAND_DISPLAY
A similar approach wouldn't work for openrc because openrc doesn't have a long running daemon, it starts the scripts and exits.

i wrote a basic dbus user script (https://git.vlhl.dev/user-services.git/tree/init.d/dbus), it works in you manually export the dbus env var on your shell or similar.

i'm thinking on the best way to automatically handle the variable. we need it to be set before pretty much everything else on user login, but i also don't want anything dbus specific on the pr.

one idea is to have a mechanism to allow a script to export a variable (thinking something like the deptree cache, that could be applied by `env_config()` and exported by the pam module). this would be general, tho i can't think of any use except dbus(?)

another (possibly simpler) idea is to have a pam module from dbus export it, and have the system-wide user-rc.conf rc_env_allow it, this depends on if upstream would support openrc, or the packagers would have to include it themselves (seems like gentoo and alpine already includes the system dbus openrc script by themselves).

that's all i've thought of so far

NyaomiDEV:
> > dinitctl setenv DISPLAY XAUTHORITY WAYLAND_DISPLAY
> > A similar approach wouldn't work for openrc because openrc doesn't have a long running daemon, it starts the scripts and exits.
> 
> i wrote a basic dbus user script (https://git.vlhl.dev/user-services.git/tree/init.d/dbus), it works in you manually export the dbus env var on your shell or similar.
> 
> i'm thinking on the best way to automatically handle the variable. we need it to be set before pretty much everything else on user login, but i also don't want anything dbus specific on the pr.
> 
> one idea is to have a mechanism to allow a script to export a variable (thinking something like the deptree cache, that could be applied by `env_config()` and exported by the pam module). this would be general, tho i can't think of any use except dbus(?)
> 
> another (possibly simpler) idea is to have a pam module from dbus export it, and have the system-wide user-rc.conf rc_env_allow it, this depends on if upstream would support openrc, or the packagers would have to include it themselves (seems like gentoo and alpine already includes the system dbus openrc script by themselves).
> 
> that's all i've thought of so far

If you want to be as general as possible, the PAM module could just set a bunch of user-defined environment variables on startup before anything else. This would probably also cover some other scenarios for people who need some environment variables set up at login, before the user shell (or graphic session) pops up. You can probably put a key=value map in `~/.config/conf.d/environment` or something like that.

navi-desu:
> The problem with this approach, however, is that if I set DBUS_SESSION_BUS_ADDRESS=$XDG_RUNTIME_DIR/bus, then the $XDG_RUNTIME_DIR environment variable in itself is not resolved to /run/user/$UID; so SOME variables have to be there... I am thinking the really base ones: HOME, USER, UID... so that at least we can "glue together" the other ones.

$HOME is set by the login shell, and so far the pam module is starting a login shell for the user to run openrc (tho what i have as a "pam module" is not very good, just two scripts called by pam_exec.so), i'd assume $USER and $UID are also set like that, but i'm not sure

> You can probably put a key=value map in ~/.config/conf.d/environment or something like that.

oh, that could work without even changing the C code, have the helper sh scripts source `/etc/conf.d/user.d/env` and `~/.config/openrc/conf.d/env`, if on user-script mode. (and we should probably make env.d too, so that package managers can install env files from packages without issues)

(right now the system-wide scripts are in /etc/{init,conf}.d/user.d, tho i'm not sure if that's good, or /etc/user.d/{init,conf}.d is better?)

NyaomiDEV:
@navi-desu

The overall problem I guess is that if you go with the key=value approach, while it's surely simple and useful, it doesn't come with the flexibility of doing dynamic environment variable allocations, since you're probably running openRC before the login shell, and you'd have no environment; just openRC executed under your user (and group? idk)

This way, each and every user needs to have its own `environment` file, with stuff like
```
DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus
```

I mean, it is doable, but probably you'd want a more generalized approach that you can just declare once in `/etc/conf.d/user.d/env`.

If you had some way to invoke shell commands inside the env file, then stuff like this could be possible; but mind you, you'd be running a non-login shell before the login shell, so no environment variables are actually available (not even PATH I guess). And it would take you much more than whatever is already used to parse `conf.d` files.
```
DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/$(id -u)/bus
```

So, the best course of action in my opinion is just to make some default environment variables available, possibly a subset of the really common ones that you can safely get. Of course, those "special" environment variables should be discarded once the environment is set, unless there's a reason to keep them around for usage in the user service files themselves (which is something you might want to have).
```
#!/sbin/openrc-run

pidfile="/run/user/$UID/bus.pid" # as an example
command="/usr/bin/dbus-daemon"
command_args="--session --syslog --nofork --address ${DBUS_SESSION_BUS_ADDRESS}"
command_background=yes
```

navi-desu:
> since you're probably running openRC before the login shell, and you'd have no environment; just openRC executed under your user (and group? idk)

i don't see why we couldn't open a login shell to run `openrc --user` to be honest, is there some consequence i don't know of?

> `DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/$(id -u)/bus`

this won't propperly work. on my system XDG_RUNTIME_DIR is set to /tmp/$(id -u)/ for example. usually elogind sets up that variable, but in systems without (like mine), it's not guaranteed to be that :/

my current idea is having the environment that runs the init script, if in user mode, source /etc/conf.d/user.d/env.d/* (oh gosh that's a lot of .d)

on this pr, the function that clears the environment before running the script makes exceptions to HOME, XDG_RUNTIME_DIR, and XDG_CONFIG_HOME and RC_USER_SERVICES. everything else is cleared, and then the many RC_* variables are set.

considering we can open a login session from pam, those variables should be usable in any env.d file defined

NyaomiDEV:
> i don't see why we couldn't open a login shell to run openrc --user to be honest, is there some consequence i don't know of?

I mean, you could; but would you `exec` it? Problem is, what if the shell is non-POSIX compliant? eg. `fish`. Sure you could probably just run `sh` as login each time, but a solution written in C code is more elegant IMHO.

Also, if you're willing to open a login shell, you could just inherit all the environment variables from there, eg. the user could just set what they want in `~/.profile` and those changes would be picked up.

navi-desu:
> I mean, you could; but would you exec openrc at the end of the shell script? Problem is, what if the shell is non-POSIX compliant? eg. fish. Sure you could probably just run sh as login each time, but a solution written in C code is more elegant IMHO.

ok i totally forgot about non posix shells. the current script itself is !#/bin/sh, but even then i'm not sure if there's a for sure way to run it with a login shell (seems like su/runuser is not posix either too)

> Also, if you're willing to open a login shell, you could just inherit all the environment variables from there, eg. the user could just set what they want in ~/.profile and those changes would be picked up; then, it's just a matter of rc_env_allowing what the user wants.

ideally we don't want to depend on the user say, having to set DBUS_* for the dbus and wireplumber scripts to work.

the problem is two rn:
The env needed by services -> those can be set by the packager in /etc/conf.d/user.d/env.d (for example) and by the user in .config/openrc/env.d (or conf.d/env.d for consistency?) -- i like this solution

The env needed by openrc --user itself (XDG_RUNTIME_DIR and XDG_CONFIG_HOME), we can't really use .config/openrc/env.d for that because well, if XDG_CONFIG_HOME points somewhere else, using a login shell could solve that, because as you said, .profile, .bash_profile, .zshprofile, etc, would be loaded... then non-posix shells make this kinda unfeasible? even if we say that XDG_RUNTIME_DIR needs to be set in the PAM env (by elogind, or a pam module, or whatever), the CONFIG_HOME would still be missing... -- here i'm not sure what to do here tbh

navi-desu:
>You could take a look at using turnstile to just not have to worry about this.
https://github.com/chimera-linux/turnstile
At this point let's just not reinvent the wheel.

i'm not sure about adding a new (and c++ even) dependency, tho... "The backend is invoked as a shell script, specifically as a login shell. [...] freshly initialized, and the shell profile is also sourced" -- we can look at how they do that without tripping over non-posix shells

NyaomiDEV:
> i'm not sure about adding a new (and c++ even) dependency, tho...

As I understand, that is init-agnostic, so another project could just glue openrc --user to that and it'll just work (= distro maintainers can just ship another package while the main openrc doesn't have any explicit dependency on turnstile)

> we can look at how they do that without tripping over non-posix shells

That's also a nice idea.

navi-desu:
those last changes:

* reworked the `rc_user_*dir()` functions to `rc_*dir()`, so that the logic of `if (rc_is_user())` doesn't need to be repeated in many places.

* added a new binary for the pam module, it starts the user login shell, with `"-c", "openrc --user <runlevel>"`, it is expected that the user's login shell supports -c (which i didn't find one that didn't) it also sets specific environment variables to the lockfile

* added a lockfile function to rc.c, this is to avoid changing runlevels back to `default` after the first login, and only changing to the `none` runlevel on the last logout.

currently this patch should be capable of automatically starting scripts on the first login, and stopping them on the last logout.

the only missing feature would be a way to deal with dbus and services that require it, since the `DBUS_SESSIONS_BUSS_ADDRESS` needs to be set for each script, and for the user's environment after login. I'm not sure the best way to handle it since as far as i know dbus is the only service that needs such a thing, so my ideas of a environment exporting system could be overkill for just dbus to use it.

flexibeast:
i've been pointed to this PR by @thesamesam as Relevant To My Interests. :-)

The most immediate reason for my interest is trying to get the OpenRC 'emacs' service working.

i currently have the following in my .zlogin:

    if [ ! $(pgrep -x ssh-agent) ]; then
        echo -n "Starting ssh-agent .... "
        eval `ssh-agent -s`
    fi
    if [ ! $(pgrep -x devmon) ]; then
        echo -n "Starting devmon ...."
        devmon --sync 1>>~/logs/devmon.log 2>&1 &
    fi
    if [ ! $(pgrep -f 'dbus-daemon --session') ]; then
        echo -n "Starting D-Bus session bus ...."
        dbus-daemon --session --fork --print-address 4 --print-pid 5 4>!~/.session-bus-address 5>!~/.session-bus-pid
        echo " $(cat ~/.session-bus-pid)"
        export DBUS_SESSION_BUS_ADDRESS="$(cat ~/.session-bus-address | tr -d '\n')"
    fi
    if [ "${E}" = 'yes' ] && [ ! $(pgrep -x emacs) ]; then
        echo -n "Starting emacs ...."
        emacs --daemon 1>>~/logs/emacs.log 2>&1 &
    fi

This ensures that my environment  has the SSH_AUTH_SOCK, SSH_AGENT_PID and DBUS_SESSION_BUS_ADDRESS env vars set appropriately before the Emacs server, my X session (via `startx`) and any other programs get started.

However, the OpenRC 'emacs' service starts before login, so in order for the Emacs server to 'see' the DBUS_SESSION_BUS_ADDRESS, i had to put together the workaround i've described in the "Warning" box in [this section of the Emacs page on the wiki](https://wiki.gentoo.org/wiki/GNU_Emacs#OpenRC) (dealing with an Emacs bug along the way).

What i gather from #432 and this PR is that the intention is to start user services via PAM, and that it's not yet clear how to set things up so that a D-Bus session bus gets started, and the DBUS_SESSION_BUS_ADDRESS set appropriately, for all services and programs in the login session. As the above snippet indicates, in my own case, `ssh-agent` also needs to have been started, and SSH_AUTH_SOCK and SSH_AGENT_PID set appropriately, in the environment as well.

One ugly kludge i'd been trying to work out how to implement with the current version of OpenRC is a way of getting a service to essentially background itself, letting subsequent services start, while it polled to check whether it could now start (e.g. by checking whether DBUS_SESSION_BUS_ADDRESS had become set).

In the context of this PR, i'm wondering about the feasibility and/or ugliness of having the user service equivalent of a 'boot' runlevel which sets up the necessary environment for all subsequent runlevels (although i might have missed where this has been covered).

Please let me know if i'm misunderstanding anything?

navi-desu:
> intention is to start user services via PAM, and that it's not yet clear how to set things up so that a D-Bus session bus gets started

as of right now, dbus is also a user script, started by openrc, so by having the emacs script `use dbus` it should start the dbus session before emacs does

> and the DBUS_SESSION_BUS_ADDRESS set appropriately, for all services and programs in the login session

and this is the problem

i thought of a system that could allow scripts to declare env. variables to save, and ever subsequent script would source them, or maybe just scripts that depend on foo would source foo's var, but since dbus was the only thing in my mind for it, i kinda let that simmer

another simple (but kinda ugly) solution would be similar to what you're doing, save DBUS_SESSION_BUS_ADDRESS (and the SSH_*) vars to files, then source them in scripts that need them.

the question is more: is this env feature fit to be a common feature for openrc-run.sh, or it's better for us to make this hacky workaround in the few scripts that use it?

with both dbus and ssh-agent making some use of it, i'd say that it may be fit to be in openrc-run? not sure, i'd like opinions

NyaomiDEV:
I am stating the obvious so that it doesn't become so obvious that we forget it: `DBUS_SESSION_BUS_ADDRESS` has to be declared for a lot of stuff to work properly (GNOME, KDE Plasma, MPRIS2 compatible players, etc. etc.)

The problem here is to have a `dbus` user service that, in the end, no matter how and why, gets its variable to the user session after it's started.

Sure, going with Systemd-style D-Bus address is the easier choice since we already know it's going to be `/run/user/<uid>/bus`, however some distros currently ship `/tmp/dbus-XXXXXXXXXX` as a bus address, and we would probably like to support both scenarios.

navi-desu:
> however some distros currently ship /tmp/dbus-XXXXXXXXXX as a bus address, and we would probably like to support both scenarios.

afaik the /tmp/dbus-XXXXXXXXXX is generated by dbus-launch, and when using the dbus user script the user shouldn't need to run dbus-launch at all (since openrc would handle that). setting the variable to a define value we know should work just the same

flexibeast:
> setting the variable to a define value we know should work just the same

What are the security implications of this? i.e. that an arbitrary program would have access to the session bus on any system with OpenRC user services?

navi-desu:
> What are the security implications of this? i.e. that an arbitrary program would have access to the session bus on any system with OpenRC user services?

the bus socket would be in XDG_RUNTIME_DIR, a 0700 dir so only your user would have access to it.

wrt any *program*, that's already the case no? any program your user runs has access to the dbus socket. under systemd it defaults to `${XDG_RUNTIME_DIR}/bus` (so `/run/user/1000/bus`), we would default the var for the same place

flexibeast:
Well, upthread @NyaomiDEV noted:

> some distros currently ship /tmp/dbus-XXXXXXXXXX as a bus address, and we would probably like to support both scenarios.

i'm thinking of programs run under other user logins on the system, where those programs will know the session bus address for the currently active seat.

navi-desu:
> i'm thinking of programs run under other user logins on the system, where those programs will know the session bus address for the currently active seat.

other users won't have access to the bus of the current active seat (unless they're root, duh). 

other logins of the same user would tho, i don't know how much of an issue is you in another login having access to your own bus (i could investigate systemd to see what they do, but i'm pretty sure they only use a single bus at /run/user/1000/bus)

but considering a system where variables can be set by init scripts at runtime, we could also randomize dbus addresses i think? i don't feel like it's much of a necessity but could be

there's also the question of if we want user services to auto-start for non-seat (or non-local) logins or not

flexibeast:
> other users won't have access to the bus of the current active seat

Oh okay, so D-Bus itself checks whether messages to a particular bus come from an authorised user?

> other logins of the same user would tho

Yeah, this wasn't the issue i had in mind; i can't see how there'd be any problems from a security perspective, but maybe things could get badly confused if the wrong session bus is used by a login session.

> we could also randomize dbus addresses i think? i don't feel like it's much of a necessity but could be

\*nod\* i'd certainly be interested to know the extent to which it's a *necessity*, i.e. why things like e.g. `dbus-run-session` make the choice to derive the address from a GUID (or whatever it technically is).

> there's also the question of if we want user services to auto-start for non-seat (or non-local) logins or not (maybe let the user define a default runlevel for local only logins, and a runlevel specific for remote logins?)

My gut response to the auto-start question, based purely on my own use-case, is "no", but maybe there are arguments for this to be user-configurable. Your idea to have distinct runlevel for local vs remote logins sounds good to me.

navi-desu:
> Oh okay, so D-Bus itself checks whether messages to a particular bus come from an authorised user?

i don't know actually if dbus does (but considering that the dbus socket `dbus-launch` creates is user:user 777, i'm pretty sure it does otherwise it'd be a problem).

it wouldn't be a problem in our case because ${XDG_RUNTIME_DIR:-/tmp/openrc/${UID}} is where ${RC_SVCDIR} would be set to for a user, and where we'd put the dbus socket. and it's standard that the runtime directory for a user is created in user:user drwx------ (0700 i think?), so nobody but the user would have access to the socket anyway

> i'd certainly be interested to know the extent to which it's a necessity, i.e. why things like e.g. dbus-run-session make the choice to derive the address from a GUID

probably to avoid conflicts, as `dbus-launch` creates the socket in `/tmp/dbus-XXXXXX`, so multiple dbus sessions would create multiple dbus sockets in /tmp

> Your idea to have distinct runlevel for local vs remote logins sounds good to me.

i have a simple and clean idea on how to make this work with pam actually. will add to todo because it's basically nothing to support and even better than the current impl imo)

navi-desu:
added a crude way to export and import variables between services. on default_start, a service with `${export_vars}` defined will export any environment variables listed, to a file in `${RC_SVCDIR}/env/${RC_SVCNAME}`.

it will also `sourcex` every file in there (`${RC_SVCDIR}/env/*`). altho ideally we'd only import the variables set by the services we depend on.

this was one of the generic ideas for environment sharing, another, less dynamic one, could be similar to systemd's environment.d, which consists of just sourcing the files in `/etc/environment.d/*` and `${XDG_CONFIG_HOME}/environment.d/*`. this would require the package with the init.d script also ship a environment file in the specific directory, but also would allow the user to easily define any specific variable to be set for scripts (tho on the current method, we can also add something like `${rc_env}` in conf.d to allow the user to do the same)

i'm still not sure which approach would be better

only thing missing for a seamless env experience would be getting pam to load the environment for the user after it starts the initial services.

an example of a init script, dbus, with environment export:

```
#!/sbin/openrc-run

DBUS_SESSION_BUS_ADDRESS="unix:path=${XDG_RUNTIME_DIR:-${RC_SVCDIR}}/bus"
pidfile="${XDG_RUNTIME_DIR:-${RC_SVCDIR}}/${RC_SVCNAME#*.}"
command="/usr/bin/dbus-daemon"
command_args="--session --syslog --nofork --address ${DBUS_SESSION_BUS_ADDRESS}"
command_background=yes
export_vars="DBUS_SESSION_BUS_ADDRESS"
```

navi-desu:
ok so apparently loading the env generated by that from pam can be unsafe, if i got it right, we can't guarantee that openrc-pam is called last, so user-controlled variables would be loaded by pam, and would persist through the other modules, potentially affecting them...

this complicates things a bit...

simonvanderveldt:
First of all thanks a lot for working on this! Really nice to see this happening :+1: 

> as for the security of it, openrc-pam will only load the variables that the sysadmin allows in /etc/conf.d/user.d/env_whitelist (and env_whitelist.d for packagers)

Do you know how this compares to how systemd handles passing environment variables? It seems like there are quiet a few ways in systemd to configure how to and which environment variables to pass (https://wiki.archlinux.org/title/systemd/User#Environment_variables) but I see no mention of any whitelisting or approval from a sysadmin.

I was also wondering if maybe it makes sense to not implement anything specific with regards to handling user environment variables and instead just let this be defined by the process that starts the user OpenRC. So when someone wants to run some user services for their graphical session they just launch the user OpenRC via their login manager/dbus-launch/etc so it gets passed the environment variables from there automatically.

navi-desu:
@simonvanderveldt 

> Do you know how this compares to how systemd handles passing environment variables? It seems like there are quiet a few ways in systemd to configure how to and which environment variables to pass (https://wiki.archlinux.org/title/systemd/User#Environment_variables) but I see no mention of any whitelisting or approval from a sysadmin.

i don't know the specifics, but it's a thing similar file being loaded, tho, the difference to us, systemd does not let you set environment variables on the login shell via pam, because variables set in one pam module can interfere with another. systemd doesn't need this because systemd itself starts dbus, and as of right now, we need this env variable mechanism for dbus (tho it can also be useful for ssh-agent and gpg-agent).

to mitigate the pontential leaking, we limit the possible variables that a script can to the user's shell pam (altho, variables set by scripts are still visible by other scripts in pam, just not the user's shell)

> I was also wondering if maybe it makes sense to not implement anything specific with regards to handling user environment variables and instead just let this be defined by the process that starts the user OpenRC. So when someone wants to run some user services for their graphical session they just launch the user OpenRC via their login manager/dbus-launch/etc so it gets passed the environment variables from there automatically.

the issue with that is two, first openrc clears variables when running, so for example, $DBUS_SESSION_BUS_ADDRESS wouldn't be visible to other scripts in this possible `gui` runlevel,

second i wanted to make user scripts a bit more self reliant. so if a script `depend()`s on dbus, starting the `dbus` user script should set the variable for that script.

in this scenario, we could have user scripts loading variables by themselves tho, a script for `sway` exports WAYLAND_DISPLAY, for example. but i still don't know how openrc will interact with the likes of a compositor or window manager being started from it, since it expects to run the script and exit. i need to test this.

navi-desu:
i'm kinda re-writing most of this patch-set once more

will update the pr soon

navi-desu:
openrc --user 3.0

the patch is a lot simpler, does a lot less (and that's good)

first all environment exporting/handling code got removed, instead we'll just pass that burden into the service package. /etc/profile.d already gives program a way to export variables into the user's environment, so a package can simply use that without openrc needing to do extra work.

secondly the rc_*_dir api got simplified to two functions, that take a subpath. this was done because that was basically all they were used for, and imo, it's a nicer api

and lastly, simpler openrc-pam, since no env code is needed. also instead of counting logins ourselves, this new code relies on utmpx. i am still unsure which is the better approach

improved how openrc-pam handles xdg_rundir when it's not set as well.

it was suggested using /etc/profile.d/00-openrc.sh or similar to launch openrc --user, while i like the idea, it lacks a way to stop openrc on log-out, as well as lacking a way to create the rundir properly (since it would require elevated privs)

any suggestions or comments about those changes is appreciated!

NyaomiDEV:
> /etc/profile.d already gives program a way to export variables into the user's environment, so a package can simply use that without openrc needing to do extra work.

While I agree with this sentiment, /etc/profile.d/*.sh is usually sourced by /etc/profile, which only is sourced in login shells. Basically it all happens after PAM starts `openrc --user`, so the question I raise is, how do we get to the exported variables before they are even exported? I'd rule out `echo`ing stuff to source from a script in `/etc/profile.d` dynamically.

navi-desu:
> so the question I raise is, how do we get to the exported variables before they are even exported

/etc/init.d/user/dbus:
```
DBUS_SESSION_BUS_ADDRESS="unix:path=${XDG_RUNTIME_DIR}/bus"
extra_commands="socket"
socket() {
    echo DBUS_SESSION_BUS_ADDRESS=${DBUS_SESSION_BUS_ADDRESS}
}
```

/etc/profile.d/50-dbus-openrc.sh:
```
if [ -z ${DBUS_SESSION_BUS_ADDRESS} ]; then
    export $(rc-service -Us dbus socket)
fi
```

> Also, /etc/profile.d/ stuff is susceptible to shell changes. What if someone uses fish?

then they gotta do that themselves. many things relies on /etc/profile{,.d,.env}, if you're using fish as your login shell, you're gonna need to fix the system integrations yourself.

distros can also just ship a fish version of the script snippet up there, it's their choice, the good thing about this is, openrc won't really care about how you export that.

NyaomiDEV:
> /etc/init.d/user/dbus:
> 
> /etc/profile.d/50-dbus-openrc.sh:

Honestly, it's pretty good and I didn't think of that initially. I wonder if there's a way to standardize an extra command, for example `env`, so that we can treat it as we would treat a `.env` file.

For that matter, a profile script that retrieves the list of running services and calls `env` on them would basically be just one script and it would be fine enough.

> why?

Dumb brain loves static stuff.

> then they gotta do that themselves. many things relies on /etc/profile{,.d,.env}, if you're using fish as your login shell, you're gonna need to fix the system integrations yourself.
> 
> distros can also just ship a fish version of the script snippet up there, it's their choice, the good thing about this is, openrc won't really care about how you export that.

Fair; though a lot of people use `fish` nowadays so probably it's just easier to just write a script, even for distro maintainers.

WhyNotHugo:
What's the goal of using a pam module instead of a direct approach (like `.profile`) to start the service manager?

A pam module puts any form of reconfiguration outside of the reach of regular users. E.g.: it impossible for a user to run something _before_ the service manager starts, temporarily disable the service manager, etc.

NyaomiDEV:
> What's the goal of using a pam module instead of a direct approach (like `.profile`) to start the service manager?
> 
> A pam module puts any form of reconfiguration outside of the reach of regular users. E.g.: it impossible for a user to run something _before_ the service manager starts, temporarily disable the service manager, etc.

.profile is sourced by the login shell, however, you don't know which login shell will be spawned, and you don't even know if that'll be a login shell. If it's a login session, then stuff like dbus and pipewire have to necessarily be started before the login session.

Of course, if you don't use a display manager you could put that in .profile, but that's not everyone's use case.

navi-desu:
> What's the goal of using a pam module instead of a direct approach (like `.profile`) to start the service manager?

with .profile, we can't ever stop services on the last log-out, as there's no system-wide "/etc/logout.profile" or something, we also can't set up xdg_runtime_dir for users without a setuid binary.

i'm not super happy with pam either, to be honest. as is right now, it's tending to a bit of scope creep, and i'm unsure if openrc should be doing all that

i'm still experimenting with alternatives, but the goals are:
- minimal initial user setup
i.e. all most users should need to do is `rc-update --user add foo`
let users with simple needs just start dbus and pipewire and forget about it
while also leaving more complex setups for those who want it
- starts at the first log in, stops at the last log out
- a way to set a proper session for the user, like xdg_runtime_dir (i'm unsure of the best place to do this)

> A pam module puts any form of reconfiguration outside of the reach of regular users. E.g.: it impossible for a user to run something _before_ the service manager starts, temporarily disable the service manager, etc.

there should be an option to disable auto-launching from the user-side, yeah.
probably a `rc_user_autolaunch=no` in ~/.config/openrc/rc.conf

then everything is handed over to the user

flexibeast:
@navi-desu:

> a way to set a proper session for the user, like xdg_runtime_dir (i'm unsure of the best place to do this)

Well, fwiw, last year i created an ebuild for [pam_xdg](https://www.sdaoden.eu/code-pam_xdg.html): https://bugs.gentoo.org/908431

WhyNotHugo:
If services are started via pam, they will not inherit the environment defined in `.profile`, which can also lead to issues. E.g.: I set `GNUPGHOME` is `.profile`, but if services are started before that is sourced, then gpg will be using the wrong directory. It's perfectly reasonable that someone may define `XDG_CONFIG_HOME` in `.profile`, and their services will use the wrong config directory.

> .profile is sourced by the login shell, however, you don't know which login shell will be spawned, and you don't even know if that'll be a login shell

If a user changes their shell to something that _doesn't_ source `.profile` to avoid starting openrc, why should they be disallowed to do so?

> If it's a graphical login session, then stuff like dbus and pipewire have to necessarily be started before the login session.

These services can be started by a program executed by the user, rather than by something managed by root and enforced via PAM. They do not _need_ to start before the login shell; I personally run `dbus`, `pipewire` and others as children of my compositor and there are no issues with that.

navi-desu:
> If services are started via pam, they will not inherit the environment defined in .profile, which can also lead to issues.

the pam module, as is right now, starts a login shell, with `-c`, so all usual env files should be sourced (by the shell)

i'm also looking at instead of starting it directly, have a system wide service `user.foo` that starts it for `foo`, and then pam just starts that. this service would call `su` with a login shell as well.

> If a user changes their shell to something that doesn't source .profile to avoid starting openrc, why should they be disallowed to do so?

the user will have an option to not auto-start services. saying that "not sourcing .profile" is the solution for that is silly, because, what if they want .profile, and just not openrc --user?

> These services can be started by a program executed by the user, rather than by something managed by root and enforced via PAM. [...] 

you then lose on the main point of a service manager, that is having it manage, auto-restart on failure, and more, your services that run as your user.

> I personally run dbus, pipewire and others as children of my compositor and there are no issues with that.

which does not work for DEs like kde plasma which need dbus running before it starts, and display managers like SDDM don't source .profile at all, those services need to start before the gui does.

and another main issue with .profile for this, is still, we can't stop them. the user logs out of all their sessions, the services will keep running.

and the user still has full control, by disabling autostart on their rc.conf (i'm still to implement this, but it'll be an option), or by adding services they want to start before login in the default runlevel, then stacking that into another runlevel with extra, post-login services, and calling that runlevel manually

navi-desu:
> Why implicitly enforce stopping of services instead of, again, letting users do it explicitly? Can users opt-out of all their processes being killed?

> Nice. A bit tangential but: have you contemplated a runlevel that gets executed even when a user does not log in?

both questions are kind of one in the same, about managing services outside of a session

for the first, well, yes, because if you're not using pam to begin with, we can't stop the runlevel automatically

the reason for pam is just, convenience for most users that just want, say, dbus and pipewire working when they're using their system. starts at login, stops at logout to preserver resources, and totally optional

starting a runlevel at boot tho, is planned, but so far as a system-wide setting, aka, you'd need to be root to set it up. i'm unsure if we can do this in a sane way without root, nor know if it's even desirable to

tho i'm thinking now, the bulk of user services is working, i wonder, i could drop the pam stuff from this pr, and ask for review/tests on the core functionality, maybe merge this and then work on pam/auto-start after?

navi-desu:
superseded by #723 for the sole reason of github doesn't let me change the source branch of a pr (seriously gh??)

PR #585, opened 2023-01-23
rc: use /etc/hostname if available

vapier:
This is significantly less overhead than an init script.


thesamesam (approved):

src/openrc/rc.c
```diff
@@ -473,12 +473,31 @@ handle_signal(int sig)
...
 }
 
+static void
+do_early_hostname(void)
+{
+	/* Set hostname if available */
+	char *buffer = NULL;
+	size_t len;
+
+	if (rc_getfile(RC_SYSCONFDIR "/hostname", &buffer, &len)) {
```
dwfreed, line 483: The `hostname` tool from net-tools on Linux has a `-b` flag that will fall back to setting the hostname to `localhost` if the file does not exist or is empty; this would probably be a good idea to implement, as on Linux, the hostname could be an empty string otherwise.  `hostname` also allows for comment lines (line starts with `#`) in `/etc/hostname`, which are ignored; this could also be useful.

dwfreed (changes requested):
This does not handle the case where the hostname is being set via `/etc/conf.d/hostname` which the initscript allows.  A deprecation warning should be added to the initscript, and a louder one should be added for users of the conf.d method.

PR #587, opened 2023-01-25
Miscellaneous warning fixes

thesamesam:


src/libeinfo/libeinfo.c
```diff
@@ -677,7 +677,7 @@ ewarn(const char *EINFO_RESTRICT fmt, ...)
 void
 ewarnx(const char *EINFO_RESTRICT fmt, ...)
 {
-	int retval;
+	_unused int retval;
```
vapier, line 680: the warning looks correct.  why don't we delete it.

src/shared/_usage.h
```diff
@@ -44,6 +44,12 @@
 	case 'q': case_RC_COMMON_getopt_case_q; break;			      \
 	default:  case_RC_COMMON_getopt_default; break;
 
+#if __GNUC__ > 2 || defined(__INTEL_COMPILER)
```
vapier, line 47: this looks like it belongs in `helpers.h` instead

src/librc/librc.c
```diff
@@ -755,8 +755,10 @@ rc_service_mark(const char *service, const RC_SERVICE state)
 					    RC_SVCDIR "/%s/%s",
 					    rc_parse_service_state(RC_SERVICE_WASINACTIVE),
 					    base);
-					if (symlink(init, was) == -1)
+					if (symlink(init, was) == -1) {
+						free(init);
```
vapier, line 759: this looks correct, but from scanning the rest of this func, seems like there's other places this leaks that need fixing too ?

vapier (changes requested):
i merged some, so you'll want to rebase

src/librc/librc.c
thesamesam: sorry, I can't spot it - could you point one out?

vapier: sorry, i was looking at the file w/out your patch applied, and thought it was fixing an earlier line

src/openrc-shutdown/broadcast.c
```diff
@@ -33,6 +33,7 @@
 
 #include "broadcast.h"
 #include "helpers.h"
+#include "_usage.h"
```
vapier, line 36: don't need this include anymore

vapier (changes requested):
merged most things.  the 2 patches left i think you can/should squash into one.

src/librc/librc.c
thesamesam: np

src/openrc-shutdown/broadcast.c
thesamesam: ugh, I thought I'd got that - thanks

vapier:
thx bud

thesamesam:
cheers!

Issue #589, opened 2023-01-27
rc: do not call free/malloc in signal handlers


martinetd:
I've spent part of this week chasing after a dead lock in "openrc shutdown" (running shutdown or reboot in a loop hangs once every few days in a `malloc()` call):
https://www.openwall.com/lists/musl/2023/01/23/1

It turns out the problem is almost certainly that we're doing a free in our SIGCHLD handler:
```
static void handle_signal(int sig) {
...
                /* Remove that pid from our list */
                if (pid > 0)
                        remove_pid(pid);
...
static void remove_pid(pid_t pid) {
...
                    LIST_REMOVE(p, entries); 
                    free(p);
```

free/malloc are not signal-safe, in particular musl libc does not take any lock as long as the process does not create threads -- if there had been we would likely just have seen a deadlock instead -- so if we have a stack like the following (bottom to top):
```
openrc functions...
malloc()
handle_signal()
free()
```
the free can be called while the malloc state is inconsistent and further corrupt mallocng's internal state.
(Yes that probably wouldn't cause problems without rc_parallel, call me a red neck)

Anyway, there are a few more mallocs in the `SIGUSR1`'s `rc_plugin_run` but as a first approximation I'd like to make that `remove_pid` delay its free. The most straight-forward solution I can think of would be to have the signal handler not call remove_pid, but instead move the list link to a "to free later" list that can be checked once in a while during the main thread.
That still is quite a bit of overengineering, but I cannot think of anything else that would be safe short of just leaking the memory. Would that be acceptable?

I'll open a PR early next week unless someone beats me to it.

N-R-K:
Just a quick look, and there seems to be more issue than just the `free` call. A couple other non-signal-safe function that I see being called:

* `fprintf` (via `eerror`)
* `malloc` (via `xasprintf`)
* `ioctl`
* `exit`
* `fclose`, `setenv` and a whole bunch more inside `rc_plugin_run`

N-R-K:
A couple of those seem to be low hanging fruits. Some of them however I presume would require significant effort to make async-signal-safe.

If there's interest in fixing these upstream, then I can look into some of them.

- - -

As for the original reported issue, marking the item as "to be freed" and freeing them elsewhere seems like it would be the least intrusive change (assuming leaking the memory away isn't acceptable).

Another solution could be to pre-allocate a buffer for backing the list and avoid malloc/free when doing list insertion/removal entirely (or at least reduce chances of it occurring inside the sig-handler).

[...]

PR #590, opened 2023-01-27
openrc: avoid unnecessary malloc inside sig-handler

N-R-K:
malloc (called by xasprintf) is not async-signal-safe. beside, the string here is constant, so there's no need to malloc it all.

eerrorx isn't async-signal-safe either (due to calling fprintf and exit) but consequence of them are _typically_ not as grave as calling malloc while it's internal state is inconsistent.

Bug: https://github.com/OpenRC/openrc/issues/589

### DIFF
--- a/src/openrc/rc.c
+++ b/src/openrc/rc.c
@@ -382,7 +382,7 @@ static void
 handle_signal(int sig)
 {
 	int serrno = errno;
-	char *signame = NULL;
+	const char *signame = NULL;
 	pid_t pid;
 	RC_PID *pi;
 	int status = 0;
@@ -414,15 +414,15 @@ handle_signal(int sig)
 
 	case SIGINT:
 		if (!signame)
-			xasprintf(&signame, "SIGINT");
+			signame = "SIGINT";
 		/* FALLTHROUGH */
 	case SIGTERM:
 		if (!signame)
-			xasprintf(&signame, "SIGTERM");
+			signame = "SIGTERM";
 		/* FALLTHROUGH */
 	case SIGQUIT:
 		if (!signame)
-			xasprintf(&signame, "SIGQUIT");
+			signame = "SIGQUIT";
 		eerrorx("%s: caught %s, aborting", applet, signame);
 		/* NOTREACHED */
 	case SIGUSR1:


N-R-K:
Since there's interest in fixing the sig-handler issue - I plan on fixing and submitting the fixes one at a time. But if one large PR is preferred instead - let me know and I can do that too.

src/openrc/rc.c
```diff
@@ -413,17 +412,13 @@ handle_signal(int sig)
...
-		if (!signame)
-			xasprintf(&signame, "SIGINT");
-		/* FALLTHROUGH */
+		eerrorx("%s: caught SIGINT, aborting", applet);
+		/* NOTREACHED */
 	case SIGTERM:
-		if (!signame)
-			xasprintf(&signame, "SIGTERM");
-		/* FALLTHROUGH */
+		eerrorx("%s: caught SIGTERM, aborting", applet);
```
vapier, line 418: you can keep the previous logic without duplicating the errorx call

```
const char *signame = NULL;
...
if (!signame)
  signame = "SIGINT";
...
```

vapier:
agreed that this is def not fixed, and we can't use errorx for a bunch of reasons (fprintf/syslog), but this prob is marginally better by not directly diving into the memory allocator

thesamesam:
Bit frustrated there's no annotations for this class of problem other than IIRC Valgrind being able to detect it at runtime.

Someone laid out the general recipe at https://news.ycombinator.com/item?id=16754394, not that it went anywhere.

src/openrc/rc.c
N-R-K: > you can keep the previous logic without duplicating the errorx call

Is that worth it though? It's more code than just calling errorx directly (and simpler to follow since it doesn't rely on fallthroughs).

But I can update the PR if that's your preference.

N-R-K:
> Bit frustrated there's no annotations for this class of problem other than IIRC Valgrind being able to detect it at runtime (I don't recall if any of the sanitizers can).

`clang-tidy` has a check for this IIRC (https://clang.llvm.org/extra/clang-tidy/checks/bugprone/signal-handler.html) - but for whatever reason it "doesn't seem to be working" on my end (I'm most likely doing something wrong here).

src/openrc/rc.c
vapier: your version requires duplicating error strings & function calls.  the current version involves setting a pointer with some cascading NULL checks.  imo that is better.

N-R-K: Updated in 83bba03

vapier:
thx bud

PR #591, opened 2023-01-28
swclock: fix codeql warning and upgrade to futimens

N-R-K:
this was reported by codeql's scan as a TOCTOU bug. while that's true in theory, i don't believe it would've had any practical effect.

a better justification for this change might be the fact that it upgrades from `utime` (which is depreciated by POSIX [0]) to `futimens`.

[0]: https://www.man7.org/linux/man-pages/man3/utime.3p.html#FUTURE_DIRECTIONS


vapier:
seems reasonable, thanks

PR #594, opened 2023-01-29
rc: signal safety improvement

N-R-K:
### rc: avoid calling free inside SIGCHLD handler

`free` is not async-signal-safe and calling it inside a signal handler
can have bad effects, as reported in the musl ML:
https://www.openwall.com/lists/musl/2023/01/23/1

the solution:

- keep track of weather remove_pid() is being called from inside a
  signal handler or not.
- if it's inside a signal handler then DO NOT call free - instead put
  that pointer into a "to be freed later" list.
- if it's not inside a signal handler then take the "to be freed later"
  list and free anything in it.

### rc: block SIGCHLD during pid list operations

the pid list will be accessed inside the SIGCHLD signal handler. so we
must ensure SIGCHLD handler doesn't get invoked while the list is at an
inconsistent state making it unsafe to interact with.

### rc: use LIST_FOREACH_SAFE in cleanup()

according to the linux manpage, the "safe" variant may not be available
on all platform. however we bundle our own `queue.h` so this should not
be an issue.

Bug: https://github.com/OpenRC/openrc/issues/589

### DIFF
--- a/src/openrc/rc.c
+++ b/src/openrc/rc.c
@@ -351,16 +351,31 @@ static char *get_krunlevel(void)
 static void
 add_pid(pid_t pid)
 {
+	sigset_t sset, old;
 	RC_PID *p = xmalloc(sizeof(*p));
 	p->pid = pid;
+
+	/* this list will be accessed inside the SIGCHLD signal handler.
+	 * so we need to ensure that the SIGCHLD handler doesn't get invoked
+	 * while the list is at an inconsistent state.
+	 */
+	sigemptyset(&sset);
+	sigaddset(&sset, SIGCHLD);
+	sigprocmask(SIG_SETMASK, &sset, &old);
 	LIST_INSERT_HEAD(&service_pids, p, entries);
+	sigprocmask(SIG_SETMASK, &old, NULL);
 }
 
 static void
 remove_pid(pid_t pid, bool inside_signal)
 {
+	sigset_t sset, old;
 	RC_PID *p, *tmp;
 
+	/* same rationale for blocking SIGCHLD as add_pid() */
+	sigemptyset(&sset);
+	sigaddset(&sset, SIGCHLD);
+	sigprocmask(SIG_SETMASK, &sset, &old);
 	LIST_FOREACH(p, &service_pids, entries) {
 		if (p->pid == pid) {
 			LIST_REMOVE(p, entries);
@@ -375,6 +390,7 @@ remove_pid(pid_t pid, bool inside_signal)
 			free(p);
 		}
 	}
+	sigprocmask(SIG_SETMASK, &old, NULL);
 }
 
 static void
--- a/src/openrc/rc.c
+++ b/src/openrc/rc.c
@@ -87,6 +87,7 @@ static RC_HOOK hook_out;
 struct termios *termios_orig = NULL;
 
 RC_PIDLIST service_pids;
+RC_PIDLIST free_these_pids;
 
 static void
 clean_failed(void)
@@ -140,6 +141,12 @@ cleanup(void)
 	}
 
 	while (p1) {
+		p2 = LIST_NEXT(p1, entries);
+		free(p1);
+		p1 = p2;
+	}
+
+	for (p1 = LIST_FIRST(&free_these_pids); p1; /* no-op */) {
 		p2 = LIST_NEXT(p1, entries);
 		free(p1);
 		p1 = p2;
@@ -350,16 +357,24 @@ add_pid(pid_t pid)
 }
 
 static void
-remove_pid(pid_t pid)
+remove_pid(pid_t pid, bool inside_signal)
 {
-	RC_PID *p;
+	RC_PID *p, *tmp;
 
-	LIST_FOREACH(p, &service_pids, entries)
-	    if (p->pid == pid) {
-		    LIST_REMOVE(p, entries);
-		    free(p);
-		    return;
-	    }
+	LIST_FOREACH(p, &service_pids, entries) {
+		if (p->pid == pid) {
+			LIST_REMOVE(p, entries);
+			LIST_INSERT_HEAD(&free_these_pids, p, entries);
+			break;
+		}
+	}
+	/* only call free if we're not inside a signal handler */
+	if (!inside_signal) {
+		LIST_FOREACH_SAFE(p, &free_these_pids, entries, tmp) {
+			LIST_REMOVE(p, entries);
+			free(p);
+		}
+	}
 }
 
 static void
@@ -397,7 +412,7 @@ handle_signal(int sig)
 
 		/* Remove that pid from our list */
 		if (pid > 0)
-			remove_pid(pid);
+			remove_pid(pid, true);
 		break;
 
 	case SIGWINCH:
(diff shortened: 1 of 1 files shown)


src/openrc/rc.c
```diff
@@ -349,22 +356,44 @@ static char *get_krunlevel(void)
...
+	sigset_t sm, oldsm;
+
 	RC_PID *p = xmalloc(sizeof(*p));
 	p->pid = pid;
+
+	/* this list might be accessed inside the signal handler. so block
+	 * singnals to ensure it doesn't get into an inconsistent state.
+	 */
+	sigfillset(&sm);
+	sigprocmask(SIG_SETMASK, &sm, &oldsm);
```
N-R-K, line 368: Since the list is only accessed inside the `SIGCHLD` handler, I think blocking just that might be okay instead of blocking everything.

src/openrc/rc.c
```diff
@@ -349,22 +356,44 @@ static char *get_krunlevel(void)
 static void
 add_pid(pid_t pid)
 {
+	sigset_t sm, oldsm;
+
 	RC_PID *p = xmalloc(sizeof(*p));
 	p->pid = pid;
+
+	/* this list might be accessed inside the signal handler. so block
+	 * singnals to ensure it doesn't get into an inconsistent state.
```
martinetd, line 365: (typo) signals

src/openrc/rc.c
martinetd: I think this is short enough that keeping code simple is better, but no preference here.

src/openrc/rc.c
```diff
@@ -349,22 +356,44 @@ static char *get_krunlevel(void)
...
-		    return;
-	    }
+	sigset_t sm, oldsm;
+	RC_PID *p, *tmp;
+
+	sigfillset(&sm);
+	LIST_FOREACH(p, &service_pids, entries) {
+		if (p->pid == pid) {
+			/* same rationale for blocking signal as add_pid() */
+			sigprocmask(SIG_SETMASK, &sm, &oldsm);
```
martinetd, line 383: LIST_FOREACH itself isn't safe if a link is removed halfway through, this needs to be blocked through the whole loop

martinetd (commented):
I'll let it run this week with this patch (adjusting the blocking in remove_pid) -- usually reproduce the hang within a day or two so should be confident enough if it runs through for a week.

I think we can do better cf. discussion in the issue, but this looks great as a first step -- thanks a lot!

src/openrc/rc.c
N-R-K: Hmm, good catch. I'll update it on the next iteration.

src/openrc/rc.c
```diff
@@ -150,6 +151,12 @@ cleanup(void)
 		p1 = p2;
 	}
 
+	for (p1 = LIST_FIRST(&free_these_pids); p1; /* no-op */) {
```
N-R-K, line 154: Any reason for not using `LIST_FOREACH_SAFE` ? Since we bundle our own `queue.h` there should be no portability issue, right?

N-R-K: Pushed in ae12fab, can revert it if there's any objection to it.

src/openrc/rc.c
```diff
@@ -139,10 +139,13 @@ cleanup(void)
 		rc_logger_close();
 	}
 
-	while (p1) {
-		p2 = LIST_NEXT(p1, entries);
-		free(p1);
-		p1 = p2;
+	LIST_FOREACH_SAFE(p, &service_pids, entries, tmp) {
+		LIST_REMOVE(p, entries);
```
martinetd, line 143: Are signals blocked by the time we get here? (sorry not able to check quickly right now)
This is a good simplification however, list_foreach_safe is perfect for this.

src/openrc/rc.c
```diff
@@ -344,22 +347,46 @@ static char *get_krunlevel(void)
...
+			LIST_REMOVE(p, entries);
+			LIST_INSERT_HEAD(&free_these_pids, p, entries);
+			break;
+		}
+	}
+	sigprocmask(SIG_SETMASK, &old, NULL);
+	/* only call free if we're not inside a signal handler */
+	if (!inside_signal) {
+		LIST_FOREACH_SAFE(p, &free_these_pids, entries, tmp) {
+			LIST_REMOVE(p, entries);
```
martinetd, line 386: fwiw strictly speaking that list remove ouside of signal is still racy with the list_insert_head inside signal above... (This is much less likely than tripping on another malloc though so I probably won't trigger any issue on this)

While using the "safe" iterator ism't bad per se I think it gives a false impression of security when there isn't any; it feels simpler to just keep the mask a bit longer.
the safe iterators only protect from our own removes, not even concurrent removes (the 'tmp' link could be removed while we hold it, or in the middle of an add to that link), so it's really not that safe at all.

N-R-K: > not even concurrent removes

I don't believe there's any concurrent removes going on in the `free_these_pids` list. Unless you were just saying that in general.

> fwiw strictly speaking that list remove ouside of signal is still racy with the list_insert_head inside signal above...

Hmmm, *this* on the other hand is something I can see happening. I guess we'll just have to block `SIGCHLD` for the entire duration of this function.

martinetd: yes, in general, just rambling _safe isn't a good name for this despite being used everywhere.

Anyway, this looks good to me with signal blocked through; I'll update my running test to this exact patch. Thanks!

martinetd (approved):
tests been running for a full 3 days without problem, I think we can safely say this fixes the hang in #589 ! There's probably more cleanup for later, but I'd be of the opinion of merging this first and continuing in a later PR -- let's remove the RFC/draft for this?

Thanks again!

thesamesam:
@vapier could you take a look? I don't feel comfortable enough yet to review this myself

vapier:
the first two commits seem straightforward enough to merge.  the last one is a bit more complicated, and i don't have access currently to my main source machine to review code direcly.

N-R-K:
From the linux manpage:

	LIST_FOREACH() doesn't allow var to be removed or freed within the
	loop, as it would interfere with the traversal.  LIST_FOREACH_SAFE(),
	which is present on the BSDs but is not present in glibc, fixes this
	limitation  by allowing  var  to  safely  be removed from the list and
	freed from within the loop without interfering with the traversal.

It basically gets rid of us having to manage the tmp variable (`p2` in the current master branch) ourselves and does that for us. And since we bundle a `queue.h` portability shouldn't be an issue (I think).

vapier:
if that was a response to me, i'm counting in terms of git commit history from oldest-to-newset.  [rc: use LIST_FOREACH_SAFE in cleanup()](https://github.com/OpenRC/openrc/pull/594/commits/0f267e07783b85471bd6f6a5ce550a989fe77f8c) looks fine.  [rc: avoid calling free inside SIGCHLD handler](https://github.com/OpenRC/openrc/pull/594/commits/9c186711a41c9f1a17434e15a2c12f34a5561f73) needs more thought/review.

N-R-K:
> if that was a response to me, i'm counting in terms of git commit history from oldest-to-newset.

Ah, I was looking at it newest to oldest instead.

N-R-K:
ping!

martinetd:
Thank you!

Issue #344, opened 2019-12-27
bash-completion: rc-service auto completion fails when grep is aliased


Spyros-Ladon:
Problem:
* `rc-service b<Tab>` fails to list services starting with 'b'.

Steps to reproduce:
* `alias grep='grep --color=always --line-number'`
* `rc-service b<Tab>`

N-R-K:
Should be fixed by #597 

Also when looking at this I noticed this:

https://github.com/OpenRC/openrc/blob/ae5e38dce5bd7b30adb3131e7e76fab34a6f5c73/bash-completion/rc-service#L95-L101

I don't believe we should assume `extglob` to be unset - I think it's possible that some users have it set. Should we guard this under a condition so that if the user had `extglob` set it doesn't get unset by us? What do you think @williamh.

PR #597, opened 2023-02-01
bash-completions/rc-service: don't use grep

N-R-K:
Closes: https://github.com/OpenRC/openrc/issues/344

### DIFF
--- a/bash-completion/rc-service
+++ b/bash-completion/rc-service
@@ -85,9 +85,9 @@ _rc_service()
 	else
 		# no option was typed
 		if [[ ${COMP_CWORD} -eq 1 ]]; then			  # if first word typed
-			words="$(rc-service --list | grep ^${cur})"		  # complete for init scripts
-			COMPREPLY=($(for i in ${words} ; do \
-			[[ ${i} == ${cur}* ]] && echo ${i} ; \
+			# complete for init scripts
+			COMPREPLY=($(for i in $(rc-service --list) ; do \
+				[[ ${i} == "${cur}"* ]] && echo ${i} ; \
 			done))
 			return 0
 		elif [[ ${COMP_CWORD} -eq 2 ]] && [[ ${prev} != -* ]]; then # if second word typed and we didn't type in a function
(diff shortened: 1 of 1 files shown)


vapier:
we use `grep` in 2 places, and it looks like we don't need it at all.
```shell
	words="$(rc-service --list | grep ^${cur})"		  # complete for init scripts
	COMPREPLY=($(for i in ${words} ; do \
	[[ ${i} == ${cur}* ]] && echo ${i} ; \
	done))
```

grep is used to filter the output of `rc-service --list`, but then we also manually filter with bash.  it's actually kind of worse in that we first treat the user input as a regex, and then we treat it as a bash glob.  it should be a string in both cases.

```diff
-	words="$(rc-service --list | grep ^${cur})"		  # complete for init scripts
-	COMPREPLY=($(for i in ${words} ; do \
+	COMPREPLY=($(for i in $(rc-service --list) ; do \
-	[[ ${i} == ${cur}* ]] && echo ${i} ; \
+	[[ ${i} == "${cur}"* ]] && echo "${i}" ; \
	done))
```

then we don't need to muck with aliases at all.   i'm not saying we want to respect them, it's just easier to not mess with traps in the first place.

N-R-K:
That seems reasonable. Did some light testing with that patch and it seems to work OK.

bash-completion/rc-service
```diff
@@ -105,8 +105,7 @@ _rc_service()
 	fi
 	if [[ ${action} == '--exists' ]] || [[ ${action} == '-e' ]] || \
 		[[ ${action} == '--resolve' ]]  || [[ ${action} == '-r' ]]; then
-		words="$(rc-service --list | grep ^${cur})"
-		COMPREPLY=($(for i in ${words} ; do \
+		COMPREPLY=($(for i in $(rc-service --list) ; do \
 			[[ ${i} == ${cur}* ]] && echo ${i} ; \
```
vapier, line 109: `"${cur}"` needs quoting here too

vapier:
thx bud

PR #599, opened 2023-02-02
openrc-run: fix rc_parallel race in svc_exec

martinetd:
svc_exec waits until SIGCHLD comes in to close its input, but in rc_parallel case the SIGCHLD might be unrelated.

Checking the proper pid is found in signal handler and only signaling signal_pipe the status code directly avoids this problem.

This problem only appears when stdout is a tty, e.g. I could reproduce both on start and stop with `while rc-service s1 start && rc-service s2 stop; do date; done`, but not with `while ... done < /dev/null | cat` (s[1-3] from #598)

It'd  just fail with something like this:
```
s3                       |s3 stopped
s1                       |s1 stopped
s2                       | * ERROR: s2 failed to stop
```
reproducing under strace shows s2 tried to write to stdout (from its stop() function) and failed with EIO, because openrc-run already moved on an close its end of the tty. That doesn't happen when no tty because stdout is just preserved from the caller's stdout and doesn't go bad.

In practice, boot log is not a tty so this is not a critical problem, just something that happens interactively from time to time.

### DIFF
--- a/src/openrc-run/openrc-run.c
+++ b/src/openrc-run/openrc-run.c
@@ -107,7 +107,8 @@ static RC_STRINGLIST *deptypes_mwua;	/* need+want+use+after deps for stopping */
 static void
 handle_signal(int sig)
 {
-	int serrno = errno;
+	int serrno = errno, status;
+	pid_t pid;
 	const char *signame = NULL;
 	struct winsize ws;
 
@@ -117,12 +118,13 @@ handle_signal(int sig)
 		break;
 
 	case SIGCHLD:
-		if (signal_pipe[1] > -1) {
-			if (write(signal_pipe[1], &sig, sizeof(sig)) == -1)
-				eerror("%s: send: %s",
-				    service, strerror(errno));
-		} else
-			rc_waitpid(-1);
+		while ((pid = waitpid(-1, &status, WNOHANG)) > 0) {
+			if (signal_pipe[1] > -1 && pid == service_pid) {
+				if (write(signal_pipe[1], &status, sizeof(status)) == -1)
+					eerror("%s: send: %s",
+					    service, strerror(errno));
+			}
+		}
 		break;
 
 	case SIGWINCH:
@@ -438,6 +440,7 @@ svc_exec(const char *arg1, const char *arg2)
 			if (errno != EINTR) {
 				eerror("%s: poll: %s",
 				    service, strerror(errno));
+				ret = -1;
 				break;
 			}
 		}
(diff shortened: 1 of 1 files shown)


src/openrc-run/openrc-run.c
```diff
@@ -448,9 +451,18 @@ svc_exec(const char *arg1, const char *arg2)
 				write_prefix(buffer, bytes, &prefixed);
 			}
 
-			/* Only SIGCHLD signals come down this pipe */
-			if (fd[0].revents & (POLLIN | POLLHUP))
+			/* Only SIGCHLD signals come down this pipe,
+			 * but with rc_parallel is might be another process */
+			if (fd[0].revents & (POLLIN | POLLHUP)) {
+				if (read(signal_pipe[0], &ret, sizeof(ret)) != sizeof(ret))
```
N-R-K, line 457: Does it make sense to do this `read` in case of `POLLHUP` ?

src/openrc-run/openrc-run.c
```diff
@@ -448,9 +451,18 @@ svc_exec(const char *arg1, const char *arg2)
 				write_prefix(buffer, bytes, &prefixed);
 			}
 
-			/* Only SIGCHLD signals come down this pipe */
-			if (fd[0].revents & (POLLIN | POLLHUP))
+			/* Only SIGCHLD signals come down this pipe,
+			 * but with rc_parallel is might be another process */
```
N-R-K, line 455: I think this comment should be added above because it's misleading in here - since with the current patch - if it's another process then nothing will be written on the pipe.

src/openrc-run/openrc-run.c
```diff
@@ -448,9 +451,18 @@ svc_exec(const char *arg1, const char *arg2)
...
+			if (fd[0].revents & (POLLIN | POLLHUP)) {
+				if (read(signal_pipe[0], &ret, sizeof(ret)) != sizeof(ret))
+					eerrorx("%s: receive exit code: %s",
+						service, strerror(errno));
+				ret = WEXITSTATUS(ret);
+				if (ret != 0 && errno == ECHILD)
+					/* killall5 -9 could cause this */
+					ret = 0;
```
N-R-K, line 463: I think it makes sense to pull up the `service_pid = 0;` in here as well - though as far as I see it shouldn't make any difference (excluding any signal shenanigans).

N-R-K (changes requested):

src/openrc-run/openrc-run.c
martinetd: not really, but the code above with master_tty did the same (without even checking the return value...) so I just kept it simple because it also shouldn't hurt: read will just return 0 and not block in this case, so we'll print an error and bail -- I think that's as good as we can do.
I guess the error message won't make sense then, but it really shouldn't happen: HUP should only happen if `signal_pipe[1]` is closed and it's only closed by this thread after the loop ends... So I won't bother unless someone thinks it'll be useful.

martinetd: fixed comment and error message, should be possible to make the difference if we somehow get an impossible HUP without IN now (read 0 -> short read message)

N-R-K (approved):

PR #603, opened 2023-02-19
openrc-init: Make openrc-init respect -q (quiet)

### LINKED ISSUE #487: "is starting" message does not honor --quiet
I have an issue with https://github.com/OpenRC/openrc/blob/6e214b261604c4ab1ffc244272443a587bb59927/src/rc/openrc-init.c#L264

I prefer a silent boot, but the only thing I can do is pipe into /dev/null to suppress this message.

Thanx.

Stoneblackdog:
Fixed #487. Althought I did change how you define default runlevels (you put it after -r), so if that's something which you didn't want, then sorry.

### DIFF
--- a/src/openrc-init/openrc-init.c
+++ b/src/openrc-init/openrc-init.c
@@ -270,7 +270,7 @@ int main(int argc, char **argv)
 			case 'r':
 				default_runlevel = optarg;
 			case 'q':
-				isquiet = true;  
+				isquiet = true;
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


thesamesam:
It's okay to force push if you hit a problem and/or mark the PR as a draft, no need to make a new one.

Stoneblackdog:
Sorry then.

thesamesam:
No need to apologise, just letting you know. Don't be nervous :)

PR #604, opened 2023-02-19
openrc-init: Make openrc-init respect -q (quiet)

### LINKED ISSUE #487: "is starting" message does not honor --quiet

Stoneblackdog:
Fixed #487. ~~Although I did change how you specify default runlevels (you have to put -r before it), so if you didn't want that, then sorry.~~

### DIFF
--- a/src/openrc-init/openrc-init.c
+++ b/src/openrc-init/openrc-init.c
@@ -272,8 +272,10 @@ int main(int argc, char **argv)
 
 	if (argc > 1)
 		default_runlevel = argv[optind];
+	else
+		default_runlevel = NULL;
 
-    if (default_runlevel && strcmp(default_runlevel, "reexec") == 0)
+	if (default_runlevel && strcmp(default_runlevel, "reexec") == 0)
 		reexec = true;
 
 	/* block all signals we do not handle */
--- a/src/openrc-init/openrc-init.c
+++ b/src/openrc-init/openrc-init.c
@@ -262,7 +262,7 @@ int main(int argc, char **argv)
 		}
 	}
 #endif
-	
+
 	default_runlevel = NULL;
 
 	while ((opt = getopt(argc, argv, "r:q")) != -1) {
@@ -270,7 +270,7 @@ int main(int argc, char **argv)
 			case 'r':
 				default_runlevel = optarg;
 			case 'q':
-				isquiet = true;  
+				isquiet = true;
 		}
 	}
 
--- a/src/openrc-init/openrc-init.c
+++ b/src/openrc-init/openrc-init.c
@@ -231,7 +231,7 @@ int main(int argc, char **argv)
 	bool reexec = false;
 	sigset_t signals;
 	int opt;
-    bool isquiet;
+	bool isquiet;
 	struct sigaction sa;
 #ifdef HAVE_SELINUX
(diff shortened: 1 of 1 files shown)


Stoneblackdog:
Also sorry for opening 3 pull requests.

N-R-K:
> Although I did change how you specify default runlevels

Don't think this change was needed. Look into `optind` on how to extract arguments when flags end.

- - -

Also, it's typically best to create PRs from a separate branch instead of `master`. But it's not a big deal, so don't worry about it for now :)

src/openrc-init/openrc-init.c
```diff
@@ -261,14 +262,18 @@ int main(int argc, char **argv)
...
+				isquiet = true;
+
+	if (!isquiet)
+		printf("OpenRC init version %s starting\n", VERSION);
 
 	if (argc > 1)
-		default_runlevel = argv[1];
-	else
-		default_runlevel = NULL;
+		default_runlevel = argv[optind];
```
N-R-K, line 274: Keep the `else` part and remove the earlier `default_runlevel = NULL;`. The initialization should happen in one place instead of being split up into two distant places.

N-R-K (changes requested):
The indentation looks very off. Ensure your editor is configured properly to use TABs and not spaces for indent. See openrc's [style guide](https://github.com/OpenRC/openrc/blob/master/STYLE-GUIDE.md).

Stoneblackdog:
I have seen the style guide. I don't know how it got there.

src/openrc-init/openrc-init.c
```diff
@@ -261,10 +262,16 @@ int main(int argc, char **argv)
 	}
 #endif
 
-	printf("OpenRC init version %s starting\n", VERSION);
+	default_runlevel = NULL;
```
N-R-K, line 265: This isn't really needed. I'd remove this line.

src/openrc-init/openrc-init.c
```diff
@@ -261,10 +262,16 @@ int main(int argc, char **argv)
...
+
+	if (getopt(argc, argv, "q") == 'q')
+				isquiet = true;
```
N-R-K, line 268: A bit too many TABs over here :)

src/openrc-init/openrc-init.c
```diff
@@ -230,6 +230,7 @@ int main(int argc, char **argv)
 	FILE *fifo;
 	bool reexec = false;
 	sigset_t signals;
+	bool isquiet;
```
N-R-K, line 233: Initialize it to `false` - otherwise it'll get used uninitialized later below.

N-R-K (changes requested):

Stoneblackdog:
Thank you for dealing with my _weird_ code. :)

N-R-K (approved):
Looks OK to me. But I'm not a maintainer here, so you'll need to wait until a maintainer can take a look.

(I suspect the manpage might need updating, though.)

PR #608, opened 2023-03-06
rc.conf: default to unicode console

### LINKED ISSUE #607: question regarding defaulting to unicode=YES
i recently noticed that when setting a default LANG that ends with `UTF-8`, things in the console that output special characters (pipes, bars, ..) appear broken if the console is not also set to utf-8. the console setting for this is set via the default-shipped termencoding service, reading the UNICODE value.

the default for it is not-unicode, from the default rc.conf https://github.com/OpenRC/openrc/blob/95b2948e6e153271249b09b9b1505c527c54e0cd/etc/rc.conf#L92  

is there:

- potential downsides to defaulting to 'YES' distro-wide
- a reason it's not the default already?

nekopsykose:
it's the high year of 2023, after all.

closes #607

### DIFF
--- a/etc/rc.conf
+++ b/etc/rc.conf
@@ -88,8 +88,8 @@
 # MISC CONFIGURATION VARIABLES
 # There variables are shared between many init scripts
 
-# Set unicode to YES to turn on unicode support for keyboards and screens.
-#unicode="NO"
+# Set unicode to NO to turn off unicode support for keyboards and screens.
+unicode="YES"
 
 # This is how long fuser should wait for a remote server to respond. The
 # default is 60 seconds, but  it can be adjusted here.


nekopsykose:
perhaps it should be changed elsewhere instead of here, given that before it was just a commented-default

vapier:
the default is "no" by virtue of being empty, and the various init scripts treating "" as "no".  changing this config is the least disruptive for people upgrading.  we can prob give it a year/few releases before commenting it out and changing the init scripts to default the value to "yes" manually.

nekopsykose:
that makes sense. thanks for elaborating vapier :)

williamh:
The convention should be to comment out the default in rc.conf and uncomment it only when it is being changed.

vapier:
as a generalism, that is absolutely true, but i as i mentioned above, i think changing defaults requires a bit more consideration.  an /etc config change, e.g. in Gentoo, encourages/forces people to review & accept the new default.  after we go through that transition period, we can rewrite the code internally to consider unset `$unicode` to be `yes` by default, and update the config file to be commented out again (with it being documented as `yes`).

williamh:
@vapier  In gentoo, we already use sed to force the unicode default to be yes, so we deviate downstream from what the OpenRC releases have as default (see the last openrc release ebuild). If we make the internal default yes, we are just moving Gentoo's default upstream.
People will see the changes in /etc/rc.conf via etc-update, so if they want to turn off unicode they can.

vapier:
i haven't looked at the ebuild to see if it needs changing. this isn't about Gentoo defaults though, this is simply doing the right thing out of the box: Unicode/UTF8 enabled should be the default, and anyone wanting to live in a 7bit/non-UTF8 world has to change their config.

if you think simply changing the commented out default is sufficient notification (and something in the release notes), I'm fine rolling that now.

nekopsykose:
i'm also fine with this being reverted in itself and being replaced by your preference for the same changes. how it's done doesn't matter to me :)  

>  In gentoo, we already use sed to force the unicode default to be yes,  

if you had done it here ages ago instead, everyone could've had those benefits, and i wouldn't have been confused for a really long time why my console was broken by default until i figured out what part of the stack even controlled this :p

of course, no real harm done. thanks for the hard work :)

Issue #622, opened 2023-04-18
Usage of _unused macro conflicts with linux-headers on PowerPC


matoro:
I am unsure what specific change (or kernel headers change) caused this, but I am unable to compile the new release on PowerPC (both BE/LE tested 64-bit) due to the following error:

```
FAILED: src/librc/librc.so.1.p/librc.c.o 
powerpc64le-unknown-linux-gnu-gcc -Isrc/librc/librc.so.1.p -Isrc/librc -I../openrc-0.47/src/librc -Isrc/shared -I../openrc-0.47/src/shared -Isrc/libeinfo -I../openrc-0.47/src/libeinfo -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Wpedantic -std=c99 -D_DEFAULT_SOURCE -Wcast-align -Wcast-qual -Wdeclaration-after-statement -Wformat=2 -Winline -Wmissing-declarations -Wmissing-format-attribute -Wmissing-noreturn -Wmissing-prototypes -Wnested-externs -Wpointer-arith -Wsequence-point -Wshadow -Wwrite-strings -Werror=implicit-function-declaration -DHAVE_MALLOC_EXTENDED_ATTRIBUTE -DHAVE_CLOSEFROM -DHAVE_CLOSE_RANGE_CLOEXEC -O3 -mcpu=native -mtune=native -O2 -pipe -fPIC -MD -MQ src/librc/librc.so.1.p/librc.c.o -MF src/librc/librc.so.1.p/librc.c.o.d -o src/librc/librc.so.1.p/librc.c.o -c ../openrc-0.47/src/librc/librc.c
In file included from /usr/include/bits/sigcontext.h:30,
                 from /usr/include/signal.h:301,
                 from /usr/include/sys/param.h:28,
                 from ../openrc-0.47/src/librc/librc.c:30:
/usr/include/asm/sigcontext.h:18:32: error: expected identifier or ‘(’ before ‘[’ token
   18 |         unsigned long   _unused[4];
      |                                ^
ninja: build stopped: subcommand failed.
```

vapier:
both projects are wrong.  we can fix ours at least.  go with `attr_xxx` instead of `_xxx` or something.  and drop support for <gcc-3 in the process.

PR #623, opened 2023-04-19
Drop use of custom _unused

thesamesam:
Rename attribute macros to namespaced RC_*

This conflicts with linux-headers which uses __unused for some padding members
on ppc64le at least.

Closes: https://github.com/OpenRC/openrc/issues/622

### DIFF
--- a/src/shared/helpers.h
+++ b/src/shared/helpers.h
@@ -30,20 +30,9 @@
 
 #define UNCONST(a)		((void *)(unsigned long)(const void *)(a))
 
-#ifdef lint
-# define _unused
-#endif
-#if __GNUC__ > 2 || defined(__INTEL_COMPILER)
-# define _dead __attribute__((__noreturn__))
-# define _noreturn __attribute__ ((__noreturn__))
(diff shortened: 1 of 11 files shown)


vapier:
having to inline the full attribute macro is unwieldy which is why having local defines is preferable

maybe someday when we require C23+ we can inline things like `[[unused]]`

dwfreed (approved):

src/shared/helpers.h
```diff
@@ -30,20 +30,9 @@
 
 #define UNCONST(a)		((void *)(unsigned long)(const void *)(a))
```
vapier, line 31: oh wow i only noticed this now.  this is horribly broken.  my desktop is broken atm so i can't fix.  this needs to be `uintptr_t`, not `unsigned long`.

vapier (approved):

src/shared/helpers.h
thesamesam: good spot, done in https://github.com/OpenRC/openrc/pull/624

PR #625, opened 2023-04-20
cgroups: set (irrelenvant) device string to cgroup2

pabloyoyoista:
Just a cosmetic change with no real changes, more than the visual impact when running mount


martinetd (approved):
No reason this'd fail but tested on my 5.10-based board; mount output went from `none on /sys/fs/cgroup ...` to `cgroup2 on /sys/fs/cgroup ...`  as expected.

Issue #626, opened 2023-04-23
Hurd FTBFS


markhindley:
Hi,

I have been investigating 2 recent Debian FTBFS on Hurd:-

> make[1]: Entering directory '/<<PKGBUILDDIR>>'
> dh_auto_configure -- --libexecdir=lib
> 	cd obj-i686-gnu && LC_ALL=C.UTF-8 meson .. --wrap-mode=nodownload --buildtype=plain --prefix=/usr --sysconfdir=/etc --localstatedir=/var --libdir=lib/i386-gnu --libexecdir=lib
> The Meson build system
> Version: 0.63.0
> Source dir: /<<PKGBUILDDIR>>
> Build dir: /<<PKGBUILDDIR>>/obj-i686-gnu
> Build type: native build
> Project name: OpenRC
> Project version: 0.45.2
> C compiler for the host machine: cc (gcc 11.2.0 "cc (Debian 11.2.0-16) 11.2.0")
> C linker for the host machine: cc ld.bfd 2.38.90.20220713
> Host machine cpu family: x86
> Host machine cpu: i686-at386
> Found pkg-config: /usr/bin/pkg-config (0.29.2)
> Did not find CMake 'cmake'
> Found CMake: NO
> Run-time dependency audit found: NO (tried pkgconfig and cmake)
> Program uname found: YES (/bin/uname)
> WARNING: You should add the boolean check kwarg to the run_command call.
>          It currently defaults to false,
>          but it will default to true in future releases of meson.
>          See also: https://github.com/mesonbuild/meson/issues/9300
> 
> ../meson.build:47:2: ERROR: C shared or static library 'kvm' not found
> 

And 

> FAILED: src/start-stop-daemon/start-stop-daemon.p/start-stop-daemon.c.o 
[...]

PR #627, opened 2023-04-23
Hurd ftbfs fixes from Debian

markhindley:
PR as requested in #626 

Mark

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -43,7 +43,7 @@ else
   os = option_os
 endif
 
-if os != 'Linux'
+if os != 'Linux' and os != 'GNU'
   kvm_dep = cc.find_library('kvm', required: true)
 else
   kvm_dep = []
--- a/src/start-stop-daemon/start-stop-daemon.c
+++ b/src/start-stop-daemon/start-stop-daemon.c
@@ -21,10 +21,8 @@
 
 #define ONE_MS           1000000
 
-#ifdef __linux__
-/* For extra SCHED_* defines. */
+/* For extra SCHED_* defines and pipe2. */
 # define _GNU_SOURCE
-#endif
 
 #include <errno.h>
 #include <fcntl.h>


meson.build
```diff
@@ -43,7 +43,7 @@ else
   os = option_os
 endif
 
-if os != 'Linux'
+if os != 'Linux' and os != 'GNU'
```
vapier, line 46: not a new issue, but this check seems to assume that we always want kvm on non-Linux systems.  this should really be changed to only require kvm on BSD systems

src/start-stop-daemon/start-stop-daemon.c
```diff
@@ -21,10 +21,8 @@
 
 #define ONE_MS           1000000
 
-#ifdef __linux__
-/* For extra SCHED_* defines. */
+/* For extra SCHED_* defines and pipe2. */
 # define _GNU_SOURCE
```
vapier, line 25: don't leave it indented

src/start-stop-daemon/start-stop-daemon.c
```diff
@@ -21,10 +21,8 @@
 
 #define ONE_MS           1000000
 
-#ifdef __linux__
-/* For extra SCHED_* defines. */
+/* For extra SCHED_* defines and pipe2. */
```
vapier, line 24: the pipe2 usage implies a bigger problem here i think.  we're using it unprotected.  we really need a test in meson to see if the system supports pipe2 and switch between it & standard pipe.

vapier (changes requested):
"ftbfs" is a Debian thing and isn't common outside that space.  just write out the term or appropriate alternative.

vapier:
these commits are better and i don't think make the status quo worse, but i still think we should fix the regression around pipe2 rather than make it a hard requirement

meson.build
```diff
@@ -43,7 +43,7 @@ else
   os = option_os
 endif
 
-if os != 'Linux'
+if os == 'Dragonfly' or os == 'FreeBSD' or os == 'NetBSD' or os == 'OpenBSD'
```
pinotree, line 46: instead of hardcoding almost all the BSD OSes, this can be simplified as:
```suggestion
if os.contains('BSD') or os == 'Dragonfly'
```

src/start-stop-daemon/start-stop-daemon.c
pinotree: while this is correct, OTOH it seems that `pipe2()` is available in all the OSes for which there is some kind of support in OpenRC:
- Linux & Hurd: provided by glibc
- FreeBSD: https://man.freebsd.org/cgi/man.cgi?query=pipe2&sektion=2&n=1
- OpenBSD: https://man.openbsd.org/pipe.2
- NetBSD: https://man.netbsd.org/pipe2.2
- DragonFlyBSD: https://man.bsd.lv/DragonFly-5.6.1/man2/pipe2.2

so adding a meson test for it with fallback to `pipe()` would effectively create dead code, because it would not be used on any of the supported OSes; considering that currently `pipe2()` is used unconditionally without issues, IMHO it would be acceptable to keep using it like that, and wait until anyone tries to port OpenRC to another OS

N-R-K: > see if the system supports pipe2 and switch between it & standard pipe.

I haven't taken a look into the usage yet, so this might not be a problem, but `pipe2` and `pipe + fcntl` aren't functionally equivalent.

With `pipe + fcntl` there exists a window between these two calls where the `fd` might not have proper mode set. This can become an issue in multi-threaded or signal handler contexts.

N-R-K:
> regression around pipe2 

Is it a regression though since it's supported by all OSes we care about? It doesn't seem worthwhile to me to add a compatibility wrapper which will almost certainly never be used (and thus never tested). It's likely more fruitful to propose to austingroup to standardize pipe2 in posix instead (which I might do this weekend).

IMO, we should merge this. It fixes *actual* build failure and doesn't worsen anything.

vapier:
this PR unconditionally defines `_GNU_SOURCE`.  that is unacceptable.  we support more systems than GNU (like BSDs).

N-R-K:
> this PR unconditionally defines `_GNU_SOURCE`. that is unacceptable. we support more systems than GNU (like BSDs).

As far as I'm aware it just makes certain prototype/defines etc visible on glibc and other libcs which honor that define.

Does it actually have any downside when defined in BSDs? I see many projects unconditionally defining it (and other similar feature test macros) without any issues.

N-R-K:
> this PR unconditionally defines `_GNU_SOURCE`. that is unacceptable. we support more systems than GNU (like BSDs).

There was already one unconditional (and unnecessary) `_GNU_SOURCE` define in `checkpath.c`. I don't quite get what the issue with that is since defining `_GNU_SOURCE` won't hurt on other platform, but in anycase, I've moved the `_GNU_SOURCE` defines into meson now. See https://github.com/OpenRC/openrc/pull/801.

pinotree:
@markhindley considering that PR #801 (thanks @N-R-K!) solved the `pipe2`/etc issues in `start-stop-daemon.c`, the only issue left for building openrc OOTB on Hurd is the change to `meson.build`

Would you please update this PR by rebasing it, dropping all the changes except the ones in `meson.build`, and leave only one commit? Thanks!

PR #628, opened 2023-04-23
einfo.h: fix EINFO_RESTRICT macro usage

xxc3nsoredxx:
Make function declarations use the EINFO_RESTRICT macro instead of __EINFO_RESTRICT which gets treated as the name of the argument.


thesamesam (approved):
Oops.

vapier:
thx mate

PR #629, opened 2023-05-10
librc: allow overriding rc_interactive on kernel command line

thesamesam:
This was originally introduced in 14625346c07a2a66fe77ce578c9423918bec1d97 with an example list (just one for rc_parallel) of options. Let's add in rc_interactive as it's a pretty obvious thing one might want to override.

See https://forums.gentoo.org/viewtopic-p-8694588.html.


vapier:
sure, seems reasonable enough

thesamesam:
cheers

PR #631, opened 2023-05-14
einfo: fix vewend applet args

xxc3nsoredxx:
Fix typo preventing the args for the vewend applet from being handled.


vapier:
nice catch, thx

PR #642, opened 2023-07-28
Fix PATH order at openrc-init

sulincix:


vapier:
looks fine, but would be good to dig in the history a little first to see why it ended up this way

WhyNotHugo:
`path_default` was defined in cfded513cd9b7febe4b7cf39a80411e4303f0655 which reads "This adds the default path setting from sysvinit.".

`_PATH` was defined initially in 0e6eb0f7640a5994755e93ea65e630d6ab917eaa. There's no explanation of the order, it's simply been this way since it was initially defined.

awilfox:
Would it be better to use `getconf PATH` (in shell) and `confstr(_CS_PATH)` (in C)?  That is actually the most efficient way to determine a "default" `PATH` on a POSIX system.

Though, unfortunately, musl's `_CS_PATH` is `/bin:/usr/bin`, but that should probably be fixed upstream…

navi-desu:
> Though, unfortunately, musl's _CS_PATH is /bin:/usr/bin, but that should probably be fixed upstream…

i'm running glibc, both `getconf PATH` and `_CS_PATH` also return `/bin:/usr/bin`

sulincix:
I think PATH variable defined by _CS_PATH. Or defined by /etc/profile or compile parameter.

PR #645, opened 2023-08-29
start-stop-daemon, supervise-daemon: use close_range with musl

ncopa:
Make sure that we use close_range also with musl libc, using syscall directly.

This fixes a severe performance issue when running openrc under docker where the maxfd may be 1G.

### DIFF
--- a/src/shared/misc.c
+++ b/src/shared/misc.c
@@ -15,6 +15,11 @@
  *    except according to the terms contained in the LICENSE file.
  */
 
+#ifdef HAVE_CLOSE_RANGE
+/* For close_range() */
+# define _GNU_SOURCE
+#endif
+
 #include <ctype.h>
 #include <errno.h>
 #include <fcntl.h>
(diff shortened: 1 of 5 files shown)


src/start-stop-daemon/start-stop-daemon.c
```diff
@@ -191,6 +194,22 @@ static inline int ioprio_set(int which RC_UNUSED,
 }
 #endif
 
+#if !defined(SYS_close_range) && defined(__NR_close_range)
```
thesamesam, line 197: I really don't think we should do this. Using `close_range` is fine, but needing hacks for it isn't.

musl should provide a prototype for it?

ncopa:
I refactored it a bit to be less hackish. Also changed start-stop-daemon to use CLOEXEC instead of closing the filedescriptors, similar to what supervise-daemon does.

src/shared/misc.c
```diff
@@ -500,3 +509,29 @@ pid_t get_pid(const char *applet,const char *pidfile)
...
+
+#ifndef HAVE_CLOSE_RANGE
+static inline int close_range(int first RC_UNUSED,
+			      int last RC_UNUSED,
+			      unsigned int flags RC_UNUSED)
+{
+#ifdef SYS_close_range
+	return syscall(SYS_close_range, first, last, flags);
+#else
+	return -1;
```
floppym, line 521: Generally it would be a good idea to set `errno = ENOSYS` before returning.

However, since we don't actually check/log any failures at the call site, I guess it doesn't matter in this specific instance.

ncopa: I agree that `errno = ENOSYS` is a good idea. Let me fix that.

src/start-stop-daemon/start-stop-daemon.c
thesamesam: I think your current approach is alright, even if the situation isn't ideal.

It would be nice if the musl maintainers had replied when you asked about it though :(

Consider my objection withdrawn, thank you for working on this!

ncopa: FWIW, I'm doing a second try of including close_range() in musl: https://www.openwall.com/lists/musl/2023/09/01/1

src/shared/misc.c
ncopa: Turns out that [syscall(2)](
https://man7.org/linux/man-pages/man2/syscall.2.html) sets `errno`, so we don't need to do it our selves.

> The return value is defined by the system call being invoked.  In general, a 0 return value indicates success.  A -1 return value indicates an error, and an error number is stored in errno.

PR #648, opened 2023-08-31
deprecate /lib/rc/tmp completely

timkenhan:
Tested with no regression...


timkenhan:
apparently i misunderstood `RC_TMPDIR` to be `/lib/rc/tmp/` instead of `/run/openrc/`

timkenhan:
apparently i misunderstood `RC_TMPDIR` to be `/lib/rc/tmp/` instead of `/run/openrc/` as it supposed to

PR #649, opened 2023-09-01
move dependency cache to /var/cache/openrc

timkenhan:


ncopa:
What is the motivation behind this?

This means that the dependency cache is unavailable before localmount is started if /var is mounted on separate partition.

timkenhan:
The motivation is to comply better with the FHS guideline. Furthermore,
some systems would have /lib under read-only filesystem (e.g. embedded
systems, immutable systems). At the same time, system with external /var
seems to diminish in number (/var/log & /var/tmp are still popular).

I would also take this opportunity to ask: is the caching even necessary
(or desirable)?

ncopa:
> The motivation is to comply better with the FHS guideline.

FHS assumes that those directories are available. The initsytems job is to make them available, an initsystem will always be special.

> Furthermore, some systems would have /lib under read-only filesystem (e.g. embedded
systems, immutable systems).

If I understand the sources correctly, openrc will try copy the cache from `$RC_LIBEXECDIR/cache`if it exists to `/run/openrc/`, and recreate it if its missing. If your /lib is read-only, you'd only need to disable the `savecache` (which is supposed to run during shutdown).

> At the same time, system with external /var seems to diminish in number (/var/log & /var/tmp are still popular).

What do you base this on?

> I would also take this opportunity to ask: is the caching even necessary (or desirable)?

I'd say yes. You can always disable it by disabling `savecache` during shutdown if you don't want it.

williamh:
We do not attempt to save the cache if /lib or /lib/cache are read only, so the concern about read-only file systems is addressed.
Also, as stated, if we move everything to /var/cache, this would not work if /var is a separate file system from /.

PR #650, opened 2023-09-01
remove /lib/rc/tmp creation from meson script

timkenhan:


ncopa:
Why was this directory created in the first place?

It was introduced with d6622a115

timkenhan:
It is apparently for a piece of code that no longer exist. There don't seem to be any part of the code that refers to this directory...

The reason why I raise this with some urgency is due to Debian maintainers having some issue maintaining the openrc package due to this directory.

thesamesam:
This needs to be explained in the commit message.

ncopa:
Also I can not find anything referring to this directory. 391d12db48754861b5cecac92ee3321597ee02c1 removed `${INSTALL} -d ${DESTDIR}/${LIBEXECDIR}/tmp` when migrating to meson.

c0fd1b49e49cec28c1f5a3a76f9db11c62e550dc added that line when moving it from `${DESTDIR}/${PREFIX}/${RC_LIB}/tmp`

0af7d5bc204cd6b7d03f22aacf4072c5f526c0ee seems to have removed the use of it.

a010d0bb69fbdda086c265d7cab650c485c3e6f3 mentions it again,

seems to been introduced with 5af58b45146ab5253ca964738f4e45287bf963d4. It looks like it used to be a tmpfs mount to store the deptree and similar.

I'd say its save to remove.

thesamesam:
No, we don't do merge commits, and the person submitting a PR should really be justifying their change before it gets merged?

ncopa:
@timkenhan `git commit --amend && git push -f ...`

timkenhan:
@ncopa thanks!

timkenhan:
i've updated the commit message. lemme know if there's anything else i need to do

Issue #619, opened 2023-04-16
Broken with >=ncurses-6.4_p20230408 (OpenRC abuses tparm API)


thesamesam:
OpenRC is broken with >=sys-libs/ncurses-6.4_p20230408, like so:
```
openrc-0.46-build/src/einfo $ ./ewarn hi
 (null)*(null) hi(null)
```

This was originally reported in Gentoo at https://bugs.gentoo.org/904277. The ncurses maintainer, Thomas Dickey, provided some analysis (see the Gentoo bug for full details):
>I see that OpenRC is passing hardcoded strings to tgoto(),
>which do not come from the terminal database (when tgoto()
>returns a null, indicating a failure for that capability),
>and using the output of tgoto() to copy/use other places.
>
>That's a defect in OpenRC.  Perhaps they'll fix it.

OpenRC needs to use `tputs`, `printw`, etc instead of e.g. `printf` for strings obtained from ncurses.

Note that newer ncurses is needed to fix some security vulnerabilities (https://bugs.gentoo.org/904247), so there's some urgency to this.

vapier:
i don't super care one way or the other

deprecated-dot:
> i don't super care one way or the other

That's the spirit.

bfrg:
I'm on Gentoo Linux and because of this "bug" I cannot install newer tmux version, since OpenRC requires the old ncurses version but that conflicts with the latest tmux. This is true for _every_ application that requires a newer ncurses. IMO something as essential as OpenRC should depend on ncurses.

PR #651, opened 2023-09-04
rework libeinfo

xxc3nsoredxx:
Rework how libeinfo does curses stuff. This is not by any means ready to merge as-is. For example, OpenRC core has not been changed where needed (places where `ecolor` is called are the most obvious ones). There also isn't a non-curses fallback yet, but that should be easy to add once everything else is satisfactory. Not to mention squashing/rewording commits.

This is just to get some feedback on what I currently have.

See also: #619


xxc3nsoredxx:
Squashed a few of the commits/cleaned up the messages.

williamh:
We don't need command line equivalents for ewarnx and eerrorx.
In a script, you can use ewarn then exit or do the same with eerror.

williamh:
Please rebase on master.

I also see a couple of changes.

On line 50 of einfo.h:

```
__attribute__((__deprecated__("this shit is gone yo")))
```

What does this apply to? Also, can you fix the message to be something like "this is no longer used"?

Please update if statements that just have one statement in the if or else blocks to not use curly braces around the statement:

```
if (foo) {
    bar;
}
```

should be:

```
if (foo)
    bar;
```


Can you also remove the vim mode lines?

Thanks,

William

xxc3nsoredxx:
I can do the single-line `if` thing, sure. The `__deprecated__` message is currently just a placeholder to alert me to things which use the `ecolor()` function in OpenRC and will be changed before I un-draft the PR. Same with removing the mode line.

src/libeinfo/einfo.h
```diff
@@ -47,6 +47,7 @@ typedef enum
 } ECOLOR;
 
 /*! @brief Returns the ASCII code for the color */
+__attribute__((__deprecated__("this shit is gone yo")))
```
vapier, line 50: sorry, but we can't accept things like this.  ignoring the puerile nature, it's not helpful or clear to anyone who hits/sees this warning.

vapier (changes requested):
there's a lot of commits in this stack.  will prob need to pull out some of the smaller ones independently.

vapier:
> there's a lot of commits in this stack. will prob need to pull out some of the smaller ones independently.

ok, i flipped through the patch series now, and it doesn't make sense to pull things out.  this is a PR showing progress of implementing a new libeinfo library and would make sense squashing all into one.

i'm inclined to go with my half-baked idea that i mentioned in the open issue.  since we already support building openrc w/out termcap (curses) support, we should lean into that: delete support for termcap entirely and just require an ANSI compliant terminal.

src/libeinfo/einfo.h
xxc3nsoredxx: It was merely meant as a message while I was working and not for the final cleaned up/squashed commits.

xxc3nsoredxx:
I can close this PR if you instead want to just drop curses and require an ANSI terminal.

vapier:
you want to send a PR to delete curses support ?

xxc3nsoredxx:
Sure, I think I can do that.

Issue #644, opened 2023-08-25
output_logger and error_logger in supervise-daemon


phirecc:
Is there a technical reason why the `{output,error}_logger` arguments are only available in start-stop-daemon but not in supervise-daemon? I'm not very familiar with the codebase but from my quick reading it looks like it could be implemented in a similar manner?

I would find it useful and some other users in https://github.com/OpenRC/openrc/issues/341 also seem to.

Thanks :)

Higgs1:
I need this also, yes please!

Many need daemons to output to `syslog()` (because syslog is easily redirected to their choice of logger). This can easily be done with s-s-d by setting `{output,error}_logger=logger`. There is no reason that supervise-daemon shouldn't also allow this, and in my mind it's imperative that supervise-daemon has this feature also.

[In this comment](https://github.com/OpenRC/openrc/issues/118#issuecomment-299223810), it was said that "supervise-daemon is intended to be entirely analogous to start-stop-daemon with the exception that it monitors the child process, and restarts it if it exits." Currently this is not the case, as supervisor-daemon does not honour these variables.

Further write-up incoming (probably). Please wait a few minutes...

Thanks

Higgs1:
I'm not confident in my ability to write a full PR, but I still want to investigate and share my findings. Maybe it will help someone who can write c but doesn't have the spoons to do the first few steps (which has been the case for me several times). If not, just delete this.

In [sh/start-stop-daemon.sh#L50-L51](https://github.com/OpenRC/openrc/blob/-/sh/start-stop-daemon.sh#L50-L51) it can be seen that the `{output,error}_logger` variables get passed to `start-stop-daemon` via command line options `--std{out,err}-logger`. [sh/supervise-daemon.sh](https://github.com/OpenRC/openrc/blob/-/sh/supervise-daemon.sh) is the equivalent file for `supervise-daemon`.

[...]

Issue #341, opened 2019-12-06
Which supervisor is recommended?


Jimmy-Z:
I've tried to use s6 but failed, I'll open another issue for this.

supervise-daemon works for me, but I wonder what's the current status? I see no mentioning of "experimental" in supervise-daemon-guide.md but openrc-run man page still says so.

runit is not even mentioned in openrc-run man page.

boardwalk:
I had the same question. So far I've found that:

* start-stop-daemon supports output_logger and error_logger so you can easily gets things into syslog
* supervise-daemon doesn't support that, but does obviously supervise your daemons and give nicer status via rc-status
* s6 (which I have not tried yet) seems to require a different directory/setup (whereas with the two above you can just drop a single script into /etc/init.d), but would support logger processes like start-stop-daemon

There is most likely much more to this. I think ideally, for me, supervise-daemon would gain output_logger and error_logger support.

agelwarg:
@boardwalk I'm running into the same issue -- wishing supervise-daemon supported `output_logger` / `error_logger`.  Have you found out anything further?

boardwalk:
@agelwarg I landed on using supervise-daemon with `output_log` / `error_log` (e.g. plain log files) and setting up logrotate. I didn't *really* need things to go to syslog in the end (I don't need to do off-box log aggregation or anything -- even if I did there are agents that can watch files).

agelwarg:
@boardwalk Makes sense and I’m sorta arriving at the same solution.  Thanks.

Issue #127, opened 2017-04-28
start-stop-daemon: add the ability to pipe stdout and/or stderr to a process rather than a file


williamh:
I have been asked to consider adding options to start-stop-daemon which would allow stdout and/or stderr to be piped to a process rather than just redirected to a file.
The suggested command line options are --stderr-pipe "command" and --stdout-pipe "command".

dwfreed:
They can just use named pipes :man_shrugging:

williamh:
@chutzpah I'm adding you here.

kmoore134:
This is something we'd be interested in testing for FreeBSD/TrueOS.

heistp:
FWIW I'd like to see this. Services written in Go usually expect to write to stdout and rely on the init facility (often systemd) to send this to the syslog. I now have an OpenRC service script for [IRTT](https://github.com/heistp/irtt), but there's no easy way I can get it to log. I could add the ability for IRTT to use Go's log/syslog package, which I'll probably have to do to get this working soon, but having this ability in OpenRC would make things simpler.

williamh:
This is available in OpenRC 0.37. The following commits are related to this feature:

a097933e
e6d01471
d4501a9f
ec8abea4

Issue #118, opened 2017-02-27
--stdout and --stderr to support redirection to syslog


dmolik:
I find myself supporting a lot of new Golang applications that don't have options to send logs to the syslog daemon. It would be great if I could handle that with a oneliner config change in an init script.

Perhaps it could be something like this:

    --stdout syslog:daemon.info \
    --stderr syslog:daemon.err

Where the tag would be populated by either `$SVCNAME` or `--name`.

I'm willing to write a PR for the change. I'm curious if there is community interest, and what the syntax/execution would be.

dwfreed:
The only way to do this is with a separate process, which is attached to the daemon's stdout/stderr with pipes, and then just taking what comes along the pipes and submitting that to syslog.  I think this is beyond the scope of start-stop-daemon.

duanyao:
I'm interested in this feature, too. I'm developing a node.js application and have to send logs to local rsyslogd explicitly via TCP, so it lose some log system independency. If openrc could forward stdout/stderr to syslog, the application could be greatly simplified, and could also run under systemd+journald without modification.

I'm not a linux/unix expert, but using a separate log-forwarding process seems less reliable, isn't it? What if one of the processes connected to the pipes dies unexpectedly? Maybe the log-forwarding process itself need to be a openrc service, or even better, becomes part of openrc.

dwfreed:
systemd already takes what's sent to syslog and puts it in the journal, if the journal is intended to take over for the usual syslog daemon.  POSIX already defines a standard method for getting entries into the syslog, and this is exposed via the syslog() library call (which is part of the specification).  The underlying interface is a Unix socket that lives at /dev/log; as such, it would be impossible to directly connect /dev/log to stdout/stderr, as read() and write() syscalls do not work on sockets, and syslog is expecting messages to be in a specific format anyway (which the library call takes care of).  As openrc has no persistent daemon of its own, a separate process, connected to the daemon to be redirected with pipes, would be needed.  I believe this is beyond the scope of openrc.

duanyao:
@dwfreed Thank for the detailed explanation!

>As openrc has no persistent daemon of its own, a separate process, connected to the daemon to be redirected with pipes, would be needed.

I just tried `supervise-daemon`, isn't it such a persistent daemon that is potentially able to read the application's stdout and sends to syslog?

dwfreed:
[...]

PR #656, opened 2023-09-25
supervise-daemon: implement output_logger and error_logger.

Higgs1:
Implements #644. Allows redirecting process stdout and stderr to another process, just like is already possible with start-stop-daemon. Also addresses #341 and #118.

Almost all of the code is copied directly from s-s-d (#127). I tried to follow your coding style as best I could, I read several previously accepted PRs, and I tried to keep the diff to a minimum.

This works correctly on my machine (Alpine, x86_64) and is currently hosting several of my public services. The stdout and stderr output is successfully redirected to syslog on another machine, which is then emailed to me sometimes.

I didn't copy over the short command arg variants (-3 and -4) because -3 conflicts with supervise-daemon's "--reexec" which doesn't exist in s-s-d. I can edit if you prefer something else. 

As an example, here's a snippet from my /etc/init.d file(s):

```sh
#!/sbin/openrc-run

supervisor=supervise-daemon
output_logger="logger -et '${RC_SVCNAME}'"
error_logger="logger -et '${RC_SVCNAME}' -p3"

depend() {
	use logger
}
```

This forwards stdout and stderr to syslog. `-p3` sets the priority level to error.


tomalok:
> Implements #644. Allows redirecting process stdin and stdout to another process, just like is already possible with start-stop-daemon. Also addresses #341 and #118.

pretty sure you mean **stderr** and **stdout** here.

Higgs1:
Thank you! I think I've fixed it now. That was such an embarrassing mistake.

src/supervise-daemon/meson.build
```diff
@@ -1,5 +1,5 @@
 executable('supervise-daemon',
-  ['supervise-daemon.c', misc_c, plugin_c, schedules_c, usage_c, version_h],
+  ['supervise-daemon.c', '../start-stop-daemon/pipes.c', misc_c, plugin_c, schedules_c, usage_c, version_h],
```
ncopa, line 2: I wonder if would make sense to move `../start-stop-daemon/pipes.c` to `shared/pipes.c`?
```suggestion
  ['supervise-daemon.c', pipes_c, misc_c, plugin_c, schedules_c, usage_c, version_h],
```

ncopa:
```diff
diff --git a/src/shared/meson.build b/src/shared/meson.build
index b80b242e..04c70fef 100644
--- a/src/shared/meson.build
+++ b/src/shared/meson.build
@@ -10,6 +10,10 @@ schedules_c = files([
   'schedules.c',
   ])
 
+pipes_c = files([
+  'pipes.c',
+  ])
+
 if selinux_dep.found()
   selinux_c = files([
     'selinux.c',
diff --git a/src/start-stop-daemon/pipes.c b/src/shared/pipes.c
similarity index 100%
rename from src/start-stop-daemon/pipes.c
rename to src/shared/pipes.c
diff --git a/src/start-stop-daemon/pipes.h b/src/shared/pipes.h
similarity index 100%
rename from src/start-stop-daemon/pipes.h
rename to src/shared/pipes.h
diff --git a/src/start-stop-daemon/meson.build b/src/start-stop-daemon/meson.build
index d363ff94..6cae791d 100644
--- a/src/start-stop-daemon/meson.build
+++ b/src/start-stop-daemon/meson.build
@@ -1,5 +1,5 @@
 executable('start-stop-daemon',
-  ['start-stop-daemon.c', 'pipes.c', misc_c, schedules_c,
+  ['start-stop-daemon.c', pipes_c, misc_c, schedules_c,
        selinux_c, usage_c, version_h],
   c_args : [cc_audit_flags, cc_branding_flags, cc_pam_flags, cc_cap_flags, cc_selinux_flags],
   link_with: [libeinfo, librc],
```

Higgs1:
Hi @ncopa , I'm a big ~~fan~~ enjoyer of Alpine Linux. 

I was originally trying to keep the PR to a minimum, but I agree that moving it to the "shared" directory makes a lot more sense. This is the first PR I've submitted to a project of uh... this caliber, so I'm still learning all the ins and outs. I've updated it with your suggestion, thank you!

PR #667, opened 2023-11-08
symlink rather then duplicate multicall binaries

jsbronder:
This was raised here, https://github.com/jsbronder/meta-openrc/pull/29.  As I said there, I think these changes are better done upstream if found acceptable instead of carrying patches or manually creating symlinks post-install.

> There is a bunch of binaries in OpenRC that are essentially multicall binaries, but redundantly compiled multiple times:
>
> https://github.com/OpenRC/openrc/blob/0.45.2/src/einfo/meson.build
> https://github.com/OpenRC/openrc/blob/0.45.2/src/service/meson.build
> https://github.com/OpenRC/openrc/blob/0.45.2/src/mark_service/meson.build
> https://github.com/OpenRC/openrc/blob/0.45.2/src/value/meson.build
>
> By symlinking them to one 'main' multicall binary per group, we can reduce rootfs size. As the minimum ARM binary size is pretty big thanks to default 64 kiB alignment in newer binutils, this can be pretty sizable. This saved about 8MB on my test build.

### DIFF
--- a/src/einfo/meson.build
+++ b/src/einfo/meson.build
@@ -1,6 +1,5 @@
 einfo_execs = [
   'einfon',
-  'einfo',
   'ewarnn',
   'ewarn',
   'eerrorn',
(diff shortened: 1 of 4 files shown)


floppym (changes requested):
meson will dereference the symlinks on installation.

However, we can use the [install_symlink](https://mesonbuild.com/Reference-manual_functions.html#install_symlink) function instead.

floppym:
I created #712 and #713 to implement the missing pieces using hard links or symlinks.

jsbronder:
> I created #712 and #713 to implement the missing pieces using hard links or symlinks.

Great, thanks for taking this on!

PR #668, opened 2023-11-12
Skip already processed files in rc_service_daemon_set

keltar:
Fixes the problem described in https://bugs.gentoo.org/916947 - start-stop-daemon hangs in infinite loop when stopping some daemons on linux 6.6+

It appears linux 6.6 reworked tmpfs, and since then it triggers this problem in openrc:
When iterating over files via `readdir`, running `rename` on a file could result in reading the same file again with next `readdir` call.

The Open Group manual for `readdir` explicitly states "If a file is removed from or added to the directory after the most recent call to [opendir()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/opendir.html) or [rewinddir()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/rewinddir.html), whether a subsequent call to readdir() returns an entry for that file is unspecified.". Linux man page don't seem to mention that, but don't seem to say anything to contradict that either. So I presume we can't rely on some specific behaviour here.

### DIFF
--- a/src/librc/librc-daemon.c
+++ b/src/librc/librc-daemon.c
@@ -402,7 +402,7 @@ rc_service_daemon_set(const char *service, const char *exec,
 	bool retval = false;
 	DIR *dp;
 	struct dirent *d;
-	RC_STRINGLIST *match;
+	RC_STRINGLIST *match, *renamelist;
 	int i = 0;
 	FILE *fp;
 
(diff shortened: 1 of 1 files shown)


thesamesam:
Thank you for looking at this. I'd hoped to the other day but hadn't had a chance.

thesamesam:
Some brief comments:
* Could you add the text from the PR into the commit message?
* Could you add a `Bug:` tag in the footer of the commit message?

PR #672, opened 2023-11-25
Take back -DPREFIX compiler flag when switched to meson

littlewu2508:
This is a continuation of https://github.com/gentoo/gentoo/pull/31017

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -205,6 +205,10 @@ if cc.has_function('strlcpy', prefix: '#define _GNU_SOURCE\n#include <string.h>'
   add_project_arguments('-DHAVE_STRLCPY', language: 'c')
 endif
 
+if rootprefix != ''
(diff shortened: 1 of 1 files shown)


vapier:
what does "take back" mean ?  what is this fixing ?

the `PREFIX` define is barely checked in the source tree.  seems like we should be able to delete it entirely.

littlewu2508:
> what does "take back" mean ? what is this fixing ?
> 
> the `PREFIX` define is barely checked in the source tree. seems like we should be able to delete it entirely.

Well, currently without using `PREFIX` macro when compiling cause openrc to misbehave on prefix systems (including https://github.com/gentoo/gentoo/pull/31017#issue-1708542948, it seems that openrc of prefix is locating init files in `/etc` rather than `${EPREFIX}/etc`).

It would be better if we can drop this macro and fix the bug of openrc on prefix system.

PR #674, opened 2023-11-27
meson.build: fix uclibc build

ffontaine:
Fix the following uclibc build failures raised with meson because `kill` and `fileno` are not standard functions:

```
../src/rc-abort/rc-abort.c: In function 'main':
../src/rc-abort/rc-abort.c:27:21: error: implicit declaration of function 'kill'; did you mean 'killpg'? [-Werror=implicit-function-declaration]
   27 |                 if (kill(pid, SIGUSR1) != 0)
      |                     ^~~~
      |                     killpg

../src/libeinfo/libeinfo.c: In function 'colour_terminal': ../src/libeinfo/libeinfo.c:319:26: error: implicit declaration of function 'fileno' [-Werror=implicit-function-declaration]
  319 |         if (f && !isatty(fileno(f)))
      |                          ^~~~~~

../src/librc/librc-misc.c: In function 'rc_getfile': ../src/librc/librc-misc.c:79:14: error: implicit declaration of function 'fileno'; did you mean 'd_fileno'? [-Werror=implicit-function-declaration]
   79 |         fd = fileno(fp);
      |              ^~~~~~
      |              d_fileno

../src/librc/librc-daemon.c: In function 'rc_service_daemons_crashed': ../src/librc/librc-daemon.c:633:37: error: implicit declaration of function 'kill'; did you mean 'killpg'? [-Werror=implicit-function-declaration]
  633 |                                 if (kill(pid, 0) == -1 && errno == ESRCH)
      |                                     ^~~~
      |                                     killpg
```

Fixes:
 - http://autobuild.buildroot.org/results/494ef392a971ddb3c5c7b01e0149c6439018dbe7

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -149,7 +149,7 @@ else
 endif
 
 if os == 'Linux' or os == 'GNU-kFreeBSD'
-  cc_os_flags = ['-D_DEFAULT_SOURCE']
+  cc_os_flags = ['-D_DEFAULT_SOURCE', '-D_GNU_SOURCE']
 elif os == 'FreeBSD'
   cc_os_flags = ['-D_BSD_SOURCE']
 elif os == 'GNU'


vapier (changes requested):
we aren't enabling `-D_GNU_SOURCE` everywhere.  this sounds like something else is broken.  `kill()` is defined in `signal.h` which we're including already in these files.

ffontaine:
To use `kill`, including `signal.h` is not enough, `_POSIX_C_SOURCE` must also be defined: https://man7.org/linux/man-pages/man2/kill.2.html. Same thing for fileno: https://man7.org/linux/man-pages/man3/fileno.3.html.

`_POSIX_C_SOURCE` was defined in the "old" makefile-based build system: https://github.com/OpenRC/openrc/commit/f4e2142089487b7f5c99da949d7bf8d038724df7 but this get "lost" when moving to meson.

I updated the PR to replace `-D_GNU_SOURCE`  by `-D_POSIX_C_SOURCE=200809L`.

Moreover, on uclibc, `getrandom `is only available if `_GNU_SOURCE` is defined so I added this define in `seedrng.c`

vapier:
> To use kill, including signal.h is not enough, _POSIX_C_SOURCE must also be defined

that is not accurate.  those defines are a GNUism invented by glibc.  glibc has always defaulted to "recent" levels of POSIX APIs and do not force people to always define things to get them.  in particular, glibc (and opnerc) default to POSIX.1-2008 which includes `kill()` (see how `_DEFAULT_SOURCE` is checked in sys/features.h).  if your version of uClibc is attempting to replicate glibcisms by checking `_DEFAULT_SOURCE` but then setting `_POSIX_C_SOURCE` to something ancient and causing `kill()` prototype to not be exposed, then that's a bug in uClibc, not openrc.

> _POSIX_C_SOURCE was defined in the "old" makefile-based build system: https://github.com/OpenRC/openrc/commit/f4e2142089487b7f5c99da949d7bf8d038724df7 but this get "lost" when moving to meson.

glibc defaults to `_POSIX_C_SOURCE=200809L` for us now which would make the setting redundant.  i don't think re-adding it is useful.  any C library that doesn't support or default to POSIX.1-2008 is probably not worth supporting at this point.

maybe we could add a check to some core code that asserts that `_POSIX_C_SOURCE >= 200809` and abort the build if it is not ...

> Moreover, on uclibc, getrandom is only available if _GNU_SOURCE is defined so I'll add this define in seedrng.c

this doesn't make sense ... `sys/random.h` is a GNUism header.  providing it, but requiring people define `_GNU_SOURCE` to get its definitions when including it, is ridiculous.  the header is def not in POSIX, so if you have a reference to show it's anything other than a GNUism, please link it.

PR #676, opened 2023-12-04
change info words "remove" to "delete" in delete sub-command

### LINKED ISSUE #664: rc-update requires the "delete" sub-command, but says it "removed"
Hello OpenRC! :-)

I just realized that in order to remove a service from a runlevel (I'm on Alpine if that matters) I need to say
rc-update delete foo bar
and then rc-update tells me it *removed* foo from bar (which is precisely what I want to do). However, saying
rc-update remove foo bar
gives me a usage error instead. Can we anyhow solve this inconsistency? It's not much, but just an annoyance for (I assume) a lot of people!?

Best,

~ Armin

mingzym:
This close #664

### DIFF
--- a/src/rc-update/rc-update.c
+++ b/src/rc-update/rc-update.c
@@ -88,7 +88,7 @@ delete(const char *runlevel, const char *service)
 
 	errno = 0;
 	if (rc_service_delete(runlevel, service)) {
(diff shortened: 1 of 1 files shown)


vapier:
sure, why not

Issue #675, opened 2023-11-28
swclock init: may send the clock backwards of current deptree reference.


macmpi:
At first run (`openrc-shutdowntime` does not exist), `swclock` sets time according to `openrc-run` file mtime.

```
	if ! swclock "${swclock_file}" 2> /dev/null; then
	swclock --warn @SBINDIR@/openrc-run
```
Until then mtime reference for `deptree` [was based on latest mtime](https://github.com/OpenRC/openrc/blob/master/src/librc/librc-depend.c#L688) found in `RC_INITDIR`, `RC_CONFDIR`, `RC_CONF` & al.
In some circumstances it may actually happen `openrc-run` mtime predates such calculated `deptree` date and therefore cause transitional havoc in service starts after `swclock` kicks-in (flock issues): one exemple is described [there](https://gitlab.alpinelinux.org/alpine/aports/-/issues/15463).

What would be the best approach to avoid that?
- set same arbitrary mtime for all open-rc files?
- add `openrc-run` in the list of `deptree` inspected locations to set initial mtime?
- have `swclock` use another default openrc file from `deptree` inspected locations, to set its first-time time reference in consistent manner?

macmpi:
Thanks for feedback.

Would it then make more sense to add `openrc-run` in the [list of files](https://github.com/OpenRC/openrc/blob/master/src/librc/librc-depend.c#L688) whose mtime is being monitored to set deptree reference then? (RC_INITDIR, RC_CONFDIR, RC_CONF,...), or use another reference file which is already in the monitored list (lowest key change)?

>clocks can easily be set incorrectly in the future and we need to pull it back.

Sure, it may happen for particular reasons.
However, currently it happens every time at first boot depending on historic changes of packaged files: seems a bit nondeterministic and unwanted: depending on history of  `openrc-run` clock may go forward or backwards, and eventually trigger deptree flock issues...
Alternatively packaged file mtime could be set as some arbitrary time (release date, time 0, etc...), but has some constraints.

Thoughts?

vapier:
the whole point of swclock is that the system doesn't have a hwclock to keep track of time.  it sounds like your system has a hwclock (otherwise, how would time be preserved across boots?), so why are you using swclock in the first place ?

macmpi:
No my system does not have `hwclock`, and uses `swclock`.

I came across that unexpected time-shifts, while looking at detailed rc logs trying to figure-out why some service eventually warned about flock issue at start on a fresh system.
[...]

PR #678, opened 2023-12-08
init.d/swclock: use current time instead of default reference file

macmpi:
A default reference file mtime may send the clock backwards of current deptree mtime.
Fixes https://github.com/OpenRC/openrc/issues/675


vapier (changes requested):
please see comments in the referenced issue.  let's keep discussion there for now.

Issue #680, opened 2023-12-15
openrc and docker cgroup v2 path conflict


ncopa:
I have been trying to solve a weird downstream bug with `rc_group_mode=unified` which recently was enabled as the default.

To reproduce the issue, start docker as an openrc service: `rc-service docker start`

Then try run a docker container with memory limits:
```console
alp319:~$ docker run --rm --memory 128M alpine echo hello
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: unable to apply cgroup configuration: cannot enter cgroupv2 "/sys/fs/cgroup/docker" with domain controllers -- it is in domain threaded mode: unknown.
```

For some reason it ends up in `domain threaded` mode:
```console
alp319:~$ cat /sys/fs/cgroup/docker/cgroup.type 
domain threaded
```

Now try restart the service, and notice that it originally is not in `domain threaded` mode:
```console
alp319:~$ doas /etc/init.d/docker restart
doas (ncopa@alp319) password: 
 * Stopping Docker Daemon ...                                                                                                                                                              [ ok ]
 * Starting Docker Daemon ...                                                                                                                                                              [ ok ]
alp319:~$ cat /sys/fs/cgroup/docker/cgroup.type 
domain
```

Run any container and, it switches back to `domain threaded` mode (it works without memory limits):
```console
alp319:~$ docker run --rm alpine echo hello
hello
alp319:~$ cat /sys/fs/cgroup/docker/cgroup.type 
domain threaded
```

A workaround for this is to set `--cgroup-parent` to something other than `docker`.

When trying to find out what sets this, I realized that openrc will create and use [`/sys/fs/cgroup/$SVCNAME`](https://github.com/OpenRC/openrc/blob/bfdafe4463c4886b0d51dc1a426ed5345d5e0e04/sh/rc-cgroup.sh#L194). In the docker case this ends up with `/sys/fs/cgroup/docker`. Now, the default --cgroup-parent in docker is ["docker"](https://github.com/moby/moby/blob/d90e5f283a7846c3678e4731fee337fcbf283ef2/cmd/dockerd/daemon_unix.go#L105), so  docker tries to manage its containers under `/sys/fs/cgroup/docker`. But since openrc has its stuff there,it gets confused, and sets `domain threaded mode` (I believe).

To properly fix this, I believe we should *not* use `/sys/fs/cgroup/$SVCNAME` to avoid the conflict with docker. We could use something like:
[...]

PR #681, opened 2023-12-15
sh/rc-cgroup.sh: add openrc. prefix the cgroupv2 path

ncopa:
Some services, like docker, creates and manages /sys/fs/cgroup/<service> themselves. Avoid conflict with the openrc created cgroup path by adding a `openrc.` prefix.

Fixes: https://github.com/OpenRC/openrc/issues/680

### DIFF
--- a/sh/rc-cgroup.sh
+++ b/sh/rc-cgroup.sh
@@ -35,7 +35,7 @@ cgroup_get_pids()
 	cgroup_pids=
 	cgroup_procs="$(cgroup2_find_path)"
 	if [ -n "${cgroup_procs}" ]; then
-		cgroup_procs="${cgroup_procs}/${RC_SVCNAME}/cgroup.procs"
+		cgroup_procs="${cgroup_procs}/openrc.${RC_SVCNAME}/cgroup.procs"
 	else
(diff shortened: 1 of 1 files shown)


donob4n:
It seems that it needs more changes for be fully coherent.

Forza-tng:
Perhaps add a rc.conf variable for the `openrc.` prefix?

williamh:
Hi,

I just applied this patch locally to test it and restarted a service.
I found that the old group for the service existed along side the new group after the restart.
e.g. /sys/fs/cgroup/<servicename> and /sys/fs/cgroup/openrc.<servicename> both existed.
I was able to remove the old cgroup successfully.
Should we add something to this pr to removeit?

ncopa:
> Hi,
> 
> I just applied this patch locally to test it and restarted a service. I found that the old group for the service existed along side the new group after the restart. e.g. /sys/fs/cgroup/ and /sys/fs/cgroup/openrc. both existed. I was able to remove the old cgroup successfully. Should we add something to this pr to removeit?

Not sure its worth it. Its only a one time problem during upgrade that a reboot will "fix".

Issue #689, opened 2024-02-04
openrc start-stop-daemon scheduler param is not working with musl libc


ipoupaille:
It is related to [https://gitlab.alpinelinux.org/alpine/aports/-/issues/15749](https://gitlab.alpinelinux.org/alpine/aports/-/issues/15749)

sched_setscheduler seems not to be implemented in musl.
I propose to replace with pthread methods: [https://gitlab.alpinelinux.org/alpine/aports/-/issues/15749#note_374795](https://gitlab.alpinelinux.org/alpine/aports/-/issues/15749#note_374795)

PR #690, opened 2024-02-05
start-stop-daemon: fix setting scheduler with musl

ncopa:
musl sched_setscheduler() is only a stub in musl and the glibc implementation does "the wrong thing". Use pthread_setschedparam instead.

ref: https://www.openwall.com/lists/musl/2016/03/01/5
fixes: https://github.com/OpenRC/openrc/issues/689

### DIFF
--- a/src/start-stop-daemon/start-stop-daemon.c
+++ b/src/start-stop-daemon/start-stop-daemon.c
@@ -31,6 +31,7 @@
 #include <getopt.h>
 #include <limits.h>
 #include <grp.h>
+#include <pthread.h>
 #include <pwd.h>
 #include <sched.h>
 #include <signal.h>
@@ -1126,7 +1127,7 @@ int main(int argc, char **argv)
 			if (sched_prio == -1)
 				sched.sched_priority = sched_get_priority_min(scheduler_index);
 
-			if (sched_setscheduler(mypid, scheduler_index, &sched))
+			if (pthread_setschedparam(pthread_self(), scheduler_index, &sched))
 				eerrorx("Failed to set scheduler: %s", strerror(errno));
 		} else if (sched_prio != -1) {
 			const struct sched_param sched =  {.sched_priority = sched_prio};
--- a/src/start-stop-daemon/meson.build
+++ b/src/start-stop-daemon/meson.build
@@ -3,7 +3,7 @@ executable('start-stop-daemon',
 	selinux_c, usage_c, version_h],
   c_args : [cc_audit_flags, cc_branding_flags, cc_pam_flags, cc_cap_flags, cc_selinux_flags],
   link_with: [libeinfo, librc],
-  dependencies: [audit_dep, dl_dep, pam_dep, cap_dep, pam_misc_dep, util_dep, selinux_dep, crypt_dep],
+  dependencies: [audit_dep, dl_dep, pam_dep, cap_dep, pam_misc_dep, util_dep, selinux_dep, crypt_dep, dependency('threads')],
   include_directories: [incdir, einfo_incdir, rc_incdir],
   install: true,
   install_dir: sbindir)


ncopa:
Seems like we either need to add `-lpthread` for FreeBSD or have an `#ifdef __linux__`. Not sure what is preferred.

src/start-stop-daemon/start-stop-daemon.c
```diff
@@ -1123,7 +1124,11 @@ int main(int argc, char **argv)
 			if (sched_prio == -1)
 				sched.sched_priority = sched_get_priority_min(scheduler_index);
 
+#ifdef __linux__
```
vapier, line 1127: why is it tied to `__linux__` ?  this API is in POSIX:
https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_setschedparam.html

ncopa: To quote @richfelker

> The reason it doesn't do anything is that Linux does not provide a way
> to set scheduling parameters for a _process_, only for threads. The
> sched_setscheduler syscall is documented as taking a pid but actually
> takes a thread id and only operates on that thread. glibc just ignores
> this and provides sched_* functions that do the wrong thing.

I read that as it is a problem with `sched_setscheduler` on Linux and other OSes may not need this workaround.

vapier: for a single threaded process, like s-s-d, pid==tid.  so it sounds like sched_setscheduler already works correctly.

richfelker: > for a single threaded process, like s-s-d, pid==tid. so it sounds like sched_setscheduler already works correctly.

It's a no-op/always-fail, which *is* working correctly, per the specified requirements, but it's probably not what you want. Calling the `sched_setscheduler` function would set the *process* scheduling, but there's no such thing on Linux (it's optional POSIX functionality). However the syscall by that name, and the glibc `sched_setscheduler` function, target a thread, behaving like `pthread_setschedparam` but with a kernel TID as the target instead of a `pthread_t`. They do not target a process. The glibc function probably does *what you want*, but that's because it's doing *the wrong thing*.

If that is what you want, the right way to do it is by calling `pthread_setschedparam` on `pthread_self()`.

richfelker: > > It's a no-op/always-fail
> 
> you've made that choice in your implementation

The only choice made here was to follow the requirements of the standard.

> > Calling the sched_setscheduler function would set the process scheduling, but there's no such thing on Linux
> 
> this seems to be bleeding into internal details. from a single-threaded process pov, the API works with glibc (and other C libraries afaik). the program wants to set scheduling details for itself, and sched_setscheduler does that.

I suggest you go read the [Thread Scheduling](https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_09_04) and [Process Scheduling](https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_08_04) sections of POSIX. The POSIX model has two different priority systems:

1. one where *processes* contend for execution resources, and after the result of that determines which process gets time to run, the threads within that process scope contend within each other based on their thread priorities, and

2. one where all threads, regardless of which process they belong to contend in a global scope

The first model is not supported by Linux. In the absence of support for that model, the `sched_setscheduler` priority is a no-op; it is specified only to have an effect when `PTHREAD_SCOPE_PROCESS` scheduling model is active, which is never possible on Linux. Threads in the `PTHREAD_SCOPE_SYSTEM` model contend only with other threads, based on their thread priority, not influenced in any way by a process priority.

> i get that in a multithreaded process, there aren't separate process & thread level scheduling in Linux that threads would be able to pick between via contention scope. but we aren't talking about multithreaded processes here.

It doesn't matter. The single existant thread in a single-threaded process is a thread, and contends according to the rules for a thread.

> > The glibc function probably does what you want, but that's because it's doing the wrong thing.
> 
> it's a funny argument when you then advocate for changing openrc to do exactly what glibc is doing, but with a diff function name.

It's not funny. The root problem here is that glibc is using the wrong name for the operation you want. What you want to do is to set the scheduling parameters of the initial thread of the process being spawned so that it (and any other threads created by it which inherit those scheduling parameters) will contend at the desired priority. `pthread_setschedparam` is the function which does that, per the standard. `sched_setscheduler` is not.

richfelker: > again, what you describe is internal details bleeding out. in a single threaded process, which is the only thing we're concerned with here, linux+glibc's sched_setscheduler behaves correctly.

Once again, no it does not. It does what you want, which is very different from behaving correctly. Behaving correctly would be setting the process priority, which is specified to be **ignored** unless one or more threads in the process have had their scheduling scope set to `PTHREAD_SCOPE_PROCESS` (or if this is the implementation-defined default). Since Linux does not support `PTHREAD_SCOPE_PROCESS`, this is never the case.

> this PR proposes making changes for all Linux systems in a way that doesn't work, and is incomplete even for musl -- i'm guessing sched_setparam is similarly broken. if we wanted to make a change in here to accommodate musl, we would need to probe the C library to figure out it's actually musl, and then put this rewrite logic into a quirks file. inline ifdefs make the code less maintainable (yes, i know s-s-d has that already).

No, all you need to do is replace the `sched_setscheduler` call targering the caller's pid with `pthread_setschedparam` targeting `pthread_self()`. This is the portable (works on glibc and everywhere else) way to do what you're currently achieving by assuming glibc's implementation of `sched_setscheduler`, which does the wrong (but wanted) thing. Note that the glibc folks even acknowledged that it's doing the wrong thing and temporarily fixed it, but reverted the fix because it was a breaking change for applications which expect the wrong glibc behavior...

ncopa: > glibc's implementation of `sched_setscheduler`, which does the wrong (but wanted) thing. Note that the glibc folks even acknowledged that it's doing the wrong thing and temporarily fixed it, but reverted the fix because it was a breaking change for applications which expect the wrong glibc behavior...

Does this mean that FreeBSD does something we don't want in this case? (assuming it follows the POSIX spec correctly?)

It sounds like it would be better to use `pthread_setchedparam` also for FreeBSD (and add `-lpthread`)

richfelker: > Does this mean that FreeBSD does something we don't want in this case? (assuming it follows the POSIX spec correctly?)

Maybe. It's possible they support the process scheduling scope and use it by default, in which case `sched_setscheduler` would have a *different* but *possibly more desirable* behavior than the glibc behavior. Or it's possible that it does something different but not-quite-right. I don't know enough about FreeBSD (or other BSDs for that matter) kernel or threads implementation to answer.

> It sounds like it would be better to use `pthread_setchedparam` also for FreeBSD

It would at least ensure you're getting more uniform behavior.

> (and add `-lpthread`)

I suspect this is the part they don't want to do, for whatever reason. I guess it could be avoided on Linux by special-casing Linux and using the raw syscall there, but yes just using `-lpthread` and `pthread_setscheduler` would be the clean cross-platform solution.

ncopa:
I updated it to use `pthread_setschedparam` everywhere so we avoid depending on implementation specific behaviour.

src/start-stop-daemon/start-stop-daemon.c
navi-desu: reading over this, seems like indeed glibc does something wrong that just happens to work

i'm in favor of using the pthread solution as, even tho we're single threaded, it's the more sound solution (does what we want, with the expected standardized behavior)

will take a look into the code for this soon

navi-desu:
thanks!

PR #693, opened 2024-02-21
start-stop-daemon:  support cpu scheduler env variable

Forza-tng:
Introduced `SSD_CPUSCHEDULER` and `SSD_CPUSCHEDULER_PRIO` environment variables to complement the `--scheduler` command line argument.

### DIFF
--- a/man/start-stop-daemon.8
+++ b/man/start-stop-daemon.8
@@ -203,6 +203,14 @@ option takes precedence.
 can also set the scheduling priority of the daemon, but the command line
 option takes precedence.
 .Pp
+.Va SSD_CPUSCHEDULER
+can also set the CPU scheduler of the daemon, but the command line
+option takes precedence.
+.Pp
+.Va SSD_CPUSCHEDULER_PRIO
+can also set the CPU scheduling priority of the daemon, but the command line
+option takes precedence.
+.Pp
 .Va SSD_OOM_SCORE_ADJ
 can also set the OOM score adjustment of the daemon, but the command line
 option takes precedence.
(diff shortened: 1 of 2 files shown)


src/start-stop-daemon/start-stop-daemon.c
```diff
@@ -379,6 +379,19 @@ int main(int argc, char **argv)
 			eerror("%s: invalid oom_score_adj `%s' (SSD_OOM_SCORE_ADJ)",
 			    applet, tmp);
 
+	// Handle SSD_CPUSCHEDULER for scheduler type
+	if ((tmp = getenv("SSD_CPUSCHEDULER"))) {
+		scheduler = strdup(tmp);
```
N-R-K, line 384: Why `strdup`?

man/start-stop-daemon.8
```diff
@@ -203,6 +203,14 @@ option takes precedence.
...
 option takes precedence.
 .Pp
+.Va SSD_CPUSCHEDULER
+can also set the CPU scheduler of the daemon, but the command line
+option takes precedence.
+.Pp
+.Va SSD_CPUSCHEDULER_PRIO
+can also set the CPU scheduling priority of the daemon, but the command line
+option takes precedence.
+.Pp
```
N-R-K, line 213: Also should be added to `supervise-deamon` for consistency.

N-R-K (changes requested):

src/start-stop-daemon/start-stop-daemon.c
N-R-K:
```
       The  implementation  of  getenv() is not required to be reentrant.  The string pointed to by the return value of getenv() may be statically allocated,
       and can be modified by a subsequent call to getenv(), putenv(3), setenv(3), or unsetenv(3).
```

Okay, I see.

PR #696, opened 2024-03-25
meson: drop broken split-usr handling

thesamesam:
meson: drop broken split-usr handling

Two issues here:

* The 'split-usr' meson option wasn't doing anything, it tried to check
  if /bin was a symlink, but nothing acted on this information.

* The actual rootprefix default was decided based on whether /bin was a symlink
  which is flaky if e.g. building on a merged-usr system for use on a non-merged-usr
  system.

People can set -Drootprefix=/usr if they wish.

There's no real advantage to installing to /usr over / as the compat. symlinks
are really here to stay. If someone really does care about this, they can bring
it back and do it properly, but it doesn't seem worth it to me at all.

Bug: https://bugs.gentoo.org/927776
Fixes: cc0037e9caaee05af0fdedafc5798c2a7aa9bdb8
Fixes: f2362cc277023550b2482215b4a1cd7142639427

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -83,16 +83,9 @@ else
   pkg_prefix = option_pkg_prefix
 endif
 
-if get_option('split-usr') == 'auto'
-  split_usr = run_command('test', '-L', '/bin', check: false).returncode() != 0
-else
-  split_usr = get_option('split-usr') == 'true'
-endif
-
 rootprefix = get_option('rootprefix')
-rootprefix_default = fs.is_symlink('/bin') ? '/usr' : '/'
 if rootprefix == ''
-rootprefix = rootprefix_default
+rootprefix = '/'
 endif
 
 bindir = rootprefix / get_option('bindir')
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -26,9 +26,6 @@ option('selinux', type : 'feature', value : 'auto',
   description : 'enable SELinux support')
 option('shell', type : 'string', value : '/bin/sh',
   description : 'Default posix compatible shell')
-option('split-usr', type : 'combo',
-  choices : ['auto', 'true', 'false'],
-  description : '''/bin, /sbin aren't symlinks into /usr''')
 option('sysvinit', type : 'boolean', value : false,
   description : 'enable SysVinit compatibility (linux only)')
 option('termcap', type : 'combo',


eli-schwartz:
> There's no real advantage to installing to /usr over / as the compat. symlinks are really here to stay. If someone really does care about this, they can bring it back and do it properly, but it doesn't seem worth it to me at all.

It is required for non-Gentoo package managers on distros that only support one or the other of split-usr/ merged-usr, and have chosen merged-usr.

Symlinks may be here to stay, but not all package managers perform install-time resolution of symlinks when calculating whether or not the package queued for installation has conflicts with existing files on disk e.g. owned by another package. In such a scenario, you will easily find yourself installing files to /sbin/* and having them conflict another "file", the /sbin symlink.

It's therefore correct to always *refer* to the file by its canonical portable name, but when copying files over to DESTDIR a bit more care is needed.

This isn't an issue when directly installing without a staged DESTDIR, and it isn't a problem if you work around the problem by `mv`ing files around after running meson install.

thesamesam:
That's a fair point. But for such cases, they can set `-Drootprefix`, I think. I have no idea how well that option works, but this PR doesn't make that worse either.

meson.build
```diff
@@ -83,14 +83,8 @@ else
...
-else
-  split_usr = get_option('split-usr') == 'true'
-endif
-
 rootprefix = get_option('rootprefix')
-rootprefix_default = fs.is_symlink('/bin') ? '/usr' : '/'
+rootprefix_default = '/'
 if rootprefix == ''
 rootprefix = rootprefix_default
 endif
```
eli-schwartz, line 90: Mysteriously, this autodetection is always overridden by gentoo by passing `-Drootprefix="${EPREFIX}"`, which leaves me wondering where the actual source of the linked bug is.

floppym: > Mysteriously, this autodetection is always overridden by gentoo by passing -Drootprefix="${EPREFIX}", which leaves me wondering where the actual source of the linked bug is.

That only affects line 86. On a normal system, ${EPREFIX} is empty so rootprefix is also empty.

The condition on line 94 is satisfied, so line 95 sets rootprefix = rootprefix_default.

rootprefix_default is the auto-detected value based on /bin.

meson.build
```diff
@@ -83,14 +83,8 @@ else
...
 
-if get_option('split-usr') == 'auto'
-  split_usr = run_command('test', '-L', '/bin', check: false).returncode() != 0
-else
-  split_usr = get_option('split-usr') == 'true'
-endif
-
 rootprefix = get_option('rootprefix')
-rootprefix_default = fs.is_symlink('/bin') ? '/usr' : '/'
+rootprefix_default = '/'
```
floppym, line 87: This variable isn't necessary. Just assign rootprefix = '/' below.

floppym (changes requested):
This makes the logic simpler without really losing any functionality, so it makes sense to me.

meson.build
thesamesam: It could be done, but I don't plan on doing it myself. I've updated the commit message a bit.

Issue #662, opened 2023-10-17
Parallezing `do_unmount`


anthonyryan1:
I've been exploring shutdown times on some OpenRC servers with up to 100 hard drives connected to each machine.

I found that one of the primary bottlenecks is the sequential nature of `do_unmount` in `rc-mount.sh`. On a machine with hundreds of mount points, it will iterate through them sequentially one after another.

My first idea was to run each unmount command in the background with `&`. But that would break assumptions about unmount order, potentially creating unmount errors for people with filesystems mounted inside racing their parent to unmount.

I'm thinking the correct way to make `do_unmount` run in parallel would be to use `fstabinfo --passno ''` to get all the pass numbers, then iterate from the largest passno down to 0 using `fstabinfo --passno =N` to decide which drives can be unmounted in parallel during each batch.

I'm currently looking for feedback and concerns before creating a pull request.

vapier:
i don't think passno works.  that assumes all mounts are in fstab to begin with, and that's def not a requirement.

this should be a compiled program so that it can actually build a graph and then process it as such using threads.

i could see there being a benefit to tickling the underlying devices in parallel in case they went to sleep -- waking back up to unmount is def slow.  i'm not sure how much overhead there actually is though if all the drives are active and we just ran umount syscall on them in serial.

Tommimon:
I'd like to give this a try and implement a c program that uses `mountinfo` output to build a dependency graph. The program output will contain a dependency list for each mount point, those lists can of course be nested. The shell script will then run the unmount operation in parallel starting from the leaves lists and proceeding to other elements when all dependencies are met.

Do you think there's a better suitable approach for this specific issue?

Tommimon:
To be sure I understood correctly, you suggest a `unmount_all.c` which, when executed once, will do:
1) retrieve data from `mountinfo.c`.
2) create a dependency graph on all the mount points present.
3) unmount them in parallel when possible.

If this is what you suggest, I think is a good Idea to avoid mixing C with shell when is not required.

doot-64:
We have a proposed solution [here](https://github.com/OpenRC/openrc/pull/698) if anyone wants to take a look.

N-R-K:
I have a WIP multi-process based parallel unmount draft. It's currently missing the fuser stuff though.

Initial impression is that the slowness you're getting is likely coming from the "unhappy" path where it fails to unmount and goes to sleep.

For testing, I used this script to create a bunch of mounts:

```
#!/bin/sh

[...]

PR #698, opened 2024-03-26
Parallel unmount

Tommimon:
Removing `rc-mount.sh` script and all the references to it. Replacing the `do_unmount` shell function with the `do_unmount.c` program.

The new program implements the same functionalities with the added feature of parallel unmounting when possible:

- Unmount every mount point in a different thread.
- Wait for every mount point in a subdirectory to be unmounted before proceeding with the current one.
- Never unmount two shared mount points at the same time.
- Check if shared mount point is still present before unmounting.
- Retrieve list of shared mount points from `/proc/1/mountinfo`.


src/do_unmount/do_unmount.c
```diff
@@ -0,0 +1,382 @@
...
+
+    /* Allocate memory for the command */
+    cmd = xmalloc(length * sizeof(char));
+    sprintf(cmd, "%s", command);
+
+    /* Craft the command with all passed arguments */
+    for (i = 0; i < argc; i++) {
+        strcat(cmd, " \"");
+        strcat(cmd, argv[i]);
+        strcat(cmd, "\"");
```
N-R-K, line 85: This should *really* be doing fork+exec (or `posix_spawn()` if you're familiar with that). This type of simplistic escapes are not robust, will fail if the argument contains special character or worse execute random stuff in case of `$(...)` or backticks.

src/do_unmount/do_unmount.c
```diff
@@ -0,0 +1,382 @@
...
+            i++;
+        }
+    }
+
+    /* Close file and free memory */
+    fclose(fp);
+    if (line)
+        free(line);
+    if (path)
+        free(path);
```
N-R-K, line 135: `free` is no-op on `NULL` argument. No need to check before calling.

src/do_unmount/do_unmount.c
```diff
@@ -0,0 +1,382 @@
...
+    int size = 0;       /* Number of paths to unmount */
+    FILE *fp;           /* File pointer to the output of the command */
+    char *path = NULL;  /* Path to add to the list */
+    size_t len = 0;     /* Length of the line read */
+
+
+    /* Open the command for reading */
+    fp = popen_vec("mountinfo", argc-2, argv+2);
+    if (fp == NULL) {
+        printf("Failed to run mountinfo command, can't unmount anything!\n");
```
N-R-K, line 152: Errors should go to `stderr`.

src/do_unmount/do_unmount.c
```diff
@@ -0,0 +1,382 @@
...
+populate_unmount_list(RC_STRINGLIST **list, int argc, char **argv)
+{
+    int size = 0;       /* Number of paths to unmount */
+    FILE *fp;           /* File pointer to the output of the command */
+    char *path = NULL;  /* Path to add to the list */
+    size_t len = 0;     /* Length of the line read */
+
+
+    /* Open the command for reading */
+    fp = popen_vec("mountinfo", argc-2, argv+2);
```
N-R-K, line 150: Shouldn't we just reuse the c code in `mountinfo.c` instead of calling an process and parsing it's output?

src/do_unmount/do_unmount.c
```diff
@@ -0,0 +1,382 @@
...
+    f_kill = "-";
+    #endif
+
+    /* Default timeout value if environment variable is not set */
+    if (timeout == NULL)
+        timeout = "60";
+
+    /* Allocate memory for the commands */
+    xasprintf(&fuser_command, "timeout -s KILL %s fuser %s %s 2>/dev/null", timeout, f_opts, mount_point);
+    xasprintf(&kill_command, "fuser %sTERM -k %s \"%s\" >/dev/null 2>&1", f_kill, f_opts, mount_point);
```
N-R-K, line 199: Same issue with non-robust escape and potential random code execution here. Use fork+exec or posix_spawn.

N-R-K (changes requested):
Just some initial impression. Haven't dug too deeply or tested the code.

src/do_unmount/do_unmount.c
Tommimon: This requires to edit `mountinfo.c` and create a new `mountinfo.h`. Is this what you are suggesting?

src/do_unmount/do_unmount.c
Tommimon: Should I use `eerrorx` instead? Which also exists the program with an error code.

src/do_unmount/do_unmount.c
Tommimon: I didn't know, it's fixed now.

src/do_unmount/do_unmount.c
Tommimon: Note that [in the original discussion](https://github.com/OpenRC/openrc/issues/662) it was explicitly requested a multi-thread program, separate from `mountinfo.c` that accesses `mountinfo.c` information.

Tommimon: Now it is using code from libmountinfo.c

Tommimon:
A replacement for `popen` and `system` is needed to solve the highlighted security issues.

PR #705, opened 2024-05-11
Add foot to color capable term list

Richard-Rogalski:
foot is a wayland color capable terminal emulator with the default TERM being set to "foot"


navi-desu:
fixed in https://github.com/OpenRC/openrc/commit/f943bc4c9a495baaf4ae945e27060e3bf490e539, closing

thank you tho!

PR #707, opened 2024-05-21
openrc-shutdown: broadcast: avoid double fclose

floppym:
Initialize fd and tp to known invalid values, and reset them after
calling close/fclose. This is necessary since these are decleared as
static variables.

Note the close will will likely still be called twice: we close fd
before tp to avoid any buffered writes.

Bug: https://bugs.gentoo.org/923326

### DIFF
--- a/src/openrc-shutdown/broadcast.c
+++ b/src/openrc-shutdown/broadcast.c
@@ -131,8 +131,8 @@ void broadcast(char *text)
 	 * These are set across the sigsetjmp call, so they can't be stored on
 	 * the stack, otherwise they might be clobbered.
 	 */
-	static int fd;
-	static FILE *tp;
+	static int fd = -1;
+	static FILE *tp = NULL;
 
 	getuidtty(&user, &tty);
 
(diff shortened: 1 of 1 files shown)


N-R-K (approved):

src/openrc-shutdown/broadcast.c
```diff
@@ -193,10 +193,19 @@ void broadcast(char *text)
 			}
 		}
 		alarm(0);
-		if (fd >= 0)
+		/*
+		 * It is possible we got here because fputs or fflush got
+		 * interrupted by SIGALARM. Close the file descriptor first to
```
N-R-K, line 198: Doesn't this result in a double close of `fd`?
If blocking is the issue why not make the fd non-blocking?

N-R-K (changes requested):

src/openrc-shutdown/broadcast.c
floppym: Yes, it does result in a double close. I should update the summary: the bug is a double fclose, since that causes a double free.

The fd is already non-blocking, but the original author doesn't trust the kernel to respect that apparently. The whole SIGALARM path should really be stripped out and simplified.

floppym:
Closing in favor of #709.

PR #708, opened 2024-05-22
Podman detection

### LINKED ISSUE #695: OpenRC detect RC system type podman ?
Hello,

perhaps a stupid question regarding using openrc in a container environment.
(testing a image)

I noticed that /src/librc/librc.c#L305 currently check for container=docker and so on, but podman is not detected.

cat /proc/1/environ
container=podman...

My workaround is to create a /.dockerenv to use the keyword -container feature.

Is there a better solution, or would it be a good Idea to check for container=* .

openrc -S  returns nothing without creating a /.dockerenv file.

EDIT: (alternative detections: )
podman says its provide /run/.containerenv

Sylviettee:
Podman is another container manager akin to Docker, being detectable by checking for the presence of `/.run/containerenv`. Fixes #695.

### DIFF
--- a/src/librc/librc.c
+++ b/src/librc/librc.c
@@ -304,10 +304,10 @@ detect_container(const char *systype RC_UNUSED)
 		return RC_SYS_RKT;
 	else if (file_regex("/proc/1/environ", "container=systemd-nspawn"))
 		return RC_SYS_SYSTEMD_NSPAWN;
-	else if (exists("/.dockerenv"))
-		return RC_SYS_DOCKER;
 	else if (exists("/run/.containerenv"))
 		return RC_SYS_PODMAN;
+	else if (exists("/.dockerenv"))
+		return RC_SYS_DOCKER;
 	/* old test, I'm not sure when this was valid. */
 	else if (file_regex("/proc/1/environ", "container=docker"))
 		return RC_SYS_DOCKER;
--- a/src/librc/librc.c
+++ b/src/librc/librc.c
@@ -257,6 +257,8 @@ detect_container(const char *systype RC_UNUSED)
 		       return NULL;
 		if (strcmp(systype, RC_SYS_JAIL) == 0)
 			return RC_SYS_JAIL;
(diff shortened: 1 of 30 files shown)


src/librc/librc.c
```diff
@@ -304,6 +306,8 @@ detect_container(const char *systype RC_UNUSED)
 		return RC_SYS_SYSTEMD_NSPAWN;
 	else if (exists("/.dockerenv"))
 		return RC_SYS_DOCKER;
+	else if (exists("/run/.containerenv"))
+		return RC_SYS_PODMAN;
```
floppym, line 310: Please move this above the /.dockerenv check. See systemd for reference.

https://github.com/systemd/systemd/blob/v255/src/basic/virt.c#L640
https://github.com/systemd/systemd/blob/v255/src/basic/virt.c#L764-L767

src/librc/librc.c
```diff
@@ -302,6 +304,8 @@ detect_container(const char *systype RC_UNUSED)
 		return RC_SYS_RKT;
 	else if (file_regex("/proc/1/environ", "container=systemd-nspawn"))
 		return RC_SYS_SYSTEMD_NSPAWN;
+	else if (exists("/run/.containerenv"))
```
navi-desu, line 307: *bsds don't have /run, podman docs suggest that /var/run/.containerenv will be used on freebsd containers: https://docs.podman.io/en/latest/markdown/podman-run.1.html

should we add both here? (and maybe under an ifdef)

navi-desu (changes requested):

src/librc/librc.c
Sylviettee: The check exists in an `#ifdef __linux__`. I could add check for `/var/run/.containerenv` under the pre-existing `#ifdef __FreeBSD__` in the function (Podman only mentions FreeBSD).

navi-desu: oh, yes. thank you

navi-desu:
thank you!

PR #709, opened 2024-05-23
openrc-shutdown: simplify broadcast function

floppym:
Get rid of the alarm/setjmp/longjmp.
Trust that the kernel will not block with O_NDELAY.
Use write() instead of stdio.

Bug: https://bugs.gentoo.org/923326

### DIFF
--- a/src/openrc-shutdown/broadcast.c
+++ b/src/openrc-shutdown/broadcast.c
@@ -17,8 +17,6 @@
 #include <fcntl.h>
 #include <paths.h>
 #include <pwd.h>
-#include <signal.h>
-#include <setjmp.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


src/openrc-shutdown/broadcast.c
```diff
@@ -179,24 +174,23 @@ void broadcast(char *text)
...
-				if (isatty(fd) && (tp = fdopen(fd, "w")) != NULL) {
-					fputs(line, tp);
-					fputs(text, tp);
-					fflush(tp);
-				}
+		alarm(2);
+		flags = O_WRONLY|O_NDELAY|O_NOCTTY;
+		if (file_isatty(term) && (fd = open(term, flags)) >= 0) {
+			if (isatty(fd) && (tp = fdopen(fd, "w")) != NULL) {
+				if ((fputs(line, tp) == EOF || fputs(text, tp) == EOF)
```
N-R-K, line 184: There's a race here:

```c
fputs(..);
// alarm fires
fputs(..); // 2nd fputs, alarm is no longer present
```

But TBH I don't think I understand why it's blocking to begin with. Shouldn't O_NDELAY prevent that??

N-R-K (changes requested):

N-R-K (approved):

thesamesam (approved):

PR #713, opened 2024-05-29
meson: de-duplicate binaries using symlinks

floppym:


floppym:
This is a possible alternative to #712.

I personally prefer the end result given by using hard links, but this symlink approach seems a bit cleaner from a meson perspective.

navi-desu:
the custom target shouldn't be necessary, as the argument for install_symlink is a str, so just

```
foreach exec: einfo_execs
  install_symlink(exec, install_dir: rc_bindir, pointing_to: 'einfo')
endforeach
```

works

floppym:
> the custom target shouldn't be necessary, as the argument for install_symlink is a str, so just

The custom targets are to populate symlinks in the build directory for use by the test suite.

navi-desu:
mind if i tweak this to install `do_{value,service,mark}` instead, and symlink to that? (already done it, just wanted to get confirmation)

floppym:
If you want to make that adjustment, I don't mind.

Just reset the author to yourself because it wasn't my idea and I prefer not to have non-functional binaries tied to my name.

N-R-K:
This would be useful to have. Bump.

PR #714, opened 2024-05-31
shared: fix incompatible function warnings in plugin loader

floppym:
Declare a more specific function pointer type to be punned through a union. Fixes this warning:

```
[73/206] Compiling C object src/openrc/openrc.p/.._shared_plugin.c.o
../src/shared/plugin.c: In function ‘rc_plugin_load’:
../src/shared/plugin.c:94:24: warning: cast between incompatible function types from ‘dlfunc_t’ {aka ‘void (*)(struct __dlfunc_arg)’} to ‘int (*)(RC_HOOK,  const char *)’ [-Wcast-function-type]
   94 |                 fptr = (int (*)(RC_HOOK, const char *))
      |                        ^
```

### DIFF
--- a/src/shared/plugin.c
+++ b/src/shared/plugin.c
@@ -37,28 +37,28 @@
 
 bool rc_in_plugin = false;
 
+typedef int (*rc_hook_t)(RC_HOOK, const char *);
+
 typedef struct plugin
 {
 	char *name;
 	void *handle;
-	int (*hook)(RC_HOOK, const char *);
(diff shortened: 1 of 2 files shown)


navi-desu:
fixed in https://github.com/OpenRC/openrc/commit/a75b9dce780b69e7ca832388610ba7369f67066d, closing

thanks still!

floppym:
Looks good to me!

PR #719, opened 2024-07-16
Generate pkg-config files directly

Kangie:
I recently ported the `openrc-plymouth-plugin` to use the Meson build system. As part of that I made a seemingly sane decision to install to `get_option('libdir'), 'rc' / 'plugins'` based on PFL listings among other things.

Unfortunately this assumption broke on certain Gentoo configurations.

This PR attempts to address this by providing a `rc_path` pkgconfig variable that can be queried so that plugins (etc) don't have to make assumptions. 

To do this we update to using the meson `pkgconfig` module to generate pkgconfig files from whole cloth and drop the old `configure_file` path.

- Remove 'pkgconfig' subdir
- use meson 'pkgconfig' module to generate and install appropriate .pc files when required.
- Raise minimum meson version to 0.56.0 (pkgconfg `variables` as dict)
- add `rc_path` variable

Plymouth Plugin Bug: https://bugs.gentoo.org/935378


Kangie:
```
/tmp/openrc/usr/lib64/pkgconfig/einfo.pc
prefix=/usr
includedir=${prefix}/include
libdir=${prefix}/lib64

Name: einfo
Description: Pretty console informational display
Version: 0.54
Libs: -L/lib64 -leinfo
Cflags: -I${includedir}/einfo
/tmp/openrc/usr/lib64/pkgconfig/OpenRC.pc
prefix=/usr
includedir=${prefix}/include
libdir=${prefix}/lib64

rc_path=/lib/rc

Name: OpenRC
Description: Universal init system
Version: 0.54
Libs: -L/lib64 -lrc
Cflags: -I${includedir}/openrc

PR #721, opened 2024-07-22
fix memory leaks

navi-desu:


src/librc/librc-misc.c
```diff
@@ -131,32 +131,31 @@ rc_proc_getent(const char *ent RC_UNUSED)
...
 	if (!exists("/proc/cmdline"))
 		return NULL;
 
 	if (!(fp = fopen("/proc/cmdline", "r")))
 		return NULL;
 
-	proc = NULL;
 	i = 0;
-	if (rc_getline(&proc, &i, fp) == -1 || proc == NULL)
+	if ((size = getline(&proc, &i, fp) == -1)) {
```
floppym, line 145: This line looks wrong: the first closing paren is in the wrong place. Should be this:
```
	if ((size = getline(&proc, &i, fp)) == -1) {
```

floppym: Addressed in https://github.com/OpenRC/openrc/pull/742.

PR #728, opened 2024-07-26
init.d/hostname: fix /etc/hostname processing

williamh:
This is an alternative to #585 in which we keep the init script but use `hostname -F` to read the value from @SYSCONFDIR@/hostname. This allows for comments in the hostname file.

Also, we deprecate using /etc/conf.d/* to set the host name.

### DIFF
--- a/init.d/hostname.in
+++ b/init.d/hostname.in
@@ -19,18 +19,28 @@ depend()
 
 start()
 {
-	local h source
-	if read -r h _ 2> /dev/null < @SYSCONFDIR@/hostname; then
+	local source
+	if [ -s @SYSCONFDIR@/hostname ]; then
 		source="@SYSCONFDIR@/hostname"
 	elif [ -n "${hostname}" ]; then
-		h=${hostname}
 		source="@SYSCONFDIR@/conf.d/${RC_SVCNAME}"
 	fi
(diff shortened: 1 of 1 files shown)


dwfreed:
This is not portable to non-Linux.  FreeBSD's hostname does not support an `-F` flag.  This is why the initscript currently reads the file itself.

floppym:
It looks like `sys-apps/coreutils[hostname]` does not support the `-F` option either.

navi-desu:
i think the point of #585 was to avoid the overhead of an init script for the simple action of setting the hostname

i think we either move it to c code, or leave the script as is (maybe adding an explicit default for when /etc/hostname is missing, as currently the script noops on that case)

design wise i like the init script more, as we don't put `openrc` itself in charge of setting the hostname, but performance wise, doing it on c code is way less overhead

either way i think deprecating /etc/conf.d/hostname is a good idea, no point to have both it and /etc/hostname, and systemd accepts passing in the hostname via a kernel cmdline flag, maybe we should do smth similar?

williamh:
I'll go back to #585 and take a look at improving it more then. It may not be a bad idea to set the hostname in c, also this might be better since `hostname -F` isn't portable.

PR #729, opened 2024-07-31
set hostname without command if linux

sulincix:
Replace hostname command with echo for linux.


navi-desu:
could you give any explanation to why do this instead of calling `hostname`?

navi-desu:
i'll close this as i see no reason to do this, but feel free to re-open should there be an explanation

PR #740, opened 2024-09-13
make libcap a hard requirement

### LINKED ISSUE #739: make capabilities (libcap) a hard requirement for Linux systems
[This bug](https://bugs.gentoo.org/939543) requests that we make capabilities (libcap) a hard requirement for Gentoo or maybe for all Linux systems.

Are there any objections to OpenRC hard requiring libcap for linux systems?

I'm working on the pr to fix this issue if there are no objections.

Please respond to this issue.

Thanks.

williamh:
For linux systems, libcap is a lightweight dependency. Also, making it optional requires that service scripts be written to deal with the possibility that capabilities are not supported.

This fixes #739.

X-Gemtoo-Bug: https://bugs.gentoo.org/939543

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -46,7 +46,7 @@ endif
 
 if os != 'Linux'
   cap_dep = []
-  kvm_dep = cc.find_library('kvm', required: true)
+  kvm_dep = cc.find_library('kvm')
 else
   cap_dep = dependency('libcap', version: '>=2.33')
   kvm_dep = []
--- a/meson.build
+++ b/meson.build
@@ -45,8 +45,10 @@ else
 endif
 
 if os != 'Linux'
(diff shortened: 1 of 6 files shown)


meson.build
```diff
@@ -45,8 +45,10 @@ else
 endif
 
 if os != 'Linux'
+  cap_dep = []
   kvm_dep = cc.find_library('kvm', required: true)
 else
+  cap_dep = dependency('libcap', version: '>=2.33', required : true)
```
navi-desu, line 51: meson's default for required is already true, we don't need to specify it:
```suggestion
  cap_dep = dependency('libcap', version: '>=2.33')
```

navi-desu (approved):

PR #744, opened 2024-09-17
librc: use return value of rc_getline for len instead of the buffer size

navi-desu:


thesamesam:
This needs a commit message.

thesamesam (approved):

PR #748, opened 2024-09-19
Remove rootprefix

williamh:


meson.build
```diff
@@ -79,23 +79,14 @@ else
...
-  libexecdir = 'lib'
-endif
-libexecdir = rootprefix / libexecdir
-rc_libexecdir = libexecdir / 'rc'
+sbindir = get_option('sbindir')
+rc_libexecdir = get_option('prefix') / libexecdir / 'rc'
 rc_bindir = rc_libexecdir / 'bin'
 rc_sbindir = rc_libexecdir / 'sbin'
 rc_shdir = rc_libexecdir / 'sh'
-sbindir = rootprefix / get_option('sbindir')
```
navi-desu, line 98: sbindir got dropped

meson.build
```diff
@@ -79,23 +79,14 @@ else
...
+bindir = get_option('bindir')
+libdir = get_option('libdir')
 libexecdir = get_option('libexecdir')
-if os == 'Linux' and libexecdir == 'libexec'
-  libexecdir = 'lib'
-endif
-libexecdir = rootprefix / libexecdir
-rc_libexecdir = libexecdir / 'rc'
+sbindir = get_option('sbindir')
+rc_libexecdir = get_option('prefix') / libexecdir / 'rc'
```
navi-desu, line 86: should we consider changing `rc` to `openrc`, or maybe making that an `rc_dir` build option?

src/librc/meson.build
```diff
@@ -1,11 +1,6 @@
 rc_h_conf_data = configuration_data()
-if rootprefix == '/'
-  rc_h_conf_data.set('PREFIX', '')
-  else
-  rc_h_conf_data.set('PREFIX', rootprefix)
-	endif
-rc_h_conf_data.set('LIB', libname)
-rc_h_conf_data.set('LIBEXECDIR', rc_libexecdir)
+rc_h_conf_data.set('RC_LIBEXECDIR', get_option('prefix') / get_option('libexecdir') / 'rc')
+rc_h_conf_data.set('RC_PLUGINDIR', get_option('prefix') / libdir / 'rc/plugins' )
```
navi-desu, line 3: we probably should move plugins from `libdir` to `rc_libexecdir / 'plugins'`

or update the pkg-config to point to `lib` instead, but i'd rather have plugins in libexec

navi-desu (changes requested):

meson.build
williamh: It is defined lower in the file. I'm probably going to remove these variables later in a separate pr.

meson.build
williamh: I'm open to changing this to openrc, probably in another pr, but I don't think we should make this configurable. The FHS says that the directory under libexec should be the program name.

meson.build
navi-desu: oh, i missed it up there, apologies.

meson.build
```diff
@@ -230,7 +221,7 @@ if get_option('pkgconfig')
     libraries : librc,
     subdirs : ['openrc'],
     variables: {
-      'rc_path': rc_libexecdir,
+      'pluginsdir': get_option('prefix') / get_option('libdir') / 'rc/plugins',
```
navi-desu, line 224: we already `libdir = get_option('libdir')` above, we probably should reuse it instead of calling get_option many times

src/librc/meson.build
```diff
@@ -1,11 +1,6 @@
 rc_h_conf_data = configuration_data()
-if rootprefix == '/'
-  rc_h_conf_data.set('PREFIX', '')
-  else
-  rc_h_conf_data.set('PREFIX', rootprefix)
-	endif
-rc_h_conf_data.set('LIB', libname)
-rc_h_conf_data.set('LIBEXECDIR', rc_libexecdir)
+rc_h_conf_data.set('RC_LIBEXECDIR', get_option('prefix') / get_option('libexecdir') / 'rc')
```
navi-desu, line 2: likewise, we already got libexecdir on the main meson.build, we can reuse the variable here

meson.build
navi-desu: fair, let's leave it for a follow up

thesamesam:
As I mentioned on IRC, these need proper commit messages including the full context and should not be squashed either.

navi-desu:
there's still a few things for us to cleanup and improve with meson, but that can be done on a later PR, overall the code LGTM

README.md
```diff
@@ -32,6 +30,17 @@ Obviously, if you're installing this onto a system that does not use
...
 remove them and then install so that the OS hooks into OpenRC.
 
+If you are installing this onto a system which has separate / and /usr
+file systems and expects / to be mounted before /usr, you will need to
+make sure the appropriate pieces are installed on / by passing the
+appropriate options to "meson setup" as follows:
+
+```
+$ meson setup \
+        --bindir /bin --libdir /lib64 --libexecdir /lib --sbindir /sbin \
```
navi-desu, line 40: all of those are basically meson's default, instead i think we should recommend:
```suggestion
        --prefix / --libexecdir lib \
```

navi-desu (approved):

PR #752, opened 2024-09-20
librc, rc, supervise-daemon: Repace hardcoded paths with function calls.

navi-desu:
This lays the groundwork for other features like user services, or a different root prefix, by allowing us to change where openrc looks up for files at runtime.

As a side effect of this change, now all sysconfdirs (aka RC_PKG_PREFIX and RC_LOCAL_PREFIX) are checked for configuration files (rc.conf and rc.conf.d).

This is the first of three patches for openrc-user support.

I am unsure about the api for rc_sysconf_dir, however, an RC_STRINGLIST would be more idiomatic to the codebase, but at the same time, the current NULL terminated double pointer allows for avoiding allocations when not necessary


navi-desu:
removed the side effect of the change, now only the sysconfdir is checked like before.

also renamed the new api functions to be more inline with how we name directory variables.

rc_scriptdirs() holds a null terminated list of strings to the base directories that init.d and conf.d exist (PKG_PREFIX, SYSCONFDIR, LOCAL_PREFIX)
rc_sysconfdir() returns the directory where to load rc.conf (and profile.env) from, SYSCONFDIR
rc_runleveldir(), the runlevels
rc_svcdir(), the service directory where state is kept

PR #753, opened 2024-09-21
support for dynamic multiplexing of services

navi-desu:
generate a new directory on the service dir, "multiplexed", which will
hold all the symlinks we generate at runtime, and a new librc function,
"rc_service_multiplex", which takes a basename or service path, and a
variant name, then generated a symlink to the base service from the
multiplex dir.

this allows multiplexing services at runtime, possibly using data
acquired at runtime, such as the name of an user or interface.

this is the second of three patchsets for user services support

as an initial barebones implementation, it's useful only for generating
services without dynamically multiplexed dependencies, as this is what's
needed for bootstrapping the openrc user script in a new session.

a future patch may improve on it, possibly integrating it in dependency
resolution and on the rc-service tool.

### DIFF
--- a/man/openrc-run.8
+++ b/man/openrc-run.8
@@ -547,6 +547,9 @@ sets the following environment variables for use in the service scripts:
 Name of the service.
 .It Va RC_SERVICE
 Full path to the service.
+.It Va RC_SCRIPTDIR
+Full path to the system directory the service lives in. For multiplexed
+services, it is the directory the main service live in.
 .It Va RC_RUNLEVEL
 Current runlevel that OpenRC is in. Note that, in OpenRC, the reboot
 runlevel is mapped to the shutdown runlevel. This was done because most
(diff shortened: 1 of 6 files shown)


navi-desu:
currently working on how to make sure multiplexed services get added to the deptree properly

probably gonna split the deptree generating function so we can also call it per-service

WhyNotHugo:
Should the documentation also explain what "multiplexed services" are? 
It's not very clear from the current documentation, and this seems to be a new concept.

navi-desu:
> Should the documentation also explain what "multiplexed services" are? It's not very clear from the current documentation, and this seems to be a new concept.

openrc-run(8) kinda just mentions it, but it is not new

it's basically when you have a `.` in the name of the service, so, when starting `net.eth0`, the following configs are attempted to be loaded, in order: `rc.conf, conf.d/net, conf.d/net.eth0`

it is usually done as a symlink, e.g. `wg-quick.vpn1 -> wg-quick`, `net.eth0 -> net.lo`, but doesn't *have* to be, openrc-run only checks the service name to find multiplexed configs

yeah the docs need improving

and doing this dynamically, would be creating that symlink as needed. so say some service does `depend { need foo.nya }`, and /etc/init.d/foo.nya does *not* exit, but /etc/init.d/foo does, openrc would then symlink foo to /run/openrc/multiplexed/foo.nya, and start that

current major use case for this would be instantiating user.<username> automatically for user services

though this feature is currently being debated if we include it at all or not, as i'm not sure it's useful at all outside user services

navi-desu:
superseded by #777 as a basic plumbing/internal impl. not available for public api usage yet, and probably won't be for a little while.

PR #754, opened 2024-09-21
rc-status: Remove unnecessary checks in print_services().

navi-desu:
Every caller of the function already gets a list of services from a valid librc api, and does filtering themselves.

The rc_service_in_runlevel specially is quite clear to me that it doesn't really matter.

The deptree check seems useless, it just checks if it's on the deptree and filters the service if not, but callers already get those service lists from proper functions. If anyone sees the use for the deptree check, do tell


navi-desu:
okay i found one issue with this patch, stacked runlevels are shown when they shouldn't

although i think i can fix it without all that complexity

navi-desu:
ls_dir is erroneusly including directories when LS_INITD is set, which makes rc_services_in_runlevel return stacked runlevels as well. this bug was hidden in rc-status due to the unnecessary double check on every printed service.

PR #765, opened 2024-12-15
consolefont: change parameter order for compatibility with kbd >=2.7

nmeum:
The newly released kbd 2.7 includes a rewrite of the setfont argument parser, with this rewritten parser command-line option need to be passed **before** the font. Otherwise, the parameters will be interpreted as font names. Without this patch, this causes the consolefont service to fail with the following error message:

	setfont: Unable to find file: -C

Note that the `setfont(8)` man page still recommends passing -C after the font name. I am not sure if this is a documentation or implementation issue.

See: https://github.com/legionus/kbd/commit/aed486b760ccc7cdb0ae9a36b428bded6ea132c7

### DIFF
--- a/init.d/consolefont.in
+++ b/init.d/consolefont.in
@@ -52,7 +52,7 @@ start()
 	[ -d /dev/vc ] && ttydev=/dev/vc/
 	x=1
 	while [ $x -le $ttyn ]; do
-		if ! setfont $consolefont $param -C $ttydev$x >/dev/null; then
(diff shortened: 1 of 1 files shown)


nmeum:
> Note that the `setfont(8)` man page still recommends passing -C after the font name.

Created a kbd issue: https://github.com/legionus/kbd/issues/122

nmeum:
This turned out to be a bug in kbd.

PR #767, opened 2025-01-05
implement readiness notification

navi-desu:


src/shared/misc.c
```diff
@@ -521,6 +521,44 @@ pid_t get_pid(const char *applet,const char *pidfile)
...
 }
 
+struct ready ready_parse(const char *applet, const char *ready_string)
+{
+	struct ready ready = {0};
+	if (sscanf(ready_string, "fd:%d", &ready.fd) != 1)
+		eerrorx("%s: invalid ready '%s'.", applet, optarg);
+
+	ready.type = READY_FD;
+	pipe(ready.pipe);
```
N-R-K, line 531: Return value needs checking here, it may fail.

Same with some of the `dup2` calls.

N-R-K (changes requested):
Just a cursory glance. Didn't test it out.

src/start-stop-daemon/start-stop-daemon.c
```diff
@@ -1103,6 +1111,11 @@ int main(int argc, char **argv)
 
 		cloexec_fds_from(3);
 
+		if (ready.type == READY_FD) {
+			close(ready.pipe[0]);
+			dup2(ready.pipe[1], ready.fd);
```
N-R-K, line 1116: This should be checked.

src/supervise-daemon/supervise-daemon.c
```diff
@@ -588,6 +591,9 @@ RC_NORETURN static void child_process(char *exec, char **argv)
 
 	cloexec_fds_from(3);
 
+	if (ready.type == READY_FD)
+		dup2(ready.pipe[1], ready.fd);
```
N-R-K, line 595: Same here.

src/start-stop-daemon/start-stop-daemon.c
navi-desu: those don't really matter at this point. if they fail, the daemon can't write the \n, so the supervisor will already be waiting on `ready` until openrc-run times out and kill it

the best we can do is check and log, if we exit same thing happens, supervisor will start waiting until timeout and s-s-d won't be doing healthchecks

i'd argue for s-s-d it's better to let the exec go through, since then we have a chance of the daemon writing it's pid file so the timeout can kill it cleanly. for s-d that doesn't matter

N-R-K: > if they fail, the daemon can't write the \n

Not necessarily. Let's say the ready fd is supposed to be 3 and the dup2 fails. And in the daemon there's code like this:

```c
somefd = open("somefile", ...); // this gets fd 3 since it's available
// ...
write(3, "ready\n", 6); // this now writes into "somefile"
```

Now you end up with the deamon potentially writing nonsense elsewhere.

navi-desu: good point, i'll add the checks then

PR #769, opened 2025-01-20
Adding back Gentoo Prefix support

littlewu2508:
This reverts 50a338b96a5a6e6774c39dbff961426ccb1b4aa6 and continues https://github.com/OpenRC/openrc/pull/672

@heroxbd and I use openrc inside Gentoo Prefix to manage services, so we'd like to see this feature back

### DIFF
--- a/man/openrc-run.8
+++ b/man/openrc-run.8
@@ -563,6 +563,9 @@ Default runlevel chosen. Default is default.
 .It Va RC_SYS
 A special variable to describe the system more.
 Possible values are OPENVZ, XENU, XEN0, UML and VSERVER.
+.It Va RC_PREFIX
+In a Gentoo Prefix installation, this variable contains the prefix
+offset. Otherwise it is an empty string.
 .It Va RC_UNAME
 The result of `uname -s`.
 .It Va RC_CMD
--- a/meson.build
+++ b/meson.build
@@ -6,6 +6,7 @@ project('OpenRC', 'c',
     'c_std=c99',
     'prefix=/usr',
     'sbindir=/sbin',
+    'gentooprefix=',
     'warning_level=3',
     ],
   meson_version : '>=0.62.0'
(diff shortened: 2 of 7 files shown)


navi-desu:
i'll talk to the folks about PREFIX, seems like it wasn't well supported. but specially now that openrc-user is being merged, we might have a better case to keep it.

williamh:
I reached out to the prefix team multiple times about this in the past, and there didn't seem to be any interest in supporting it. This is the reason it was removed, so I'm against this unless we have more interest from that team.

heroxbd:
Hi William, 

> I reached out to the prefix team multiple times about this in the past, and there didn't seem to be any interest in supporting it. This is the reason it was removed, so I'm against this unless we have more interest from that team.

I hope everything is going on well.

Did you contact me via my email or IRC?  Sorry that I missed your message. I am still catch up with my inbox.  We are interested in supporting OpenRC on Prefix.  

The previous stopper was the Prefix support for meson, which is now fixed.  Please assign future bugs and request review of PRs to the Prefix team (and me) regarding Prefix support of OpenRC when you see appropriate.

Yours,
Benda

navi-desu:
with #942, openrc will take svcdir from the env var RC_SVCDIR, and respect XDG_CONFIG_DIRS for init scripts.

i'll follow it up with another patch for rc.conf, so it'll be possible to add `$prefix/etc/` to XDG_CONFIG_DIRS or RC_PATH, and in there have rc.conf that sets `rc_sys=prefix`

those two changes move the build time checks this PR adds to runtime checks, does gentoo prefix need anything else? i'd rather not have `#ifdef PREFIX` options across the codebase if all we realistically need is flexibility on file lookups

littlewu2508:
>  does gentoo prefix need anything else? 

I think setting the correct RC_PATH and RC_SVCDIR is enough for prefix. Thank you!

navi-desu:
closing this as #942 was merged. please let me know if any other issues pop up with prefix support!

PR #771, opened 2025-01-24
openrc-run.8: Note variables referenced by default 'start'/'stop'.

flexibeast:
This PR updates the openrc-run.8 man page to note which variables are referenced by the default implementations of `start()` and `stop()`.

In creating this PR, i found that the `capabilities`, `no_new_privs` and `secbits` variables, referenced by the default `start()` function, are not documented in openrc-run(8); is this intentional, or can i create a PR to address this?

### DIFF
--- a/man/openrc-run.8
+++ b/man/openrc-run.8
@@ -8,7 +8,7 @@
 .\" This file may not be copied, modified, propagated, or distributed
 .\"    except according to the terms contained in the LICENSE file.
 .\"
-.Dd November 30, 2017
+.Dd January 24, 2025
 .Dt openrc-run 8 SMM
 .Os OpenRC
 .Sh NAME
@@ -79,7 +79,15 @@ Shows which services would be stopped and/or started without actually stopping
 or starting them.
 .El
 .Pp
-The following variables affect the service script:
+The following variables affect the service script. Variables marked with
+.Sq (*)
+are referenced by the default
+.Fn start
+implementation; variables marked with
+.Sq (**)
+are referenced by the default
+.Fn stop
+implementation.
 .Bl -tag -width "RC_DEFAULTLEVEL"
 .It Ar extra_commands
 Space separated list of extra commands the service defines. These should
@@ -120,7 +128,7 @@ List of arguments passed to start-stop-daemon when starting the daemon.
 .It Ar supervise_daemon_args
 List of arguments passed to supervise-daemon when starting the daemon.
 If undefined, start_stop_daemon_args is used as a fallback.
-.It Ar command
+.It Ar command No (**)
 Daemon to start or stop via
 .Nm start-stop-daemon
 or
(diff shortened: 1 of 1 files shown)


N-R-K:
> In creating this PR, i found that the capabilities, no_new_privs and secbits variables, referenced by the default start() function, are not documented in openrc-run(8); is this intentional, or can i create a PR to address this?

I don't think this is intentional. PRs should be welcome.

> This PR updates the openrc-run.8 man page to note which variables are referenced by the default implementations of start() and stop().

I like the idea, but:

- I don't like how it's using `(*)` and `(**)`, they look ugly
- The "default" start and stop can vary based on `supervisor=`.

flexibeast:
> I like the idea, but:
> 
> * I don't like how it's using `(*)` and `(**)`, they look ugly
> * The "default" start and stop can vary based on `supervisor=`.

Okay. Can you suggest an alternative way of conveying this information? (As an aside, [here's the situation which led me to create this PR](https://bugs.gentoo.org/948143#c3).)

N-R-K:
I think we can have a separate section in a table format. Something like this:

```
.Sh DEFAULT START STOP
The following table lists all the variables that are used by the default start
and stop functions.
.Bl -column "Variable" "start\-stop\-daemon" "supervise\-daemon" "s6" -offset indent
.It Sy "Variable" Ta Sy "start\-stop\-daemon" Ta Sy "supervise\-daemon" Ta Sy "s6"
.It "pidfile"     Ta "start,stop" Ta "start,stop" Ta ""
.El
```

Screenshot:

image

flexibeast:
Sorry i've not replied sooner than this; my health issues have become particularly severe other the last year or so ....

Can i use tbl(1) to create this table, or do you want to avoid using preprocessors?

N-R-K:
Hey, no need to worry. Health always comes first.

> Can i use tbl(1) to create this table, or do you want to avoid using preprocessors?

Should be fine as long as it's rendered properly with `man`, do also note that we care about non-gnu systems like bsds as well.

flexibeast:
> Should be fine as long as it's rendered properly with `man`, do also note that we care about non-gnu systems like bsds as well.

Yeah, tbl is a standard preprocessor for systems that use man-db+groff for rendering man pages (cf. groff(1)'s `-t` option), and on systems that use mandoc(1) by default (cf. [this table from a tutorial of mine](https://github.com/flexibeast/guides/blob/master/mdoc.md#appendix-b-os-mdoc-support)), mandoc(1) implements a subset of the tbl language (cf. [tbl(7)](https://man.openbsd.org/tbl.7)). It's easy for me to check the rendering in both cases.

navi-desu:
ftr eventually i'd like to sunset start-stop-daemon, so it'd end up being supervise-daemon and s6, and additionally skarnet wants s6 to match as much of supervise-daemon's functionalities as possible

we could end up with just a table of `variable` `function` and optionally `notes`. start, stop, and status, use different subset of variables

navi-desu:
closing as #973 was merged

PR #782, opened 2025-03-04
Support exporting variables into an starting environment

navi-desu:


src/librc/librc-misc.c
```diff
@@ -443,3 +444,18 @@ rc_conf_value(const char *setting)
...
+void rc_export_variable(const char *name, const char *value) {
+	char *envpath;
+	FILE *envfile;
+
+	xasprintf(&envpath, "%s/env", rc_svcdir());
+
+	if (!(envfile = fopen(envpath, "a+")))
+		return;
+	if (flock(fileno(envfile), LOCK_EX) == -1)
+		return;
```
N-R-K, line 457: This leaks the FILE handle.

Should log the failure too, I think.

src/librc/librc-misc.c
```diff
@@ -443,3 +444,18 @@ rc_conf_value(const char *setting)
...
+	/* TODO: Remove repeated entries, before inserting a new one. */
+	fprintf(envfile, "export %s=%s\n", name, value);
+	fclose(envfile);
```
N-R-K, line 460: Should not assume that the writes succeeded.
```suggestion
        int err = fflush(envfile);
	if (fclose(envfile) != 0 || err)
	    // log
```

src/rc-update/rc-update.c
```diff
@@ -35,6 +35,7 @@ const char *extraopts = NULL;
 const char *usagestring = ""							\
 	"Usage: rc-update [options] add <service> [<runlevel>...]\n"	\
 	"   or: rc-update [options] del <service> [<runlevel>...]\n"	\
+	"   or: rc-update [options] export <variables>...\n"	\
```
N-R-K, line 38: Is this useful? Services need to be manually restarted by the user for it to take effect, right?

N-R-K (changes requested):

src/rc-update/rc-update.c
```diff
@@ -238,116 +240,96 @@ int main(int argc, char **argv)
...
-			action = DODELETE;
+			action = DO_DELETE;
 		else if (strcmp(argv[optind], "show") == 0)
-			action = DOSHOW;
-		if (action)
-			optind++;
+			action = DO_SHOW;
 		else
-			eerrorx("%s: invalid command `%s'",
-			    applet, argv[optind]);
```
N-R-K, line 262: Showing usage string is fine, but I think this message shouldn't be removed. I *do* want to explicitly see what I did wrong, not just a blanket usage string.

src/rc-update/rc-update.c
```diff
@@ -238,116 +240,96 @@ int main(int argc, char **argv)
...
+		optind++;
 	}
-	if (!action)
-		action = DOSHOW;
+	if (action == DO_NONE)
+		action = DO_SHOW;
 
-	runlevels = rc_stringlist_new();
+	if (optind >= argc && action != DO_SHOW)
+		usage(EXIT_FAILURE);
```
N-R-K, line 270: Same here, it should still show you what you did wrong. Not just print usage with no other context.

src/librc/librc-misc.c
```diff
@@ -443,3 +444,18 @@ rc_conf_value(const char *setting)
 
 	return rc_config_value(rc_conf, setting);
 }
+
+void rc_export_variable(const char *name, const char *value) {
+	char *envpath;
+	FILE *envfile;
+
+	xasprintf(&envpath, "%s/env", rc_svcdir());
```
N-R-K, line 452: This is leaking too, isn't it?

N-R-K (changes requested):

src/rc-update/rc-update.c
navi-desu: for running services yes you have to restart them, this is for the activation environment, stuff like, exporting WAYLAND_DISPLAY for gui daemons (notification, xdg-desktop), or a dbus script that exports the bus address on start_post so that services depending on it can find the bus

the former kindof works if you `rc_env_allow` it, but it's inconsistent. the latter doesn't work at all

i'm also thinking of having `export="FOO NYA=MEW"` as a part of the script themselves, unsure if that's worth it over services calling `rc-update export` manually, and lastly per-runlevel envs could also be nice.

src/librc/librc-misc.c
navi-desu: should we be logging from librc? most functions usually don't so ideally we just indicate failure (flock and fopen set errno on failure for us too)

navi-desu: changed it to return failure, the caller would then log should it want to

src/rc-update/rc-update.c
```diff
@@ -238,116 +240,100 @@ int main(int argc, char **argv)
...
 		case_RC_COMMON_GETOPT
 		}
+	}
 
-	verbose = rc_yesno(getenv ("EINFO_VERBOSE"));
-
-	if ((action & DOSHOW   && action != DOSHOW) ||
-	    (action & DOADD    && action != DOADD) ||
-	    (action & DODELETE && action != DODELETE))
-		eerrorx("%s: cannot mix commands", applet);
```
N-R-K, line 247: Was this just dead code?

navi-desu: yeah, nothing sets `action` before that check, so it's always zero, and even the code that *does* set it later, do assign it fully, not just bits

so i have no idea what was intended here but that error just never happens

if i were to guess, maybe action used to be set by getopt flags?

src/rc-update/rc-update.c
thesamesam: Maybe include this in the commit message too?

navi-desu:
i rebased this on top of #819, so that's now a dependency of this mr

also reimplemented it with envdirs instead, makes it easier to unset variables, and we don't have to worry about escaping things at all

navi-desu:
added more details in commit messages

man/rc-update.8
```diff
@@ -59,6 +63,10 @@ Delete the
 from the
 .Ar runlevel
 or the current one if none given.
+.It Ar export Ar variable ...
+Exports listed
+.Ar variables
+to the starting environment of all services.
```
N-R-K, line 69: Maybe something like:

> [...] of all services (re)started afterwards.

To make it clear that exporting variable won't affect already running services?

src/value/value.c
```diff
@@ -36,23 +36,29 @@ int main(int argc, char **argv)
...
+		if (!(option = rc_service_value_get(service, argv[1])))
+			return false;
+		printf("%s", option);
+		free(option);
+		return true;
+	case SET:
+		return rc_service_value_set(service, argv[1], argv[2]);
+	case EXPORT:
+		option = strsep(&argv[1], "=");
+		return rc_export_variable(service, option, argv[1] ? argv[1] : getenv(option));
```
N-R-K, line 62: This will return `true` (i.e 1) on success. Intended?

src/librc/librc-misc.c
```diff
@@ -447,3 +448,27 @@ rc_conf_value(const char *setting)
 
 	return rc_config_value(rc_conf, setting);
 }
+
+bool rc_export_variable(const char *service, const char *name, const char *value) {
+	int dirfd = openat(rc_dirfd(RC_DIR_ENVIRONMENT), service, O_RDONLY | O_DIRECTORY);
```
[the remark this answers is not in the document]
navi-desu: `RC_CONF_OLD` is `/etc/conf.d/rc` so at the very least they'd get deprecation warnings (i have no idea why we changed the config from there to `/etc/rc.conf`, and tbh, i liked it better there but oh well)

so it's kindof implicit that `rc` is a reserved service name. we could add an explicit check though

man/rc-update.8
navi-desu: i thought that `starting environment` covered that, though maybe it's not clear enough?

N-R-K: Hmm, actually I think that's fine.

src/librc/librc-misc.c
```diff
@@ -447,3 +450,44 @@ rc_conf_value(const char *setting)
...
+	}
+
+	if (!value) {
+		unlinkat(dirfd, name, 0);
+		return true;
+	}
+
+	envfile = do_fopenat(dirfd, name, O_WRONLY | O_TRUNC | O_CREAT);
+	fputs(value, envfile);
+	return fclose(envfile) == 0;
```
N-R-K, line 473: You also need to check the `fputs` return value. Also, shouldn't we check of `envfile` being `NULL` as well?

src/librc/librc-misc.c
```diff
@@ -447,3 +450,44 @@ rc_conf_value(const char *setting)
...
 }
+
+bool rc_export_variable(const char *service, const char *name, const char *value) {
+	int dirfd = openat(rc_dirfd(RC_DIR_ENVIRONMENT), service, O_RDONLY | O_DIRECTORY);
+	FILE *envfile;
+
+	if (dirfd == -1)
+		return false;
+
+	if (flock(dirfd, LOCK_EX) == -1) {
```
N-R-K, line 461: Uhh, when is the unlock happening?

N-R-K (changes requested):

src/librc/librc-misc.c
navi-desu: oh... it was when fclose closed the file fd... before we changed it to be an envdir, blep

this should flock the envfile inside the dir anyway, not the dir itself

src/librc/librc-misc.c
```diff
@@ -447,3 +450,51 @@ rc_conf_value(const char *setting)
...
+
+	fputs(value, envfile);
+	return fclose(envfile) == 0;
+}
+
+void rc_import_variables(const char *service) {
+	DIR *envdir = do_opendirat(rc_dirfd(RC_DIR_ENVIRONMENT), service ? service : "rc");
+	struct dirent *d;
+	if (!envdir)
+		return;
```
N-R-K, line 487: Should return a `bool` to differentiate between a `ENOENT` failure vs some actual other failure to open.

src/librc/librc-misc.c
```diff
@@ -447,3 +450,51 @@ rc_conf_value(const char *setting)
...
+
+	while ((d = readdir(envdir))) {
+		char *value;
+		size_t size;
+
+		if (!do_getfileat(rc_dirfd(RC_DIR_ENVIRONMENT), d->d_name, &value, &size))
+			continue;
```
N-R-K, line 494: Same here. Silently ignoring actual io failure can make it hard to debug when it actually happens.

src/openrc-run/openrc-run.c
```diff
@@ -943,8 +945,10 @@ svc_stop_deps(RC_SERVICE state)
 		return;
 
 	TAILQ_FOREACH(svc, tmplist, entries) {
-		if (rc_service_state(svc->value) & RC_SERVICE_STOPPED)
+		rc_import_variables(svc->value);
+		if (rc_service_state(svc->value) & RC_SERVICE_STOPPED) {
```
N-R-K, line 949: So service specific env are supposed to overwrite the global one?

Also should we be importing the env of a stopped service?

N-R-K (changes requested):

N-R-K:
Also, wouldn't we be able to get rid of this and simply export them into the starting env during init?

https://github.com/OpenRC/openrc/blob/6411b4b0cce1781da20ec0a6c0babc4bd30d1f3f/src/shared/misc.c#L67-L71

Seems to be a cleaner and more deterministic approach.

src/openrc-run/openrc-run.c
navi-desu: > Also should we be importing the env of a stopped service?

no i'm just stupid and forgot the dep stop is about dependees, not dependers

> So service specific env are supposed to overwrite the global one?

i dunno actually, i could see it being argued either way

N-R-K: > i dunno actually, i could see it being argued either way

Yes, that's why I am asking. Either way we need to document it. Or we could also provide some sort of special "const" functionality for the global env which disallows services overriding it. E.g something like: `rc.const` which will be sourced at the very end.

navi-desu:
> Also, wouldn't we be able to get rid of this and simply export them into the global starting env during init?

for XDG_RUNTIME_DIR, no, since the envdir would be located in $XDG_RUNTIME_DIR/openrc/environment -- for the others we could tho it'd be a lot of "setting the same values at every invocation" and the tmpfs IO related to that, just to re-export the same vars again later?

src/openrc-run/openrc-run.c
navi-desu: okay, i say service-specific ones should take priority, because:

1: allows you to set something generically for all services, then override it in specific conf.d's
2: services could "choose" to allow users to override vars by doing `: ${FOO:=default}`
3: matches the conf.d semantics (i.e. rc.conf goes first, conf.d/$svc overrides)

N-R-K:
It'd be nice to have some user facing documentation on how this works. Maybe in `service-script-guide.md`.

amano-kenji:
If you are trying to change openrc-run, the man page for openrc-run should change as well, or people will not know about this.

navi-desu:
it's not ready yet, i'm still changing things

docs will be updated before it's merged

amano-kenji:
Does this pull request export variables into a starting environment and also new login sessions? There can be multiple login sessions.

If I log in sway and then a linux virtual console, I have two login sessions. pam module should export DBUS_SESSIONS_BUS_ADDRESS to the second login session and the later ones as well as the first session that started dbus user service.

navi-desu:
new version.

i'm unsure about the librc api, but seems okay enough?

i've also for now added a new binary, rc-env, i still prefer `rc-update env`, and will try again to do that instead -- it's just a pain because of how the opt parsing and help text is structured

pam_openrc now imports the variables of services added to 'boot' to the pam environment

src/librc/librc-misc.c
```diff
@@ -55,35 +55,25 @@ rc_yesno(const char *value)
...
+	fp = do_fopenat(dirfd, file, O_RDONLY);
 	if (!fp)
 		return false;
 
 	/* assume fileno() never fails */
-	fd = fileno(fp);
-
-	if (fstat(fd, &st))
+	if (fstat(fileno(fp), &st))
 		goto finished;
```
swagtoy, line 72: Make sure `*buffer` is set to NULL at the start of this function. If `fstat` fails, it jumps to `finished` and attempts to free `*buffer`, but we do not know what `*buffer` is set to at this specific point. It could be uninitialized.

In the uses later of this it is initialized to NULL, but i do think it's good practice to ensure this is unset.

src/librc/librc.c
```diff
@@ -1268,3 +1269,73 @@ rc_services_scheduled(const char *service)
...
+	if (!service)
+		service = "rc";
+
+	if (mkdirat(rc_dirfd(RC_DIR_ENVIRONMENT), service, 0644) == -1 && errno != EEXIST)
+		return false;
+	if ((dirfd = openat(rc_dirfd(RC_DIR_ENVIRONMENT), service, O_RDONLY | O_DIRECTORY)) == -1)
+		return false;
+
+	if (!value) {
+		unlinkat(dirfd, name, 0);
```
swagtoy, line 1288: Should dirfd still be closed here? It's also possible for this to fail.

src/librc/librc.c
```diff
@@ -1268,3 +1269,73 @@ rc_services_scheduled(const char *service)
...
+		char *value = NULL;
+		size_t size;
+
+		if (d->d_name[0] == '.')
+			continue;
+
+		if (!do_getfileat(dirfd(envdir), d->d_name, &value, &size))
+			continue;
+
+		xasprintf(&envlist[i++], "%s=%s", d->d_name, value);
```
swagtoy, line 1334: I'm still in favor of having a second function to retrieve these values, since it makes it easier to pretty-print or use in, say, an application that has 2 textboxes for each key/value. (otherwise you have to rely on strchr and all that). Is this still possible? :-)

src/openrc-run/openrc-run.c
```diff
@@ -599,6 +599,21 @@ svc_start_check(void)
...
+svc_putenv(const char *svcname)
+{
+	char **env;
+	size_t count;
+
+	if ((count = rc_service_environ(svcname, &env)) == 0)
+		return;
+
+	for (size_t i = 0; i < count; i++)
+		putenv(env[i]);
```
swagtoy, line 612: should check putenv for errors and propagate them?

swagtoy (changes requested):
Looks good! Just some comments :-)

src/pam_openrc/pam_openrc.c
```diff
@@ -131,10 +132,35 @@ exec_openrc(pam_handle_t *pamh, bool opening, bool quiet)
...
 	(void) argc;
 	(void) argv;
 
-	return exec_openrc(pamh, true, flags & PAM_SILENT);
+	if ((err = exec_openrc(pamh, true, flags & PAM_SILENT)) != PAM_SUCCESS)
+		return err;
+
+	rc_set_user();
+
+	services = rc_services_in_runlevel("boot");
```
swagtoy, line 147: These must be `rc_stringlist_free`'d, no?

swagtoy (changes requested):

src/librc/librc-misc.c
navi-desu: the api of getfile is the same as getline, which means a buffer can be re-used (thus why re `xrealloc`)

it's the caller's responsibility to pass in either *buffer = NULL or re-use *buffer from a past call

src/openrc-run/openrc-run.c
navi-desu: we usually abort on allocation failure, which i really dislike and would like to handle errors properly

here i'll add a eerror but otherwise attempt to proceed with the service

> Also, ENOMEM can also be a sign of something wrong with malloc internal structures indicating a serious bug or worse corruption (but glibc malloc will usually yell regardless). They are always nice to check.

that is not our problem, we assume a working malloc. if malloc is broken, we are to handle it the same as out of memory

navi-desu: how would *we* have broken malloc? what scenario would that come up, and what "help" would we even give other than "error: ENOMEM"?

if you mean we go out of bounds and corrupt some malloc data, that's not on the field of "helping highlight broken malloc", that is "we did out of bound writes and that's *bad*™"

swagtoy: I mean the latter. Course it's bad, but the earlier it's caught the easier it can be smelt out. I agree through ENOMEM is not helpful for that in particular 

For now you can leave it alone

thesamesam: If internal malloc data structures are corrupted, the implementation is free to abort or not return an error. We hope that useful implementations would abort.

I think it's unlikely that they'll return ENOMEM in such a case: it seems unlikely that an implementation could reliably detect corruption but wouldn't then abort on it.

swagtoy: It's hard to detect corruption. Libcurl just returns a memory error. But this is an application, so abort is fine, Sam put it best 

Malloc will abort anyway if something is effed up (i recall)

WhyNotHugo:
Isn't it much simpler to export variables into an envdir from one init script and have dependant service import that envdir from their own script?

Such an approach doesn't require adding complexity to OpenRC itself.

WhyNotHugo:
Perhaps I'm missing something, but if service A sets a variable for service B, reading the definition for service B provides no indication that it's receiving this environment variable from some external source.

navi-desu:
> Isn't it much simpler to export variables into an envdir from one init script and have dependant service import that envdir from their own script?
> Such an approach doesn't require adding complexity to OpenRC itself.

not really for a few reasons:

1) the goal is to have it be transparent, if a variable is *required* to be set, like how DBUS_SESSION_BUS_ADDRESS and SSH_AUTH_SOCK are, then doing `depend() { need dbus ssh-agent; }` should be enough, the depend function encodes information about external factors which the service requires to work and i'd like to not split that

as well as, the service that export the variable should ideally be tied to the exported variables, and having the logic in openrc makes state cleanup easier (keeping exported variables in ram after the service is stopped would be bad, since `$svc restart` is expected to have a clean state

s6/runit services might encode the variables in an envdir, but openrc has the script as the source of truth, we'd use envdir as an implementation detail

2) pam_openrc must have access to said environment variables, variables exported by `boot` services will be expected to be exported in the pam session -- otherwise manual setup would be required for anyone using dbus via openrc, and that's not too great

this requirement means each service needs a standardized way of exporting said variables

3) currently the implementation is done via an envdir yes, but i'm changing that to an env file, reasons being: we don't need the flexibility for modification that an envdir has, each service starts with exported envs being an empty set, and adds to them as they go, and have the set cleared at the end -- so having a single file would reduce the number of syscalls and allocations necessary, while making processing easier

> Perhaps I'm missing something, but if service A sets a variable for service B, reading the definition for service B provides no indication that it's receiving this environment variable from some external source.

`A` doesn't set variables "for `B`", it exports variables "required" for any service to access `A`, so it's more like "service `B`, by doing `need A`, it asks for all needed variables from `A`

WhyNotHugo:
I missed the fact that the "depends" declaration triggers importing the variable, thanks for the clarification.

xz-dev:
What happens when a user has multiple concurrent PAM sessions running different desktop environments (e.g., KDE on tty7 and GNOME on tty6)?

Will the exported environment variables from one session overwrite those from the other?

navi-desu:
> What happens when a user has multiple concurrent PAM sessions running different desktop environments (e.g., KDE on tty7 and GNOME on tty6)?

currently all pam sessions share the service directory, so they all share the same state

note that this is also the case on systemd, dinit, and pretty much all init systems

openrc does (in master) support setting different RC_SVCDIR, but running multiple parallel instances is not something tested nor that we guarantee works

it is something i want to think of properly doing eventually, but that is far off

so for now, if you're running multiple DEs at once, for now it is expected that they can share the services, e.g., pipewire, dbus, etc, (since even ignoring the environment sharing, you wouldn't be able to start two different "dbus" or "pipewire" services anyway)

swagtoy:
Just for the record, starting multiple sessions like that would go
haywire anyway. Especially more so if they were the _same_
environment, in which case you'd need something like userdb (which is
some logind nonse), and afaik KDE has nothing for that right now,
gnome only much more recently. This may also just lead to some dbus
conflicts too on some components, which i know GDM had some hacks to
work around in the past, not sure about KDE.


Also; a lot of those environments will usually try not to step on
their toes with environment variables anyway; i.e. one such case is
that I needed to export DISPLAY or WAYLAND_DISPLAY (well, this works
automagically for the most part, but for now...), in which case, these
systems will not attempt to allocate over themselves, but at the
moment of course since i have hardcoded these env variables it's
impossible to do what you ask without this branch. After all, they
needed to work around the conflicts on systemd too :) So at the very
minimum, you might have dbus instances step on each other if the
display managers don't start their own dbus, in which case, there be
dragons anyway. Do at your own risk.

So ideally, running multiple DE's at once is not recommend. Just keep
this all in mind if you plan on actually doing that.

swagtoy:
Other concerns pop up too; like what if you close the lid of a laptop,
press the power button, etc? This kind of thing buries a bit beyond
env variables and into logind and is thus; risky behavior in the first
place.

swagtoy:
heads up: In the coming week I will test this PR out and experiment with its usage as a way to propagate env variables with gnome-session-openrc. I feel like I noticed a possible design quirk (wrt local service environments) but I won't comment on it much until I really prove it's there; it has to do with how things work when services exit (and when the user logs out/logs in). i could've missed a line of code here which shows how that's handled, though.

Code overall seems fine to me.

navi-desu:
okay i'm tired of pushing this off and i'm happy with this version

any changes, we can do before 0.64

swagtoy:
that's fine. thanks! I'll play with it some

PR #785, opened 2025-03-07
Allow only building the pam_openrc module for user services

nmeum:
This is useful for Alpine's OpenRC package where we build OpenRC without PAM support (since PAM is not supported by the Alpine base system, i.e. BusyBox) but want to provide the pam_openrc module as a standalone subpackage for people wanting to use OpenRC user services with the autostarting feature.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -64,7 +64,8 @@ else
   cc_pam_flags = []
 endif
 
-if not pam_dep.found() and get_option('pam')
+pam_required = get_option('pam') or get_option('pam_user')
+if not pam_dep.found() and pam_required
   error('Pam was requested but could not be located')
 endif
 
@@ -98,12 +99,12 @@ selinux_dep = dependency('libselinux', required : get_option('selinux'))
 pam_misc_dep = []
 if selinux_dep.found()
   cc_selinux_flags = '-DHAVE_SELINUX'
-  if pam_dep.found() and get_option('pam')
+  if pam_dep.found() and pam_required
     pam_misc_dep = dependency('pam_misc', required: false)
     if not pam_misc_dep.found()
       pam_misc_dep = cc.find_library('pam_misc', required: false)
     endif
-    if not pam_misc_dep.found() and get_option('pam')
+    if not pam_misc_dep.found() and pam_required
       error('Pam was requested but could not be located')
     endif
   else
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -14,6 +14,8 @@ option('os', type : 'combo',
   description : 'Operating System (autodetected if not specified)')
 option('pam', type : 'boolean',
   description : 'enable PAM support')
+option('pam_user', type : 'boolean',
+  description : 'only enable PAM support for the pam_openrc user services module')
 option('pkg_prefix', type : 'string',
   description : 'default location where packages are installed')
 option('pkgconfig', type : 'boolean',
--- a/src/pam_openrc/meson.build
+++ b/src/pam_openrc/meson.build
@@ -1,4 +1,4 @@
-if get_option('pam') and pam_dep.found()
+if (get_option('pam') or get_option('pam_user')) and pam_dep.found()
   shared_module('pam_openrc',
     ['pam_openrc.c', misc_c, version_h],
     c_args : [cc_branding_flags],


navi-desu:
this brings up an issue with regards to the `openrc-user` daemon.

pam-based setups usually have modules (such as `elogind`, `pam_xdg`, etc) provide resources for the session, ranging from setting up `XDG_RUNTIME_DIR`, which openrc needs, to in some cases mounting the home folder.

since `src/openrc-user` starts a session for said user, it'd also need pam support enabled to work correctly on the majority of systems that rely on such pam modules

in another hand, if we enable pam for `src/openrc-user` and splitting it into a sub package, would mean non-pam users won't be able to make use of `openrc-user` or the `etc/init.d/user` script

is that a compromise (limiting *all* forms of auto-starting to pam-only users, be it boot or login), or is there any suggestions for this mix-match of a problem?

sertonix:
> since src/openrc-user starts a session for said user, it'd also need pam support enabled to work correctly on the majority of systems that rely on such pam modules

When openrc-user is started through pam the other modules (and a session) would already be loaded. So is it possible to do pam -> pam_openrc -> openrc-user or does it have to be openrc-user -> pam -> pam_openrc?

Or does it work differently?

navi-desu:
> When openrc-user is started through pam the other modules (and a session) would already be loaded. So is it possible to do pam -> pam_openrc -> openrc-user or does it have to be openrc-user -> pam -> pam_openrc?

nop this doesn't work for a few reasons:
* for starting at boot, we won't be started via a pam module, at all
* for starting at login, pam does *not* run close session in reverse order, so:
```
session pam_elogind.so
session pam_openrc.so
```
open session > elogind sets up XDG_RUNTIME_DIR > openrc uses it to store service data
close session > elogind clears XDG_RUNTIME_DIR > openrc no longer has any data about services

and doing it the other way around means openrc won't have XDG_RUNTIME_DIR set up when it runs

relying on that is really inconsistent and edge-case prone too, since it'd rely on close_session to actually be called for clean-up

this all is reason openrc-user was made into a daemon instead of simple pam hooks

(to note, currently we still do rely on it to be called, but i'm going to change it so instead we get an fd from openrc-user, and leak it to the session, thus in case of a crash we still clean up properly)

sertonix:
(Not sure how openrc-user is started correctly then but) compiling multiple versions of openrc-user (with and without pam) shouldn't cause any incompatibilities, right?

navi-desu:
> compiling multiple versions of openrc-user (with and without pam) shouldn't cause any incompatibilities, right?

no incompatibility, it's a daemon and basically standalone, the one without pam will simply not call `pam_{open,close}_session`, and that works as long as the user sets XDG_RUNTIME_DIR some other method, like on their shell's *rc file

they can't be both installed at the same time though, without at least some heavy and ugly patching of both the meson files and init.d/user

sertonix:
Otherwise is it reasonable to out the pam stuff in (yet another) split binary?

> they can't be both installed at the same time though, without at least some heavy and ugly patching of both the meson files and init.d/user

This could be handled in the alpine package.

navi-desu:
how doable would it be to, build openrc without pam, split openrc-user from it to a sub package "openrc-user-nopam", then reconfigure it with pam, rebuild only openrc-user and pam_openrc, and split that into "openrc-user", then having those two packages conflict with each other?

> Otherwise is it reasonable to out the pam stuff in (yet another) split binary?

technically but then we'd have one daemon doing nothing, holding another daemon doing nothing

sertonix:
> how doable would it be to, build openrc without pam, split openrc-user from it to a sub package "openrc-user-nopam", then reconfigure it with pam, rebuild only openrc-user and pam_openrc, and split that into "openrc-user", then having those two packages conflict with each other?

Shouldn't be an issue. We do similar stuff with polkit. (The debug symbols will be broken but that is a bug in abuild)

sertonix:
I wrote a PoC here: https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/81007#note_487993

We do need a guarantee that librc and libeinfo are ABI compatible regardless of the `pam` configuration option for that to work.

navi-desu:
> We do need a guarantee that librc and libeinfo are ABI compatible regardless of the pam configuration option something like this to work.

librc does not link against pam in any configuration, nor change symbols based on -DHAVE_PAM

the binaries making use of HAVE_PAM directly are start-stop-daemon, supervise-daemon, openrc-user

indirectly via shared/selinux.c, we have openrc-run, checkpath, start-stop-daemon

and none of those change api/abi level stuff (that i know of) aka function signatures and struct definitions are constant, pam is only used in implementation details, and afaik, we have no intention of changing that in the foreseeable future

nmeum:
Sorry, didn't notice that openrc-user also has a `HAVE_PAM` ifdef. Otherwise, I wouldn't have opened this in the first place. I am sure we can handle this on the packaging side. Thanks for your input though :)

PR #786, opened 2025-03-08
init.d/*: Quote the argument of ":" commands

ulm:
This avoids globbing, see: https://www.shellcheck.net/wiki/SC2223


navi-desu:
considering all of those values are set in admin-written configs, and outside vars are filtered, so i'm not quite sure this matters at all, but it also doesn't hurt so let's merge

eventually i'd like to run shellcheck in tests/ci as well too

PR #788, opened 2025-03-09
shared, *-daemon: support systemd style READY=1

navi-desu:


src/shared/misc.c
```diff
@@ -576,8 +602,24 @@ bool ready_wait(const char *applet, struct ready ready)
...
+		switch (ready.type) {
+		case READY_NONE:
+			break;
+		case READY_FD:
+			if (memchr(buf, '\n', bytes))
+				return true;
 			break;
+		case READY_SOCKET:
+			buf[bytes] = '\0';
+			if (strstr(buf, "READY=1")) {
```
N-R-K, line 614: `memmem`? should be available on all targets we care about.

src/shared/misc.c
```diff
@@ -546,24 +548,48 @@ pid_t get_pid(const char *applet,const char *pidfile)
...
+			struct sockaddr_un unix;
+		} addr = { .unix = { .sun_family = AF_UNIX } };
+		int written;
+
+		ready.type = READY_SOCKET;
+		written = snprintf(addr.unix.sun_path, sizeof(addr.unix.sun_path), "%s/supervise-%s.sock", rc_svcdir(), applet);
+		if (written > (int) sizeof(addr.unix.sun_path))
+			eerrorx("%s: socket path '%s/supervise-%s.sock' too long.", applet, rc_svcdir(), applet);
+		else if (written < 0)
+			eerrorx("%s: snprintf failed.", applet);
```
N-R-K, line 567: I don't think snprintf can ever fail, unless you give it invalid format specifier or length argument that's bigger than `INT_MAX`. Neither of which happens here.

src/shared/misc.c
navi-desu: `memmem` works but it'd say the usage of it here would be odd-looking

considering i want to support a few other systemd-notify messages later (namely, STATUS=), it doesn't really matter, so for now, let's just leave it at `strstr` i think

N-R-K: Yeah, it would've also required some meson probe since it's quite new in POSIX, even though practically available everywhere. Not worth it just for this.

PR #789, opened 2025-03-11
shared/misc: fix types on close_range wrapper

dwfreed:
The syscall expects unsigned integers.  Notably, if this wrapper is called, `UINT_MAX` is passed as `last`, and conversion from unsigned to signed is implementation-defined.


N-R-K (approved):

navi-desu:
thanks

PR #790, opened 2025-03-13
librc/run/shutdown: precalc strlen moved out of bounds loop

GermanAizek:
@navi-desu,

Example with clang-19 -O2

Before:

before

After:

after

### DIFF
--- a/src/librc/librc-daemon.c
+++ b/src/librc/librc-daemon.c
@@ -158,6 +158,7 @@ rc_find_pids(const char *exec, const char *const *argv, uid_t uid, pid_t pid)
 			my_ns[0] = '\0';
 	}
 
+	size_t len_my_ns = strlen(my_ns);
 	while ((entry = readdir(procdir)) != NULL) {
 		if (sscanf(entry->d_name, "%d", &p) != 1)
 			continue;
@@ -172,7 +173,7 @@ rc_find_pids(const char *exec, const char *const *argv, uid_t uid, pid_t pid)
 				proc_ns[0] = '\0';
 		}
 		free(buffer);
-		if (pid == 0 && strlen(my_ns) && strlen (proc_ns) && strcmp(my_ns, proc_ns))
+		if (pid == 0 && len_my_ns && strlen (proc_ns) && strcmp(my_ns, proc_ns))
 			continue;
 		if (uid) {
 			xasprintf(&buffer, "/proc/%d", p);
--- a/src/openrc-run/openrc-run.c
+++ b/src/openrc-run/openrc-run.c
@@ -308,6 +308,9 @@ write_prefix(const char *buffer, size_t bytes, bool *prefixed)
 	else
 		ewarnv("Couldn't open the prefix lock, please make sure you have enough permissions");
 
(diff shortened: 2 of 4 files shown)


thesamesam:
Note that there's at least one sight-impaired contributor to this repository who will be looking at this PR. Please include godbolt links as well as screenshots to make it accessible.

N-R-K:
Technically, this should be fine; assuming that those strings don't change inside the loop (didn't verify myself).

But my gut feeling is that it falls into intangible micro-optimization. If we wanted to be serious about string performance, we'd rid of nul-strings entirely. (Not that I advocate doing it, as it'd be a pretty big and error prone rewrite for little gain).

navi-desu:
the assembly shows, the call to strlen happens before label that is used for looping

merging this is still good imo because it's clearer, and doesn't rely on compiler optimization to make the code not be awful

N-R-K:
> because it's clearer

Is it? It de-syncs the string pointer and the string length. And so if someone in the future edits it so that the pointer changes during the loop but then forgets to change the length then bad stuff happens.

src/openrc-run/openrc-run.c
```diff
@@ -320,9 +323,9 @@ write_prefix(const char *buffer, size_t bytes, bool *prefixed)
 		}
 
 		if (!*prefixed) {
```
vapier, line 325: this codepath is called once per-line.  while calling this function with multiline strings is certainly valid, and i'm sure done in places, it's not the common case, and I have a hard time seeing strlen being called on tiny strings showing up in any perf situation.

len_ec & ec_normal will be like 7 bytes each.  len_prefix will be ~length of longest service name, and that's likely to also be quite small.

src/openrc-shutdown/broadcast.c
```diff
@@ -144,10 +144,11 @@ void broadcast(char *text)
 
 	setutxent();
 
+	size_t len_path_dev = strlen(_PATH_DEV);
 	while ((utmp = getutxent()) != NULL) {
 		if (utmp->ut_type != USER_PROCESS || utmp->ut_user[0] == 0)
 			continue;
-		if (strncmp(utmp->ut_line, _PATH_DEV, strlen(_PATH_DEV)) == 0)
```
vapier, line 150: `_PATH_DEV` is a constant string.  if the compiler can't optimize that to a constant integer at compile time, it's not worth caring about.  drop this change please.

src/librc/librc-daemon.c
```diff
@@ -172,7 +173,7 @@ rc_find_pids(const char *exec, const char *const *argv, uid_t uid, pid_t pid)
 				proc_ns[0] = '\0';
 		}
 		free(buffer);
-		if (pid == 0 && strlen(my_ns) && strlen (proc_ns) && strcmp(my_ns, proc_ns))
+		if (pid == 0 && len_my_ns && strlen (proc_ns) && strcmp(my_ns, proc_ns))
```
vapier, line 176: these strlen calls simply want to know if the string is empty.  that can be checked by `my_ns[0]` and `proc_ns[0]` -- no need for strlen at all.

PR #795, opened 2025-03-15
openrc-init: use self-pipe to handle signals safely

N-R-K:
the current code uses a bunch of signal unsafe functions inside
the signal handlers. this gets those code outside using
self-pipe trick so that there's no more issues.

also save and restore errno inside signal handlers in order to
avoid clobbering errno for the calling code.

the SIGCHLD handler is still seperate since we want to reap
zombies as soon as possible.

Ref: https://github.com/OpenRC/openrc/issues/589

### DIFF
--- a/src/openrc-init/openrc-init.c
+++ b/src/openrc-init/openrc-init.c
@@ -31,6 +31,8 @@
 #include <sys/reboot.h>
 #include <sys/types.h>
 #include <sys/wait.h>
+#include <fcntl.h>
+#include <poll.h>
 #include <utmp.h>
 
 #ifdef HAVE_SELINUX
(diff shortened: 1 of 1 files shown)


src/openrc-init/openrc-init.c
```diff
@@ -195,39 +210,18 @@ static void reap_zombies(void)
 		else if (pid == -1) {
 			if (errno == ECHILD)
 				break;
-			perror("waitpid");
+			write(2, errmsg, sizeof(errmsg) - 1);
```
navi-desu, line 213: `STDERR_FILENO` is preferred iirc

N-R-K: Done. Also added a `perror` to the `read(sigpipe)` call just in case.

Issue #797, opened 2025-03-15
openrc-init: Faster shutdown


N-R-K:
Currently openrc-init has a hardcoded 3 sec delay here during shutdown:

https://github.com/OpenRC/openrc/blob/5c2b288a86fc3e9f2baf9a9b07493e3bab9f9c0e/src/openrc-init/openrc-init.c#L113-L118

This is quite annoying and I've patched it down to 1 second in my system for quite some time now.

Instead of sleeping, maybe we can check if there's any children left to reap or not? And if there's nothing to reap, we can shutdown early.

PR #798, opened 2025-03-15
openrc-init: faster shutdown 

N-R-K:
rather than waiting 3 seconds always, check if there's any
reapable children left or not. if there are none left, we can
shutdown early.

Fixes: https://github.com/OpenRC/openrc/issues/797

### DIFF
--- a/src/openrc-init/openrc-init.c
+++ b/src/openrc-init/openrc-init.c
@@ -46,6 +46,12 @@
 #include "helpers.h"
 #include "version.h"
 
+/* on linux CLOCK_BOOTTIME provides proper POSIX CLOCK_MONOTONIC behavior. */
+#if defined(__linux__)
+#  undef  CLOCK_MONOTONIC
+#  define CLOCK_MONOTONIC CLOCK_BOOTTIME
+#endif
+
 static const char *path_default = "/sbin:/usr/sbin:/bin:/usr/bin";
 static const char *rc_default_runlevel = "default";
 static int sigpipe[2] = { -1, -1 };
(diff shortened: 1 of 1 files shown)


navi-desu:
for the second commit, i'd rather a more descriptive shortline, maybe "openrc-init: speed up shutdown by waiting for children", the rest of the commit seems fine (i was unsure if should've just rebased it and pushed myself, so, asking)

N-R-K:
> i'd rather a more descriptive shortline

Done. I'd wait 1 or 2 days before merging though, giving others a chance to review. There might be some weird edge cases where this may regress.

src/openrc-init/openrc-init.c
```diff
@@ -94,15 +101,40 @@ static void handle_reexec(char *my_name)
...
-	ts.tv_nsec = 0;
-	nanosleep(&ts, NULL);
-	printf("Sending the final kill signal\n");
+	clock_gettime(CLOCK_MONOTONIC, &start);
+
+	/* ensure *at least* half a second of grace time before KILL */
+	tmp.tv_sec = 0;
+	tmp.tv_nsec = 500 * ms_to_ns;
+	while (clock_nanosleep(CLOCK_MONOTONIC, 0, &tmp, &remain) < 0 && errno == EINTR)
+		tmp = remain;
```
navi-desu, line 116: is there a need to wait this half a second if we're waiting at least 3 seconds anyway?

navi-desu:
didn't manage to get time to test this myself today, sorry. i'll do so tomorrow and then probably merge it

src/openrc-init/openrc-init.c
navi-desu: i meant at most, sorry.

the issue is then hitting the break before any process actually exits, if i got it right

makes sense

PR #801, opened 2025-03-17
build: _GNU_SOURCE related cleanups

N-R-K:


N-R-K:
This also looks very fishy

https://github.com/OpenRC/openrc/blob/028eadec44f6321652fede7fb6eae77b0131bd90/meson.build#L184-L186

It's testing with `_GNU_SOURCE` defined but doesn't add it to project args...

navi-desu (approved):

PR #803, opened 2025-03-17
Quiet boot

N-R-K:


src/openrc-init/openrc-init.c
```diff
@@ -255,7 +259,10 @@ int main(int argc, char **argv)
 	}
 #endif
 
-	printf("OpenRC init version %s starting\n", VERSION);
+	quiet = rc_yesno(getenv("EINFO_QUIET"));
```
N-R-K, line 262: Using `EINFO_QUIET` for this is kinda bad. Since it doesn't really have anything to do with the `einfo` stuff and also because this will be forwarded to `openrc` as well, which the user might not be wanting.

Maybe something like "OPENRC_INIT_QUIET" ?

navi-desu: can one even pass environment variables to init at boot?

a kernel cmdline like `--quiet` makes sense as well imo

navi-desu:
is there a point to silencing the few lines openrc-init prints, if the underlying `openrc` process will output a lot right away?

unsure if silencing both is desirable though

N-R-K:
> unsure if silencing both is desirable though

It is for some, see https://github.com/OpenRC/openrc/issues/487#issuecomment-1426406542.

Though this patch doesn't fully silence everything, some services might still print to the console.

src/openrc-init/openrc-init.c
navi-desu: the `quiet` kernel parameter is already known, and systemd even defaults to only showing errors should `quiet` be a param, so i think i makes sense to adopt that, and, if `quiet` is set we set EINFO_QUIET to true-- can't do anything about scripts that don't use einfo tho

N-R-K: > the `quiet` kernel parameter is already known

The problem is, it's not passed to the init since the kernel accepted it. So you need to read `/proc/cmdline` and parse it to get it. I don't think this is worthwhile to do given PID 1 should be simple and properly parsing `/proc/cmdline` will not be super trivial. Yes, systemd does it, but I don't consider that *by itself* a convincing reason to adopt it. I'm much more open towards a `--quiet|-q` flag.

navi-desu: librc already had `rc_proc_getent` that does that (used by openrc itself), and openrc-init does already link librc, so i don't see it as any increase in complexity, something like

```suggestion
    char *kernel_quiet = rc_proc_getent("quiet");
    if (kernel_quiet) {
	quiet = true;
	setenv("EINFO_QUIET", "yes", true);
    } else {
        quiet = rc_yesno(getenv("EINFO_QUIET"));
    }
    free(kernel_quiet);
```

N-R-K: > librc already had `rc_proc_getent` that does that

Hmmm, didn't know that. But looking at it

https://github.com/OpenRC/openrc/blob/028eadec44f6321652fede7fb6eae77b0131bd90/src/librc/librc-misc.c#L127-L128

I see that it's doing the exact thing that I *did not* want to do: blindly splitting on spaces. If I put this into my kernel param:

```
somemod.param="one quiet two"
```

This will incorrectly pick up "quiet" from `somemod`. This is why I said *proper* `/proc/cmdline` parse will not be trivial, as you'd need to do escaping and double quote handling.

N-R-K: Well, even if you fix the bug, it's still not a good idea. What if I want the kernel to be quiet but openrc to still print things?

src/openrc-init/openrc-init.c
```diff
@@ -54,6 +54,7 @@
 static const char *path_default = "/sbin:/usr/sbin:/bin:/usr/bin";
 static const char *rc_default_runlevel = "default";
 static int sigpipe[2] = { -1, -1 };
+static int quiet = 0;
```
navi-desu, line 57:
```suggestion
static bool quiet = false;
```

navi-desu (changes requested):

navi-desu:
thinking about it openrc-init should probably just use libeinfo, for consistent styling and automatic respect for flags like that

but william might have things to say about that, so for now, this lgtm

PR #806, opened 2025-03-18
openrc-run: Support standard input redirection too

nmeum:
OpenRC already supports redirecting standard output/error of services. This patch adds support for redirecting standard input as well. I encountered the lack of this feature while working on a user service for a Wayland compositor status bar which reads status text to be displayed from stdin. Input redirection allows me to connect this user services with another user service providing the status text via a named pipe.

Note that input redirection is also supported by systemd (and potentially other service supervisors) through the `StandardInput=` directive [\[1\]][1]. I intentionally did not implement redirection from a program as an input as I believe that in such a case the input program (assuming it is long-running) should be supervised by OpenRC as well and the two should be connected through a named pipe. 

[1]: https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#StandardInput=

### DIFF
--- a/man/openrc-run.8
+++ b/man/openrc-run.8
@@ -158,18 +158,23 @@ use this to change the user id, and optionally group id,  before
 or
 .Xr supervise-daemon 8
 launches the daemon.
-.It Ar output_log
-This is the path to a file or named pipe where the standard output from
-the service will be redirected. If you are starting this service with
+.It Ar input_file
+This is the path to a file or named pipe to which the standard input of the
+service will be redirected to. If you are starting this service with
 .Xr start-stop-daemon 8 ,
 ,  you must set
 .Pa command_background
 to true. Keep in mind that this path will be inside the chroot if the
 .Pa chroot
-variable is set.
+variable is set. Contrary to output and error redirection, input redirection
+assumes that the file exists already.
+.It Ar output_log
+The same thing as
+.Pa input_file
+but for the standard output.
 .It Ar error_log
 The same thing as
-.Pa output_log
+.Pa input_file
 but for the standard error output.
 .It Ar output_logger
 This is a process which will be used to log the standard output from the
--- a/src/start-stop-daemon/start-stop-daemon.c
+++ b/src/start-stop-daemon/start-stop-daemon.c
@@ -94,7 +94,7 @@ enum {
 
 const char *applet = NULL;
 const char *extraopts = NULL;
-const char getoptstring[] = "I:KN:PR:Sa:bc:d:e:g:ik:mn:op:s:tu:r:w:x:1:2:3:4:" \
+const char getoptstring[] = "I:KN:PR:Sa:bc:d:e:g:ik:mn:op:s:tu:r:w:x:0:1:2:3:4:" \
 	getoptstring_COMMON;
 const struct option longopts[] = {
 	{ "capabilities", 1, NULL, LONGOPT_CAPABILITIES},
@@ -124,6 +124,7 @@ const struct option longopts[] = {
 	{ "chroot",       1, NULL, 'r'},
 	{ "wait",         1, NULL, 'w'},
 	{ "exec",         1, NULL, 'x'},
+	{ "stdin",        1, NULL, '0'},
 	{ "stdout",       1, NULL, '1'},
 	{ "stderr",       1, NULL, '2'},
 	{ "stdout-logger",1, NULL, '3'},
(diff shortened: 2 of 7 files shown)


man/openrc-run.8
```diff
@@ -158,7 +158,7 @@ use this to change the user id, and optionally group id,  before
 or
 .Xr supervise-daemon 8
 launches the daemon.
-.It Ar output_log
+.It Ar input_file
 This is the path to a file or named pipe where the standard output from
```
navi-desu, line 162: still mentions standard output

nmeum: Good catch! Fixed.

navi-desu:
> I intentionally did not implement redirection from a program as an input as I believe that in such a case the input program (assuming it is long-running) should be supervised by OpenRC as well and the two should be connected through a named pipe.

the same could be argued for {output,error}_logger tbh, which fwiw i want to default to posix `logger` in #783 -- this now gives me thoughts about if we want to supervise the logger too

src/start-stop-daemon/start-stop-daemon.c
```diff
@@ -1071,6 +1078,13 @@ int main(int argc, char **argv)
 		stdin_fd = devnull_fd;
 		stdout_fd = devnull_fd;
 		stderr_fd = devnull_fd;
+		if (redirect_stdin) {
+			if ((stdin_fd = open(redirect_stdin,
+				    O_RDONLY)) == -1)
```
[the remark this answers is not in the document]
nmeum: > Intentional? Then it should be documented properly since the current documentation makes it seem like -0 -1 -2 behave the same way.

Oh you are absolutely right. Didn't think about this too much as I just had my "another services creates a named pipe" use case in mind. I guess we should create it as well to be consistent with the other options? However, creating an empty regular file also doesn't make much sense it this use case, or does it? What do you think? Definitely should be documented in any case.

> What's systemd's behavior in cases where the stdin file doesn't exist, does it create it?

Unfortunately, [the documentation](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#StandardInput=) is not very explicit about it:

> The file:path option may be used to connect a specific file system object to standard input. An absolute path following the ":" character is expected, which may refer to a regular file, a FIFO or special file. If an AF_UNIX socket in the file system is specified, a stream socket is connected to it. The latter is useful for connecting standard input of processes to arbitrary system services.

Need to check the source or try it out.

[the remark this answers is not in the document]
nmeum: Looks like systemd also does not create the file:

https://github.com/systemd/systemd/blob/v257.4/src/core/exec-invoke.c#L424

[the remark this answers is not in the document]
nmeum: Right, but we don't support sockets for output redirection either so I think it is out-of-scope for this PR.

I updated the man pages to document that input redirection doesn't create the file. Note that this is also how shell input redirection works, i.e. `< foo` doesn't create the file `foo` and errors if it doesn't exist so I think this is the behavior we want here.

man/openrc-run.8
```diff
@@ -158,18 +158,23 @@ use this to change the user id, and optionally group id,  before
 or
 .Xr supervise-daemon 8
 launches the daemon.
-.It Ar output_log
-This is the path to a file or named pipe where the standard output from
+.It Ar input_file
+This is the path to a file or named pipe where the standard input from
 the service will be redirected. If you are starting this service with
```
N-R-K, line 163:
```suggestion
This is the path to a file or named pipe to which the standard input of the
service will be redirected to. If you are starting this service with
```

man/openrc-run.8
```diff
@@ -158,18 +158,23 @@ use this to change the user id, and optionally group id,  before
...
 .Xr start-stop-daemon 8 ,
 ,  you must set
 .Pa command_background
 to true. Keep in mind that this path will be inside the chroot if the
 .Pa chroot
-variable is set.
+variable is set. Input redirection only opens the file for reading and
+(contrary to output redirection) doesn't create it, if it doesn't exist.
```
N-R-K, line 170:
```suggestion
(contrary to output and error redirection) doesn't create it, if it doesn't exist.
```

N-R-K (commented):
Code wise looks okay. Some small documentation rewording suggestions:

N-R-K (approved):

PR #807, opened 2025-03-19
s-s-d, supervise-daemon: handle "/dev/stderr" for output_log

N-R-K:
useful to be able to redirect both standard output and error into a single logger process, which currently isn't quite possible.


N-R-K:
> redirect both standard output and error into a single logger process

Other ideas are welcome. This is just a quick patch.

PR #809, opened 2025-03-20
openrc-init: unmask signals in case of error also

N-R-K:


N-R-K:
Fixed in https://github.com/OpenRC/openrc/pull/810 instead.

PR #810, opened 2025-03-21
rc_exec: add a centralized exec function

N-R-K:
Depends on: https://github.com/OpenRC/openrc/pull/801


src/shared/rc_exec.h
```diff
@@ -0,0 +1,55 @@
...
+	RC_EXEC_DEVNULL = -3,
+};
+
+typedef struct {
+	const char **argv;
+	int redirect_stdin;
+	int redirect_stdout;
+	int redirect_stderr;
+	bool setsid : 1;
+} rc_exec_args_t;
```
navi-desu, line 40: `_t` suffix is reserved to the system/platform, by posix (though it matters very little here since the `rc` prefix is ours)

but also, we should avoid unnecessary typedefs, a goal for 1.0 is to remove them from librc/rc.h, so i'd like to avoid adding more

since this is internal, we could even drop the `rc_` prefix all together also?

src/shared/rc_exec.c
```diff
@@ -0,0 +1,162 @@
...
+			goto exit;
+		args->redirect_stderr = stderr_pipe[1];
+	}
+
+	if (args->redirect_stdin == RC_EXEC_DEVNULL)
+		args->redirect_stdin = devnull();
+	if (args->redirect_stdout == RC_EXEC_DEVNULL)
+		args->redirect_stdout = devnull();
+	if (args->redirect_stderr == RC_EXEC_DEVNULL)
+		args->redirect_stderr = devnull();
```
navi-desu, line 84: this would open 3 different file descriptors to /dev/null even when nothing uses it... and leak them to the parent (with cloexec but still)

maybe something like:
```c
args->redirect_std = devnullfd != -1 ? devnullfd : (devnullfd = devnull());
```
or

have devull() take an int pointer and store it there, and either option, close on exit?

src/shared/rc_exec.h
N-R-K: I added it since we're already using struct typedefs. But sure, I can remove it and also drop the `rc_` prefix.

src/shared/rc_exec.c
N-R-K: > this would open 3 different file descriptors

It would only open a single one per process, the `devnullfd` is `static`.

> and leak them to the parent

Do you think it's worth closing it? Seems harmless, especially since it's `CLOEXEC`.

navi-desu: oh sorry, i missed the static, nah, it's fine then since it'll be used by any possible call to exec until the end of the process, resolving.

PR #811, opened 2025-03-22
rc_unmount: add parallel unmounting

N-R-K:
this turns mouninfo into a multicall binary which can also do
parallel unmounting. most mountinfo code is unchanged; only
notable change is find_mounts() now returns the number of mounts
found and also takes the "process" function as an argument; so
that is_mounted() can be implemented on top of it.

rc_unmount mostly follows the logic of previous do_unmount. some
notable changes:

- rc_unmount is "lazy" when it comes to retrying failed
  unmounts. it will greedily keep running unmount as long as it
  can before it looks at the "waiting" queue.
- rc_unmount will check if the mountpoint is still mounted or
  not when umount returns non-zero exit code. this is due to the
  fact that multiple umount calls might race to unmount a shared
  mount. so if umount fails _but_ the mountpoint is no longer
  mounted, we assume success.
- do_unmount used to fail if fuser did not find any pids using
  the mount. rc_unmount tries one more time; the rationale being
  that there's a gap between the umount call and the fuser call,
  and so whatever was using the mount before might have stopped
  using it now and so it's worth another attempt.

Closes: https://github.com/OpenRC/openrc/issues/662


src/mountinfo/mountinfo.c
```diff
@@ -356,6 +385,174 @@ find_mounts(struct args *args)
...
+			return -1;
+		}
+		/* it's possible that whatever was using the mount stopped
+		 * using it now, so allow 1 more retry */
+		rp->try_count = TRY_MAX;
+		return 0;
+	} else if (strcmp(buf, " - ") == 0) {
+		/* NRK: huh?? */
+		eerror("Unmounting %s failed", rp->mntpath);
+		return -1;
```
N-R-K, line 535: I don't quite get when this is supposed to happen. Git blame takes me [to here](https://github.com/OpenRC/openrc/commit/ff162b39ace915ed9803ff7d091157855cf6d48c) but why would fuser not being preset output "-"?

tools/manymounts.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+# can be used for testing rc_unmount:
+#    # ./tools/manymounts.sh
+#    # rc_unmount -- -p '^/tmp/manymounts.*'
```
N-R-K, line 4: Not sure if this should be kept or not. Included it so that others can test easily. Can remove if desired.

N-R-K (commented):
Depends on: https://github.com/OpenRC/openrc/pull/810

src/mountinfo/mountinfo.c
N-R-K: > Git blame takes me [to here](https://github.com/OpenRC/openrc/commit/ff162b39ace915ed9803ff7d091157855cf6d48c) but why would fuser not being preset output "-"?

Was pointed out by @dwfreed that the pid var is set to '-' initially and the command is only run if fuser is preset. I missed this completely.

Removed this part since we keep track of exec failure and treat it as a proxy for fuser not being present.

src/mountinfo/mountinfo.c
```diff
@@ -382,19 +578,67 @@ int main(int argc, char **argv)
...
+			if (strcmp(tmps, "--") == 0)
+				break;
+			if (umount_args_num >= (int)ARRAY_SIZE(umount_args))
+				eerrorx("%s: Too many umount arguments", applet);
+			umount_args[umount_args_num++] = tmps;
+		}
+
+		tmps = rc_conf_value("rc_fuser_timeout");
+		if (!tmps || sscanf(tmps, "%d", &rc_fuser_timeout) != 1)
+			rc_fuser_timeout = 60;
```
N-R-K, line 631: rc.conf documents this option as following:

```sh
# This is how long fuser should wait for a remote server to respond. The
# default is 60 seconds, but  it can be adjusted here.
#rc_fuser_timeout=60
```

Which makes it seem like it only accepts integer seconds as value, so this change might be fine.

But worth noting that the older script simply passed this over to `timeout(1)` as is. Which meant it would accept stuff like "40m" or "16s" etc even though not explicitly documented.

Do we care about this? If yes: do we just parse similar to `timeout(1)` so that everything works as usual? Or emit a warning if sscanf fails instead of silently defaulting to 60 so the user knows something was messed up?

(I thought I had commented this already but apparently not).

N-R-K: Using `parse_duration` from `timeutils.h` it can accept most of the suffixes now (`s`, `m` and `h`). `d` (for days) is missing, but that seems unlikely someone would set a timeout in days.

It only accepts integer though, whereas `timeout(1)` accepts fractional values.

src/mountinfo/mountinfo.c
```diff
@@ -356,6 +381,157 @@ find_mounts(struct args *args)
...
+			"fuser", sig, "-k", fuser_opt, rp->mntpath, NULL
+		};
+		struct exec_result res;
+		struct exec_args args = exec_init(argv);
+		args.redirect_stdout = args.redirect_stderr = EXEC_DEVNULL;
+		snprintf(sig, sizeof sig, "%s%s", fuser_kill_prefix,
+			rp->try_count == TRY_MAX ? "KILL" : "TERM");
+		res = do_exec(&args);
+		if (res.pid > 0)
+			rc_waitpid(res.pid);
```
N-R-K, line 530: The first fuser command to query the pids has a timeout and is async, this one to send the signals on the other hand is synchronous and has no timeout (copied from the `rc-mount.sh` script).

But this doesn't seem to make sense (?) If we anticipate fuser might take a while then shouldn't we be timing out *both* of the calls rather than just one of them?

Also the only reason we're doing 2 commands (query pids + send signal) is to avoid killing ourselves if we happen to be using that mount. If we can figure out whether we're using it or not beforehand then it can be changed into a single fuser call which directly sends the signal (async, with timeout).

navi-desu:
rebased and renamed it back to do_unmount, and removed the manpage entry, i'd like to keep it internal for now, like it used to be

PR #819, opened 2025-03-28
librc: use dirfd and *at functions.

navi-desu:
gives us atomic operations, removes almost all path string allocations, and reduces the number of syscalls

### DIFF
--- a/src/librc/librc-depend.c
+++ b/src/librc/librc-depend.c
@@ -689,17 +689,6 @@ rc_deptree_update_needed(time_t *newest, char *file)
 	RC_STRING *s;
 	struct stat buf;
 	time_t mtime;
-	char *path;
-	const char *service_dir = rc_svcdir();
-
-	/* Create base directories if needed */
-	if (mkdir(service_dir, 0755) != 0 && errno != EEXIST)
-		fprintf(stderr, "mkdir '%s': %s\n", service_dir, strerror(errno));
-
-	for (size_t i = 1; i < RC_DIR_SYS_MAX; i++) {
-		if (mkdirat(rc_dirfd(RC_DIR_SVCDIR), dirnames[i], 0755) != 0 && errno != EEXIST)
-			fprintf(stderr, "mkdir `%s': %s\n", dirnames[i], strerror(errno));
-	}
 
 	/* Quick test to see if anything we use has changed and we have
 	 * data in our deptree. */
--- a/src/librc/librc-depend.c
+++ b/src/librc/librc-depend.c
@@ -685,6 +685,7 @@ bool
 rc_deptree_update_needed(time_t *newest, char *file)
 {
 	bool newer = false;
+	const int *dirfds;
 	RC_STRINGLIST *config;
 	RC_STRING *s;
 	struct stat buf;
@@ -704,13 +705,9 @@ rc_deptree_update_needed(time_t *newest, char *file)
 		mtime = time(NULL);
 	}
 
-	for (const char * const *dirs = rc_scriptdirs(); *dirs; dirs++) {
-		static const char *subdirs[] = { "init.d", "conf.d", NULL };
-		for (const char **subdir = subdirs; *subdir; subdir++) {
-			xasprintf(&path, "%s/%s", *dirs, *subdir);
-			newer |= !deep_mtime_check(AT_FDCWD, path, true, &mtime, file);
-			free(path);
-		}
+	for (size_t i = 0, count = rc_scriptdirfds(&dirfds); i < count; i++) {
+		newer |= !deep_mtime_check(dirfds[i], "init.d", true, &mtime, file);
+		newer |= !deep_mtime_check(dirfds[i], "conf.d", true, &mtime, file);
 	}
 
 	newer |= !deep_mtime_check(rc_dirfd(RC_DIR_SYSCONF), "rc.conf", true, &mtime, file);
--- a/src/librc/librc-depend.c
+++ b/src/librc/librc-depend.c
@@ -145,22 +145,9 @@ make_deptree(void) {
 	return deptree;
 }
 
-RC_DEPTREE *
-rc_deptree_load(void) {
-	char *deptree_cache;
-	RC_DEPTREE *deptree;
-
-	xasprintf(&deptree_cache, "%s/deptree", rc_svcdir());
-	deptree = rc_deptree_load_file(deptree_cache);
-	free(deptree_cache);
-
-	return deptree;
-}
-
-RC_DEPTREE *
-rc_deptree_load_file(const char *deptree_file)
+static RC_DEPTREE *
+deptree_load_file(int dirfd, const char *pathname)
 {
-	FILE *fp;
 	RC_DEPTREE *deptree;
 	RC_DEPINFO *depinfo = NULL;
 	RC_DEPTYPE *deptype = NULL;
@@ -170,8 +157,9 @@ rc_deptree_load_file(const char *deptree_file)
 	char *p;
 	char *e;
 	int i;
+	FILE *fp;
 
-	if (!(fp = fopen(deptree_file, "r")))
+	if (!(fp = do_fopenat(dirfd, pathname, O_RDONLY)))
 		return NULL;
 
 	deptree = make_deptree();
@@ -205,10 +193,22 @@ rc_deptree_load_file(const char *deptree_file)
 			deptype = make_deptype(depinfo, type);
 		rc_stringlist_add(deptype->services, e);
 	}
-	fclose(fp);
 	free(line);
+	fclose(fp);
 
 	return deptree;
+}
+
+RC_DEPTREE *
+rc_deptree_load(void)
+{
+	return deptree_load_file(rc_dirfd(RC_DIR_SVCDIR), "deptree");
+}
+
+RC_DEPTREE *
+rc_deptree_load_file(const char *deptree_file)
+{
+	return deptree_load_file(AT_FDCWD, deptree_file);
 }
 
 static bool
@@ -584,18 +584,15 @@ rc_deptree_order(const RC_DEPTREE *deptree, const char *runlevel, int options)
    oldest (or newest) found.
 */
 static bool
-deep_mtime_check(const char *target, bool newer,
-	    time_t *rel, char *file)
+deep_mtime_check(int target_dir, const char *target, bool newer, time_t *rel, char *file)
 {
 	struct stat buf;
 	bool retval = true;
 	DIR *dp;
 	struct dirent *d;
-	char path[PATH_MAX];
-	int serrno = errno;
 
 	/* If target does not exist, return true to mimic shell test */
-	if (stat(target, &buf) != 0)
+	if (fstatat(target_dir, target, &buf, 0) != 0)
 		return true;
 
 	if (newer) {
@@ -616,21 +613,18 @@ deep_mtime_check(const char *target, bool newer,
 		}
 	}
 
-	/* If not a dir then reset errno */
-	if (!(dp = opendir(target))) {
-		errno = serrno;
+	if (!S_ISDIR(buf.st_mode) || !(dp = do_opendirat(target_dir, target)))
 		return retval;
-	}
 
 	/* Check all the entries in the dir */
 	while ((d = readdir(dp))) {
 		if (d->d_name[0] == '.')
 			continue;
-		snprintf(path, sizeof(path), "%s/%s", target, d->d_name);
-		if (!deep_mtime_check(path, newer, rel, file)) {
+		if (!deep_mtime_check(dirfd(dp), d->d_name, newer, rel, file)) {
 			retval = false;
 		}
 	}
+
 	closedir(dp);
 	return retval;
 }
@@ -640,8 +634,7 @@ deep_mtime_check(const char *target, bool newer,
  * the return value arguments are non-null).
  */
 static bool
-mtime_check(const char *source, const char *target, bool newer,
-	    time_t *rel, char *file)
+mtime_check(int dirfd, const char *source, const char *target, bool newer, time_t *rel, char *file)
 {
 	struct stat buf;
 	time_t mtime;
@@ -652,7 +645,7 @@ mtime_check(const char *source, const char *target, bool newer,
 		return false;
 	mtime = buf.st_mtime;
 
-	retval = deep_mtime_check(target,newer,&mtime,file);
+	retval = deep_mtime_check(dirfd, target,newer,&mtime,file);
 	if (rel) {
 		*rel = mtime;
 	}
@@ -660,18 +653,16 @@ mtime_check(const char *source, const char *target, bool newer,
 }
 
 bool
-rc_newer_than(const char *source, const char *target,
-	      time_t *newest, char *file)
+rc_newer_than(const char *source, const char *target, time_t *newest, char *file)
 {
 
-	return mtime_check(source, target, true, newest, file);
+	return mtime_check(AT_FDCWD, source, target, true, newest, file);
 }
 
 bool
-rc_older_than(const char *source, const char *target,
-	      time_t *oldest, char *file)
+rc_older_than(const char *source, const char *target, time_t *oldest, char *file)
 {
-	return mtime_check(source, target, false, oldest, file);
+	return mtime_check(AT_FDCWD, source, target, false, oldest, file);
 }
 
 typedef struct deppair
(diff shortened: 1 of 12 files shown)


src/librc/librc.c
```diff
@@ -620,6 +617,24 @@ rc_is_user(void)
...
+		return -1;
+
+	if ((idx = scriptdir - scriptdir_array) > SCRIPTDIR_MAX)
+		return -1;
+
+	if (scriptdirfds[idx] <= 0)
+		scriptdirfds[idx] = open(*scriptdir, O_RDONLY | O_CLOEXEC | O_DIRECTORY);
+
+	return scriptdirfds[idx];
+}
```
navi-desu, line 636: instead of this, we could make the scriptdirs enum public, remove the `rc_*dir` functions and make them actual extern symbols from rc.h

then instead of
```c
for (const char * const *dirs = rc_scriptdirs(); *dirs; dirs++) {
    int dirfd = rc_scriptdirs(dirs);
}
```
the api would be:
```c
for (enum rc_scriptdir_entry dir = 0; dir < RC_SCRIPTDIR_LAST; dir++) {
    int dirfd = rc_scriptdir(dir);
    /* or */
    const char *dirpath = rc_scriptdir[dir];
}
```

seems nicer to me imo

navi-desu:
i'll fix the ci tomorrow i left a trailing whitespace somewhere

N-R-K (commented):
Just a cursory glance. Haven't looked too deeply since it's a pretty large change.

src/shared/helpers.h
```diff
@@ -120,6 +130,57 @@ RC_UNUSED static bool existss(const char *pathname)
...
+		return NULL;
+
+	/* O_CREAT and O_TRUNC in modes 'a', 'w+', 'a+' don't
+	 * really matter after the file has been opened */
+	if (mode & O_RDWR)
+		fmode = "r+";
+	else if (mode & O_RDONLY)
+		fmode = "r";
+	else if (mode & O_WRONLY)
+		fmode = "w";
```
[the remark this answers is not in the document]
N-R-K: Actually, `RDONLY` maybe 0, [according to the posix manpage](https://www.man7.org/linux/man-pages/man3/open.3p.html). Which is also the case on my system.

> In historical implementations the value of O_RDONLY is zero.
> Because of that, it is not possible to detect the presence of
> O_RDONLY and another option. Future implementations should encode
> O_RDONLY and O_WRONLY as bit flags so that:
>
>     O_RDONLY | O_WRONLY == O_RDWR

So this should instead be structured like this:

```c
	if (mode & O_WRONLY)
	else if (mode & O_RDWR)
	else // assume O_RDONLY
```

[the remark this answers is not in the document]
navi-desu: reading a bit further,

> Future implementations should encode O_RDONLY and O_WRONLY as bit flags so that:
>           O_RDONLY | O_WRONLY == O_RDWR

in platforms that do this, it means `mode & O_WRONLY` would match `O_RDWR`, so maybe we do this instead? (unsure if this missed any case)

```c
if ((mode & O_RDWR) == O_RDWR)
else if ((mode & O_WRONLY) == O_WRONLY)
else // assume O_RDONLY
```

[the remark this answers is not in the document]
N-R-K: > it means `mode & O_WRONLY` would match `O_RDWR`, so maybe we do this instead?

Yup, that's what I meant to do but messed up the order due to copy paste.

src/librc/librc.c
navi-desu: issue: the array returned by system-wide is shorter than user services, and in a different order as well...

src/librc/librc-depend.c
```diff
@@ -211,6 +198,17 @@ rc_deptree_load_file(const char *deptree_file)
 	return deptree;
 }
 
+RC_DEPTREE *
+rc_deptree_load(void) {
```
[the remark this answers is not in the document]
navi-desu: i did meant to follow the style guide, fixed

src/librc/librc.c
```diff
@@ -45,6 +46,36 @@
 
 /* File stream used for plugins to write environ vars to */
 FILE *rc_environ_fd = NULL;
+int dirfds[RC_DIR_MAX];
```
N-R-K, line 49: Shouldn't this be `static`?

src/librc/librc.c
```diff
@@ -431,15 +425,13 @@ get_runlevel_chain(const char *runlevel, RC_STRINGLIST *level_list, RC_STRINGLIS
...
-	dirs = ls_dir(path, LS_DIR);
+	dirs = ls_dir(rc_dirfd(RC_DIR_RUNLEVEL), runlevel, LS_DIR);
 	TAILQ_FOREACH(d, dirs, entries) {
 		nextlevel = d->value;
 
 		/* Check for loop */
 		if (rc_stringlist_find(ancestor_list, nextlevel)) {
-			fprintf(stderr, "Loop detected in stacked runlevels attempting to enter runlevel %s!\n",
-			    nextlevel);
+			fprintf(stderr, "Loop detected in stacked runlevels attempting to enter runlevel %s!\n", nextlevel);
```
N-R-K, line 434: Looks like a useless change?

src/librc/librc.c
```diff
@@ -1333,33 +1232,26 @@ RC_STRINGLIST *
...
-	xasprintf(&sched_dir, "%s/scheduled", rc_svcdir());
-	dirs = ls_dir(sched_dir, 0);
+	if (!(dp = opendir_dup(rc_dirfd(RC_DIR_SCHEDULED))))
+		return list;
 
-	TAILQ_FOREACH(dir, dirs, entries) {
-		xasprintf(&file, "%s/%s/%s", sched_dir, dir->value, service);
-		if (exists(file))
-			rc_stringlist_add(list, file);
+	/* TODO: should be the full filename */
```
N-R-K, line 1255: Did you mean to leave this as a todo or are you planning to fix it before merging?

src/librc/librc-misc.c
```diff
@@ -146,16 +146,15 @@ rc_proc_getent(const char *ent RC_UNUSED)
 }
 
 RC_STRINGLIST *
-rc_config_list(const char *file)
+config_list(FILE *fp)
```
N-R-K, line 149: Seems very weird for a function named `config_list` to close a `FILE *`. I think the `!fp` and `fclose` should happen inside `rc_config_list()` instead.

src/librc/librc-misc.c
```diff
@@ -301,51 +306,52 @@ static RC_STRINGLIST *rc_config_kcl(RC_STRINGLIST *config)
...
+	rc_stringlist_free(rc_conf_d_files);
+	closedir(dp);
+
+	return;
 }
 
-RC_STRINGLIST *
-rc_config_load(const char *file)
+static RC_STRINGLIST *
+config_load(FILE *fp)
```
N-R-K, line 348: Same comment here about closing the FILE handle. Though this one is a `static` function so maybe that's fine.

N-R-K (changes requested):

src/librc/librc-misc.c
navi-desu: > Though this one is a static function so maybe that's fine.

fwiw it the other one only isn't static because we need it in librc-depend.c as well

src/librc/librc-misc.c
navi-desu: did it like that to make the functions calling it not need to care, since it's internal, but sure i can move that out

(sometimes i wish c had proper notation for "this function clears/consumes this value")

src/librc/librc.c
navi-desu: i'm planning to fix it, just, forgot to do so before undrafting

navi-desu: ah, i remember it now, tho the TODO is quite wrong

the documentation of that function reads: `/*! Return a NULL terminated list of services that are scheduled to start`

the old code returns a list of file paths to symlinks inside the schedule dir, which while technically a service since it links to one, isn't quite what one would expect i think (the returned list would become invalid as soon as the sheduled start happens for example)

so i'm unsure if it's a bug, or bad design, and if the latter if we should change it or not

we have zero consumers for that function inside openrc itself also

src/librc/librc-misc.c
navi-desu: i changed it to take the dirfd and pathname instead, seems cleaner to me, and less repeated code on callers

src/librc/librc.c
navi-desu: style change, but i shouldn't have commited it here, removed

src/librc/librc.c
N-R-K: > we have zero consumers for that function inside openrc itself also

Any consumer outside of openrc?

And also it would be good to go thru git blame or w/e to try and trace if there ever was a consumer of this. Looking at the usage code might reveal what the intention of the function originally was.

src/librc/librc-misc.c
N-R-K: > i changed it to take the dirfd and pathname instead, seems cleaner to me

That works. And caller not having to bother with either fopen or fclose seems nicer too.

src/librc/librc.c
navi-desu: > Looking at the usage code might reveal what the intention of the function originally was.

it was added in the first commit adding c code to openrc, 5af58b45146ab5253ca964738f4e45287bf963d4, where it was used on rc_service_state as "if this returns anything at all, set the RC_STATE_SCHEDULED bit in state", later the commit cb9da6a262b60255cd037f20b4cde3ab2c8a1e6a changed the null terminated char ** into RC_STRINGLIST just, inlined that portion all together, changing the call to a ls_dir() + loop + exists()-- leaving the scheduled_by function unused

> Any consumer outside of openrc?

out of the 3 projects i know/could find using librc directly, none

N-R-K: Can we just mark it as deprecated and have it be a stub that always returns an empty list (or NULL)?

src/shared/helpers.h
```diff
@@ -120,6 +130,61 @@ RC_UNUSED static bool existss(const char *pathname)
 	return (stat(pathname, &buf) == 0 && buf.st_size != 0);
 }
 
+RC_UNUSED static FILE *fopenat(int dirfd, const char *pathname, int mode)
```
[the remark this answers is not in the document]
navi-desu: prefixed them with `do_`, also changed `opendir_dup` to `do_dopendir`

i think that makes sense considering `do_exec` too

src/librc/librc.c
navi-desu: actually this PR fixes the behaviour, since in master it returns a path to the schedule symlink, not a list of sevices that schedule the current service at all

bleh

navi-desu:
currently this looks good imo, i'm only unsure about the rc_scripdirfd bits

i'll wait a bit to see if anyone has any opinion, and either merge, or ax those commits and think about that later on...

src/librc/librc-depend.c
```diff
@@ -616,21 +613,18 @@ deep_mtime_check(const char *target, bool newer,
 		}
 	}
 
-	/* If not a dir then reset errno */
-	if (!(dp = opendir(target))) {
-		errno = serrno;
```
N-R-K, line 621: Intentional? Why was it resetting errno?

src/librc/librc.c
```diff
@@ -596,6 +596,24 @@ rc_is_user(void)
 	return is_user;
 }
 
+int scriptdirfds[SCRIPTDIR_MAX];
+
+int rc_scriptdirfd(const char * const *scriptdir) {
+	const char * const *scriptdir_array = is_user ? rc_dirs.scriptdirs : scriptdirs;
+	size_t idx;
+
+	if (!scriptdir || !*scriptdir || scriptdir < scriptdir_array)
```
N-R-K, line 612: Comparing potentially unrelated pointers via relational operator is UB.
https://www.iso-9899.info/n3220.html#6.5.9p6

src/openrc/rc.c
```diff
@@ -93,37 +93,26 @@ clean_failed(void)
...
-			xasprintf(&path, "%s/%s", failed_dir, d->d_name);
-			if (unlink(path))
-				eerror("%s: unlink `%s': %s",
-				    applet, path, strerror(errno));
-			free(path);
-		}
-		closedir(dp);
-	}
+	while ((d = readdir(dp))) {
+		if (d->d_name[0] == '.' && (d->d_name[1] == '\0' || (d->d_name[1] == '.' && d->d_name[2] == '\0')))
```
N-R-K, line 102: Can you change to just using `strcmp()`?

src/librc/librc-depend.c
```diff
@@ -689,16 +689,6 @@ rc_deptree_update_needed(time_t *newest, char *file)
...
-	const char *service_dir = rc_svcdir();
-
-	/* Create base directories if needed */
-	if (mkdir(service_dir, 0755) != 0 && errno != EEXIST)
-		fprintf(stderr, "mkdir '%s': %s\n", service_dir, strerror(errno));
-
-	for (size_t i = 1; i < RC_DIR_SYS_MAX; i++) {
-		if (mkdirat(rc_dirfd(RC_DIR_SVCDIR), dirnames[i], 0755) != 0 && errno != EEXIST)
-			fprintf(stderr, "mkdir `%s': %s\n", dirnames[i], strerror(errno));
-	}
```
N-R-K, line 731: What's the rationale for this change? If there is any worthwhile reason, then should be included in the commit message.

src/librc/librc.c
```diff
@@ -596,6 +596,24 @@ rc_is_user(void)
...
+		return -1;
+
+	if ((idx = scriptdir - scriptdir_array) > SCRIPTDIR_MAX)
```
N-R-K, line 615: Same issue with pointer arithmetic. Doing arithmetic with unrelated pointer is UB: https://www.iso-9899.info/n3220.html#6.5.7p10

> When two pointers are subtracted, both shall point to elements of the same array object [...]

And so if this condition is true, you're in UB land already. Compiler can straight up remove this check entirely.

(Also, it probably meant to be `>=`).

src/librc/librc.c
```diff
@@ -596,6 +596,24 @@ rc_is_user(void)
 	return is_user;
 }
 
+int scriptdirfds[SCRIPTDIR_MAX];
+
+int rc_scriptdirfd(const char * const *scriptdir) {
```
N-R-K, line 608: This function seems pretty fragile honestly... Maybe we could do something else with less footguns. Like an iteration function?

```c
char *dir;
int dirfd;
while (rc_scriptdir_iter(&dir, &dirfd)) { /* ... */ }
```

N-R-K (changes requested):
Looks mostly okay, aside from `rc_scriptdirfd()` and friends.

src/librc/librc.c
navi-desu: i knew about the issue of cross allocation pointer, and that's why i was so unsure about it

i like the iter, tho i think we can bundle it in a struct similar to `struct dirent`, and
```c
struct rc_scriptdir {
    const int idx;
    const int dirfd;
    cosnt char * const path;
}

/* ... */

struct rc_scriptdir sd = {0};
while (rc_scriptdir(&sd)) {
    /* ... */
}
```

i like this, and i even like this a lot more than `const char * const *` we return right now, so i'm in favor of replacing that api completely with this, wdyt?

N-R-K: Seems fine to me.

src/librc/librc.c
navi-desu: revised api returns the lenght and stores an array in the argument, safer, less ub potential, overall simpler imo

src/librc/librc.c
navi-desu: new api: `size_t rc_scriptdirs(struct rc_scriptdir **)`, returns lenght, stores an array of `struct rc_scriptdir { int dirfd, char *path }`, wdyt?

to also consider, without a structure: `size_t rc_scriptdir(_Nullable int **dirfds, _Nullable const char **paths)`

src/librc/librc-depend.c
navi-desu: "we create those on demand now so why send off a bunch of syscalls here" was all i thought tbh

src/librc/librc-depend.c
navi-desu: i think it's so that errno doesn't become `ENOTDIR` in a "successful" call, since it used to rely on `opendir` failing to not recurse, since there's a check for S_ISDIR now, there's no reason to do that

(arguably we should split those two calls, so that !S_ISDIR returns retval, while !do_opendirat returns false, though considering our usage of the function, doesn't matter much)

PR #820, opened 2025-03-29
cgroups: refactor cgroups

navi-desu:
add a deprecation warning for cgroups v1, and rework rc-cgroups.sh into cgroups.sh with an api focused on cgroups v2

move into a new layout with a supervision root cgroup as /rc, services will then go in /rc/$RC_SVCNAME, and compatibility with the old /openrc.$RC_SVCNAME is kept though restarting a service will use the new layout

this also fixes a bug where cgroup_cleanup would not ever trigger due to a bug in cgroup_running

i'll also add support for cgroups in user services to this PR later


N-R-K:
It'd be easier to review (and cleaner git log for reading and/or bisecting) if the cosmetic changes and functional changes were separated into different commits.

navi-desu:
> It'd be easier to review (and cleaner git log for reading and/or bisecting) if the cosmetic changes and functional changes were separated into different commits.

tried my best to split them, tho 4/5 will still be a big commit since it changes the internal cgroups api

N-R-K:
> tho 4/5 will still be a big commit

That's fine. I don't suggest splitting needlessly just to reduce commit size. If a cosmetic change unintentionally changes functionality then it's not immediately obvious whether it was intentional or not if it's lumped into a functional commit. Having them on their own commit makes this clearer.

negril:
It might be neat if `user.$USER` is done as `user/$USER` so that one can limit user services as a whole.

PR #829, opened 2025-04-06
install agetty symlinks via meson

N-R-K:
also adds a 'agetty' array option so that user can specify which symlinks he wants to create.

Closes: https://github.com/OpenRC/openrc/pull/355


navi-desu:
lgtm, tho i'd rather we use 'they' when referring to users in commit messages and documentation

PR #830, opened 2025-04-06
wait for exclusive fds properly

navi-desu:
* misc: fix the exclusive lockfile fd remaining open in the caller of `exec_service`
* openrc-run: properly wait for the exclusive flock in svc_wait and svc_exec
* openrc: don't solve start dependencies, instead have services start their own dependencies, it simplifies the overall mechanism, aligns the code path of `openrc` with the codepath of `rc-service`, and i think that allows for better parallelism (not sure, need to test some more)

### DIFF
--- a/src/openrc-run/openrc-run.c
+++ b/src/openrc-run/openrc-run.c
@@ -35,6 +35,7 @@
 #include <time.h>
 #include <unistd.h>
 #include <stdbool.h>
+#include <inttypes.h>
 
 #if defined(__linux__) || (defined(__FreeBSD_kernel__) && defined(__GLIBC__)) \
 	|| defined(__GNU__)
@@ -50,16 +51,14 @@
 #include "rc.h"
 #include "rc_exec.h"
 #include "misc.h"
+#include "timeutils.h"
 #include "plugin.h"
 #include "selinux.h"
 #include "_usage.h"
 #include "helpers.h"
 
-#define WAIT_INTERVAL	20000000	/* usecs to poll the lock file */
 #define WAIT_TIMEOUT	60		/* seconds until we timeout */
 #define WARN_TIMEOUT	10		/* warn about this every N seconds */
-
-static const struct timespec interval = { .tv_nsec = WAIT_INTERVAL };
 
 const char *applet = NULL;
 const char *extraopts = "stop | start | restart | status | describe | zap";
@@ -364,8 +363,7 @@ svc_exec(const char *arg1, const char *arg2)
 	int slave_tty;
 	sigset_t sigchldmask;
 	sigset_t oldmask;
-	struct timespec timeout = { .tv_sec = WAIT_TIMEOUT };
-	struct timespec warn = { .tv_sec = WARN_TIMEOUT };
+	int64_t wait_timeout, warn_timeout, now;
 	RC_STRINGLIST *keywords;
 	bool forever;
 
@@ -401,6 +399,9 @@ svc_exec(const char *arg1, const char *arg2)
 			fcntl(slave_tty, F_SETFD, flags | FD_CLOEXEC);
 	}
 
+	now = tm_now();
+	wait_timeout = now + (WAIT_TIMEOUT * 1000);
+	warn_timeout = now + (WARN_TIMEOUT * 1000);
 	service_pid = fork();
 	if (service_pid == -1)
 		eerrorx("%s: fork: %s", applet, strerror(errno));
@@ -432,45 +433,53 @@ svc_exec(const char *arg1, const char *arg2)
 	}
 
 	for (;;) {
-		if ((s = poll(fd, master_tty >= 0 ? 2 : 1, WAIT_INTERVAL / 1000000)) == -1) {
+		int timeout;
+		if (forever) {
+			timeout = -1;
+		} else if ((timeout = warn_timeout - tm_now()) < 0) {
+			timeout = 0;
+		}
+
+		if (poll(fd, master_tty >= 0 ? 2 : 1, timeout) == -1) {
 			if (errno != EINTR) {
 				eerror("%s: poll: %s", applet, strerror(errno));
 				ret = -1;
 				break;
 			}
+			continue;
 		}
 
-		if (s == 0 && !forever) {
-			timespecsub(&timeout, &interval, &timeout);
-			if (timeout.tv_sec <= 0) {
-				kill(service_pid, SIGKILL);
+		if (fd[1].revents & (POLLIN | POLLHUP)) {
+			bytes = read(master_tty, buffer, BUFSIZ);
+			write_prefix(buffer, bytes, &prefixed);
+		}
+
+		/* signal_pipe receives service_pid's exit status */
+		if (fd[0].revents & (POLLIN | POLLHUP)) {
+			if ((s = read(signal_pipe[0], &ret, sizeof(ret))) != sizeof(ret)) {
+				eerror("%s: receive failed: %s", applet, s < 0 ? strerror(errno) : "short read");
 				ret = -1;
 				break;
 			}
-			timespecsub(&warn, &interval, &warn);
-			if (warn.tv_sec <= 0) {
-				ewarn("%s: waiting for service (%d seconds)", applet, (int)timeout.tv_sec);
-				warn = (struct timespec) { .tv_sec = WARN_TIMEOUT };
-			}
-		} else if (s > 0) {
-			if (fd[1].revents & (POLLIN | POLLHUP)) {
-				bytes = read(master_tty, buffer, BUFSIZ);
-				write_prefix(buffer, bytes, &prefixed);
-			}
+			ret = WEXITSTATUS(ret);
+			if (ret != 0 && errno == ECHILD)
+				/* killall5 -9 could cause this */
+				ret = 0;
+			break;
+		}
 
-			/* signal_pipe receives service_pid's exit status */
-			if (fd[0].revents & (POLLIN | POLLHUP)) {
-				if ((s = read(signal_pipe[0], &ret, sizeof(ret))) != sizeof(ret)) {
-					eerror("%s: receive failed: %s", applet, s < 0 ? strerror(errno) : "short read");
-					ret = -1;
-					break;
-				}
-				ret = WEXITSTATUS(ret);
-				if (ret != 0 && errno == ECHILD)
-					/* killall5 -9 could cause this */
-					ret = 0;
-				break;
-			}
+		if (forever)
+			continue;
+
+		if ((now = tm_now()) >= wait_timeout) {
+			kill(service_pid, SIGKILL);
+			ret = -1;
+			break;
+		} else if (now >= warn_timeout) {
+			/* the timer might get off by a few ms, add 500ms so we get round numbers matching svc_wait. */
+			ewarn("%s: waiting for service (%"PRId64" seconds)", applet, (wait_timeout - now + 500) / 1000);
+			if ((warn_timeout += (WARN_TIMEOUT * 1000)) > wait_timeout)
+				warn_timeout = wait_timeout;
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


src/openrc-run/openrc-run.c
```diff
@@ -96,6 +93,7 @@ static int exclusive_fd = -1, master_tty = -1;
 static bool sighup, skip_mark, in_background, deps, dry_run;
 static pid_t service_pid;
 static int signal_pipe[2] = { -1, -1 };
+static bool timedout = false;
```
N-R-K, line 96: Needs to be `volatile`, and good idea to make it `sig_atomic_t` as well.

src/openrc-run/openrc-run.c
```diff
@@ -433,47 +434,56 @@ svc_exec(const char *arg1, const char *arg2)
 		fd[1].revents = 0;
 	}
 
+	if (!forever)
+		alarm(WARN_TIMEOUT);
```
N-R-K, line 438: I really don't like this. It's racy, the alarm might trigger before you enter the poll() call and then it will proceed to possibly sleep forever.

poll already accepts a timeout, so there's really no reason to do racy sigalarm stuff instead of just using the timeout argument.

N-R-K (changes requested):

src/openrc-run/openrc-run.c
navi-desu: good catch--

does that applies to the other two flags set by signals as well? `sighup` and `skip_mark`

src/openrc-run/openrc-run.c
navi-desu: i don't see how it'd pass 10 seconds before `poll`, which is the very next call, so i did chose this to avoid needing to query passed time and subtract things when something is actually read

but if it's considered better, we can change it:
initialize timeout to 10s
clock_gettime before poll, if == 0, reset timeout to 10s, warn, loop
if > 0, process the input, then clock_gettime, subtract the pre-poll time, subtract this difference from the timeout, if timeout > 0, loop, if not, warn and loop again

a bit messy, any better way to do this, or does this look good enough?

src/openrc-run/openrc-run.c
N-R-K: > does that applies to the other two flags set by signals as well? `sighup` and `skip_mark`

Very likely, but haven't looked.

I do see that there's a `signal_pipe`, I assume the rest of the code isn't fully structured around the poll event loop hence why we can't use the `signal_pipe` for everything else and need these flags (?).

navi-desu: the signal_pipe is only used to get the return status of the exec'd shell script, not the actual signal-- probably done so we could `poll` between the child-exit (aka the pipe) and the tty

`skip_mark` is used to skip marking the state if we receive SIGUSR1 (right now, from anywhere, but the idea is that a script can tell us to not mark the service started/stopped), `sighup` is used so that we only exit when the current command finishes executing

src/openrc-run/openrc-run.c
N-R-K: > any better way to do this

Use an absolute time as the target. In pseudo-code:

```c
	next_warn = now() + 10;
	for (;;) {
		r = poll(.., timeout: next_warn - now());
		if (r < 0) // ...

		if (fd[1].revents & ...) // process ...
		if (fd[0].revents & ...) // process ...
		
		if (now() >= next_warn) {
			if (now() >= kill_timer)
				// kill...
			else
				// warn...
			next_warn += 10;
		}
	}
```

Also, using a milisec instead of `struct timespec` will simplify a lot of things.
This is what I do in https://github.com/OpenRC/openrc/pull/811 with `clock_ms()`.

src/openrc-run/openrc-run.c
```diff
@@ -433,47 +434,56 @@ svc_exec(const char *arg1, const char *arg2)
 		fd[1].revents = 0;
 	}
```
N-R-K, line 435: Not directly related, but this branch can be removed entirely. `events` and `revents` are set above already. And for `fd` can we can just do `fd[1].fd = master_tty` always. The `master_tty >= 0` check inside poll will take care of ignoring it.

navi-desu: i noticed the same

also noticed the buffer we xmalloc to a constant BUFSIZ size, and made it a static array bc there's no reason to malloc there that i can see

src/openrc-run/openrc-run.c
```diff
@@ -423,61 +425,61 @@ svc_exec(const char *arg1, const char *arg2)
...
-			timespecsub(&warn, &interval, &warn);
-			if (warn.tv_sec <= 0) {
-				ewarn("%s: waiting for service (%d seconds)", applet, (int)timeout.tv_sec);
-				warn = (struct timespec) { .tv_sec = WARN_TIMEOUT };
-			}
-		} else if (s > 0) {
-			if (fd[1].revents & (POLLIN | POLLHUP)) {
-				bytes = read(master_tty, buffer, BUFSIZ);
-				write_prefix(buffer, bytes, &prefixed);
+		if (s == 0) {
```
navi-desu, line 451: can `s == 0` ever happen in the `forever` case (so -1 timeout)? if so we should guard this with `!forever && s == 0`, but i don't think it can

src/openrc-run/openrc-run.c
navi-desu: added a volatile to all signal-used bools

src/openrc-run/openrc-run.c
```diff
@@ -423,61 +425,61 @@ svc_exec(const char *arg1, const char *arg2)
...
 	for (;;) {
-		if ((s = poll(fd, master_tty >= 0 ? 2 : 1, WAIT_INTERVAL / 1000000)) == -1) {
+		int timeout;
+		if (forever) {
+			timeout = -1;
+		} else if ((timeout = next_warn - clock_ms()) < 0) {
+			timeout = 0;
+		}
+
+		if ((s = poll(fd, master_tty >= 0 ? 2 : 1, timeout > 0 ? timeout : 0)) == -1) {
```
N-R-K, line 442: Seems like it'd never do `-1` in the forever case?

src/shared/helpers.h
```diff
@@ -108,6 +109,13 @@ RC_UNUSED static void xclose_memstream(FILE *f)
 
 #undef ERRX
 
+RC_UNUSED static size_t clock_ms(void)
```
N-R-K, line 112: Use `int64_t`. We don't want to be stuck with 32 bit time on 32 bit system. Also, it's important to use signed so that subtraction work naturally.

src/openrc-run/openrc-run.c
N-R-K: I'd just ignore `s` entirely (aside from the negative return) and do another clock call to figure out where we're at. And there's no real reason trying to avoid it, it's cheap, very likely coming from vDSO.

src/openrc-run/openrc-run.c
```diff
@@ -423,61 +425,61 @@ svc_exec(const char *arg1, const char *arg2)
...
-			if (warn.tv_sec <= 0) {
-				ewarn("%s: waiting for service (%d seconds)", applet, (int)timeout.tv_sec);
-				warn = (struct timespec) { .tv_sec = WARN_TIMEOUT };
-			}
-		} else if (s > 0) {
-			if (fd[1].revents & (POLLIN | POLLHUP)) {
-				bytes = read(master_tty, buffer, BUFSIZ);
-				write_prefix(buffer, bytes, &prefixed);
+		if (s == 0) {
+			if ((kill_timeout -= WARN_TIMEOUT) > 0) {
```
N-R-K, line 452: Use an absolute target for this also. You can set it right before the `fork()` call. And you can also do `next_warn = min(kill_timeout, next_warn+10)` so that we don't exceed the kill timer.

N-R-K (changes requested):

src/openrc-run/openrc-run.c
navi-desu: i swear i had removed that. well, now it's gone

src/openrc-run/openrc-run.c
```diff
@@ -423,61 +426,61 @@ svc_exec(const char *arg1, const char *arg2)
...
-				if ((s = read(signal_pipe[0], &ret, sizeof(ret))) != sizeof(ret)) {
-					eerror("%s: receive failed: %s", applet, s < 0 ? strerror(errno) : "short read");
-					ret = -1;
-					break;
-				}
-				ret = WEXITSTATUS(ret);
-				if (ret != 0 && errno == ECHILD)
-					/* killall5 -9 could cause this */
-					ret = 0;
+			ewarn("%s: waiting for service ("PRId64" seconds)", applet, (wait_timeout - now) / 1000);
```
navi-desu, line 459:
```
../src/openrc-run/openrc-run.c:459:31: warning: too many arguments for format [-Wformat-extra-args]
  459 |                         ewarn("%s: waiting for service ("PRId64" seconds)", applet, (wait_timeout - now) / 1000);
```
i hate compilers sometimes

src/openrc-run/openrc-run.c
```diff
@@ -423,61 +426,62 @@ svc_exec(const char *arg1, const char *arg2)
...
+	fd[1].fd = master_tty;
 	fd[0].events = fd[1].events = POLLIN;
 	fd[0].revents = fd[1].revents = 0;
-	if (master_tty >= 0) {
-		fd[1].fd = master_tty;
-		fd[1].events = POLLIN;
-		fd[1].revents = 0;
-	}
 
+	warn_timeout = clock_ms() + (WARN_TIMEOUT * 1000);
```
N-R-K, line 434: Slight nitpick:

```c
now = clock_ms();
wait_timeout = now + (WAIT_TIMEOUT * 1000);
warn_timeout = now + (WARN_TIMEOUT * 1000);
... = fork();
```

Is more consistent than doing two different `clock_ms()`. Also has the added benefit of both of these being set near each other.

N-R-K (commented):
The poll part of the changes look much better now.

Will review and test the rest of the changes sometime this week.

src/openrc-run/openrc-run.c
```diff
@@ -508,61 +512,47 @@ svc_wait(const char *svc)
...
 	rc_stringlist_free(keywords);
 
 	xasprintf(&file, "%s/exclusive/%s", rc_svcdir(), basename_c(svc));
 
-	timeout.tv_sec = WAIT_TIMEOUT;
-	timeout.tv_nsec = 0;
-	warn.tv_sec = WARN_TIMEOUT;
-	warn.tv_nsec = 0;
+	if ((fd = open(file, O_RDONLY | O_NONBLOCK)) == -1) {
+		free(file);
```
N-R-K, line 528: This leaks I think. Suggestion:

```c
fd = open(file, ...);
free(file);
if (fd < 0) // ...
```

src/openrc-run/openrc-run.c
```diff
@@ -363,7 +363,7 @@ svc_exec(const char *arg1, const char *arg2)
 	int flags = 0;
 	struct pollfd fd[2];
 	int s;
-	char *buffer;
+	char buffer[BUFSIZ];
```
N-R-K, line 366: nit: Since this array doesn't have `static` linkage, I'd remove the word "static" from the commit message just to avoid confusion. Just "use an array instead of malloc" is fine.

N-R-K (changes requested):

src/openrc-run/openrc-run.c
navi-desu: can't wait to merge the dirfd patch so we have less of those awkward filepath allocations...

will fix in a bit

src/openrc-run/openrc-run.c
navi-desu: missing `%` actually, as always, the fault is the programmer

navi-desu:
okay so, the changes to the dependency code can have some racy edge cases as it was, so i'll drop them for now and work on them again at a later date

so for now, let's just get the refactors for proper waiting in

PR #831, opened 2025-04-07
openrc-run: rename ready -> notify

navi-desu:
readiness notification is experimental, and imo naming it 'ready' was a mistake, most other projects call them notify-fd, notify-socket, etc

systemd's protocol in particular, we plan to support a few more keys than just "READY=1" (namely "STATUS" to begin with), so 'notify' is a better variable name in my view


N-R-K:
Manpages need update too.

N-R-K (commented):
Shouldn't we change `init.d/user.in` also?

navi-desu:
i forgor, will do in a bit

Issue #800, opened 2025-03-16
[META] Avoid side-effects during option parsing


N-R-K:
Currently a lot of the tools perform side-effects during option parsing. Ideally, it should simply set the options and later code would act on it.

One major problem of the current approach is that `--user` flag behaves unreliably depending on the order. E.g `rc-status --user [...]` behaves one way and `rc-status [...] --user` behaves another.

Another issue is that some options will stop doing further processing which hinders the ability to report incorrect or excess arguments.

E.g a user may enter `rc-status --crashed default` in order to see all the crashed services in the `default` runlevel. `--crashed` actually just ignores any further arguments, even bogus ones like `rc-status --crashed --lol` will "work".

I've given some `rc-status` examples since that's what I was working on while I noticed this issue but the issue affects other tools too.

PR #832, opened 2025-04-09
supervise-daemon: accept units in respawn-{delay,period}

navi-desu:
currently the units are:
ms, miliseconds
s, sec, seconds
m, min, minutes
h, hours

didn't test it yet, will tomorrow, my brain is a pile of mush rn

### DIFF
--- a/src/supervise-daemon/supervise-daemon.c
+++ b/src/supervise-daemon/supervise-daemon.c
@@ -180,9 +180,9 @@ static int tty_fd = -1;
 #endif
 static pid_t child_pid;
 static int respawn_count = 0;
-static int respawn_delay = 0;
+static struct timespec respawn_period;
+static struct timespec respawn_delay;
 static int respawn_max = 10;
-static int respawn_period = 0;
 static char *fifopath = NULL;
 static int fifo_fd = 0;
 static char *pidfile = NULL;
@@ -624,9 +624,8 @@ RC_NORETURN static void supervisor(char *exec, char **argv)
 	sigset_t old_signals;
 	sigset_t signals;
 	struct sigaction sa;
-	struct timespec ts;
-	time_t respawn_now= 0;
-	time_t first_spawn= 0;
+	struct timespec respawn_now;
+	struct timespec first_spawn;
 
 	/* block all signals we do not handle */
 	sigfillset(&signals);
@@ -736,13 +735,13 @@ RC_NORETURN static void supervisor(char *exec, char **argv)
 			do_healthcheck = 0;
 			healthcheck_respawn = 0;
 			alarm(0);
-			respawn_now = time(NULL);
-			if (first_spawn == 0)
+			clock_gettime(CLOCK_MONOTONIC, &respawn_now);
+			if (!timespecisset(&first_spawn))
 				first_spawn = respawn_now;
-			if ((respawn_period > 0)
-					&& (respawn_now - first_spawn > respawn_period)) {
+			timespecsub(&respawn_now, &first_spawn, &respawn_now);
+			if (timespecisset(&respawn_period) && timespeccmp(&respawn_now, &respawn_period, >)) {
+				first_spawn = (struct timespec) {0};
 				respawn_count = 0;
-				first_spawn = 0;
 			} else
 				respawn_count++;
 			if (respawn_max > 0 && respawn_count > respawn_max) {
@@ -752,9 +751,10 @@ RC_NORETURN static void supervisor(char *exec, char **argv)
 				failing = 1;
 				continue;
 			}
-			ts.tv_sec = respawn_delay;
-			ts.tv_nsec = 0;
-			nanosleep(&ts, NULL);
+			do {
+				struct timespec rem = respawn_delay;
+				nanosleep(&rem, &rem);
+			} while (errno == EINTR);
 			if (exiting)
 				continue;
 			child_pid = fork();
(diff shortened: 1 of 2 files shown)


src/supervise-daemon/supervise-daemon.c
```diff
@@ -793,6 +793,51 @@ RC_NORETURN static void supervisor(char *exec, char **argv)
...
+			break;
+		}
+		if (!match) {
+			eerror("Invalid respawn unit %20s", ext);
+			return false;
+		}
+		/* fallthrough */
+	case 1:
+		if (unit == RESPAWN_MS)
+			spec->tv_nsec = value * unit;
```
N-R-K, line 829: You probably want to normalize this in case it reaches >= 1sec.

src/supervise-daemon/supervise-daemon.c
```diff
@@ -793,6 +793,51 @@ RC_NORETURN static void supervisor(char *exec, char **argv)
...
+	case 2:
+		for (size_t i = 0; exts[i].ext; i++) {
+			if (strncmp(exts[i].ext, ext, 20) != 0)
+				continue;
+			unit = exts[i].unit;
+			match = true;
+			break;
+		}
+		if (!match) {
+			eerror("Invalid respawn unit %20s", ext);
```
N-R-K, line 823: I'd remove this and let the the caller do the logging.

src/supervise-daemon/supervise-daemon.c
```diff
@@ -793,6 +793,51 @@ RC_NORETURN static void supervisor(char *exec, char **argv)
...
+		/* fallthrough */
+	case 1:
+		if (unit == RESPAWN_MS)
+			spec->tv_nsec = value * unit;
+		else
+			spec->tv_sec = value * unit;
+		break;
+	default:
+		return false;
+		eerrorx("Invalid respawn value '%s'", optarg);
```
N-R-K, line 835: Same here, remove.

(This one is also useless since it's after a `return`).

src/supervise-daemon/supervise-daemon.c
```diff
@@ -1134,7 +1177,13 @@ int main(int argc, char **argv)
 			parse_schedule(applet, NULL, sig);
 
 		str = rc_service_value_get(svcname, "respawn_delay");
-		sscanf(str, "%d", &respawn_delay);
+		if (!str || sscanf(str, "%zd:%zd", &respawn_delay.tv_sec, &respawn_delay.tv_nsec) != 2)
```
N-R-K, line 1180: %zd is for `size_t`. Neither tv_sec or tv_nsec is of that type so this could potentially do OOB write.

For `nsec` you'd need `%ld`. `time_t` doesn't have any printf specifier so you can't directly scan into it, you'd need to use some intermediate variable to scan into and then assign.

Also wouldn't it be better to store the value string directly and reparse with `parse_respawn`? It would make a new version of supervise-daemon be able to read the value written by an older version and be backwards compatible.

src/supervise-daemon/supervise-daemon.c
```diff
@@ -1209,13 +1260,13 @@ int main(int argc, char **argv)
 
 		rc_service_value_set(svcname, "pidfile", pidfile);
 		varbuf = NULL;
-		xasprintf(&varbuf, "%i", respawn_delay);
+		xasprintf(&varbuf, "%zd:%zd", respawn_delay.tv_sec, respawn_delay.tv_nsec);
```
N-R-K, line 1263: Same comment about wrong specifier. Would be better to store string directly, IMO.

N-R-K (changes requested):

src/supervise-daemon/supervise-daemon.c
```diff
@@ -793,6 +793,51 @@ RC_NORETURN static void supervisor(char *exec, char **argv)
...
+	struct { const char *ext; enum delay_unit unit; } exts[] = {
+		{ "ms", RESPAWN_MS }, { "miliseconds", RESPAWN_MS },
+		{ "s", RESPAWN_SECOND }, { "sec", RESPAWN_SECOND }, { "seconds", RESPAWN_SECOND },
+		{ "m", RESPAWN_MINUTE }, { "min", RESPAWN_MINUTE }, { "minutes", RESPAWN_MINUTE },
+		{ "h", RESPAWN_HOUR }, { "hour", RESPAWN_HOUR }, { NULL, 0 }
+	};
+
+	bool match = false;
+	char ext[20];
+	size_t value;
```
N-R-K, line 811: Also, `size_t` is for memory quantities. Using it for time is kinda gross.

navi-desu: i've actually never thought of size_t like that, but it does make some sense

so far i've used it as my default "i need to store a number" because it was presented to me as just "the native int size"

`time_t` is the appropriate time here i assume?

src/supervise-daemon/supervise-daemon.c
navi-desu: i thought of logging here since the caller wouldn't know why it failed, but tbh printing out the string it should be obvious as there's not a lot that can fail

src/supervise-daemon/supervise-daemon.c
N-R-K: > `time_t` is the appropriate type here i assume?

Yes, but also, you can't sscanf `time_t` so :man_shrugging:

I'd use `int` like the previous code or maybe `long` to be consistent with `tv_nsec`.

For bonus point, check for overflow before doing the multiplication.

navi-desu: i'll... do the overflow check tomorrow

navi-desu:
added a default value of 200ms for respawn_delay and 5 minutes for respawn_period (if i got the math right), unsure how much the period default should be

src/supervise-daemon/supervise-daemon.c
```diff
@@ -180,8 +180,8 @@ static int tty_fd = -1;
 #endif
 static pid_t child_pid;
 static int respawn_count = 0;
-static struct timespec respawn_period;
-static struct timespec respawn_delay;
+static struct timespec respawn_period = { .tv_sec = 5 * 60 };
+static struct timespec respawn_delay = { .tv_nsec = 200 * 1000 };
```
N-R-K, line 184: This change needs manpage update too. And also, this is 200 micro-seconds, not milliseconds like you (presumably) intended.

src/supervise-daemon/supervise-daemon.c
```diff
@@ -752,9 +751,10 @@ RC_NORETURN static void supervisor(char *exec, char **argv)
 				failing = 1;
 				continue;
 			}
-			ts.tv_sec = respawn_delay;
-			ts.tv_nsec = 0;
-			nanosleep(&ts, NULL);
+			do {
+				struct timespec rem = respawn_delay;
```
N-R-K, line 755: This just overwrites it always.

src/supervise-daemon/supervise-daemon.c
```diff
@@ -924,9 +972,9 @@ int main(int argc, char **argv)
 			break;
 
 		case 'D':  /* --respawn-delay time */
-			n = sscanf(optarg, "%d", &respawn_delay);
-			if (n	!= 1 || respawn_delay < 1)
+			if (!parse_respawn(&respawn_delay, optarg))
 				eerrorx("Invalid respawn-delay value '%s'", optarg);
+			rc_service_value_set(svcname, "respawn_delay", optarg);
```
N-R-K, line 977: This seems to be going in the opposite direction of https://github.com/OpenRC/openrc/issues/800

N-R-K (changes requested):

PR #835, opened 2025-04-10
sysfs: mount /sys/fs/bpf if available

lotheac:
eg. cilium expects /sys/fs/bpf to be mounted. systemd mounts it here as well.


thesamesam (approved):
Seems reasonable, thanks.

https://docs.ebpf.io/linux/concepts/pinning/ mentions this too:
> If your Linux distribution does not automatically mount the BPF file system you can do so manually by executing mount -t bpf bpffs /sys/fs/bpf as root or making it part of a setup/initialization script.

navi-desu:
thanks!

PR #837, opened 2025-04-12
add basic unit tests

navi-desu:


src/librc/librc.c
```diff
@@ -587,6 +589,9 @@ free_rc_dirs(void)
 	free(rc_dirs.runleveldir);
 	rc_dirs.runleveldir = NULL;
 	free(rc_dirs.svcdir);
+	for (size_t i = 0; i < ARRAY_SIZE(scriptdirs); i++)
+		if (rc_dirs.scriptdirs_data[i])
+			free(rc_dirs.scriptdirs_data[i]);
```
N-R-K, line 594: Check is pretty useless since `free(NULL)` is no-op. I'd also set it back to NULL after `free()`, that's what the rest is doing.

src/librc/librc.c
```diff
@@ -599,6 +604,24 @@ rc_is_user(void)
 	return is_user;
 }
 
+void
+rc_set_root(const char *root)
+{
+	int len = strlen(root);
+	if (root[len - 1] == '/')
+		len--;
```
N-R-K, line 612: Doesn't work for "/", besides, it's unnecessary anyways isn't it? Multiple slashes in path are redundant but don't cause any issue. But if you want to strip trailing slashes, then here's the proper way to do it (note `len>1` rather then `len>0`):

```c
while (len > 1 && root[len-1] == '/') len--;
```

src/openrc-run/openrc-run.c
```diff
@@ -63,14 +63,15 @@ static const struct timespec interval = { .tv_nsec = WAIT_INTERVAL };
 
 const char *applet = NULL;
 const char *extraopts = "stop | start | restart | status | describe | zap";
-const char getoptstring[] = "dDsSvl:Z" getoptstring_COMMON;
+const char getoptstring[] = "dDsSvl:Z" getoptstring_DIRS getoptstring_COMMON;
```
N-R-K, line 66: Since `--root` is only for testing purposes, shouldn't we try to use something less invasive?

N-R-K (changes requested):
I like tests. I don't like how testing specific hacks (`--root`, `$RC_ROOT`) are bleeding into regular builds though.

src/openrc-run/openrc-run.c
navi-desu: it is *not* just for testing purposes, another use of it is to setup a chroot, image, etc, which is a feature i really missed while setting up my laptop and trying to build gentoo for arm

src/librc/librc.c
navi-desu: it's unnecessary though i just find `foo//nya` ugly, and that would propagate to the symlinks created...

and i now i wonder if the symlinks (rc-update add, del) should be relative, since trying to use --root to build a chroot would give broken symlinks as is right now

src/openrc-run/openrc-run.c
thesamesam: Exactly.

N-R-K: Okay, that makes sense. But in that case, the commit messages should document it more clearly. Since the PR title was about adding tests, I assumed this would only be used for testing.

navi-desu: i realized `--root` doesn't make much sense for rc-service (is there any valid case to start a service from a chroot... from outside it?)

maybe we keep `RC_ROOT` internally, but only expose `--root` for `rc-update`?

thesamesam: I can think of some contrived cases for `rc-status --root` but they don't make much sense or are very unlikely, e.g. some `/run` inside the root that isn't a tmpfs.

PR #841, opened 2025-04-16
unblock signal handlers during signal_setup

navi-desu:
openrc-run can be called from places that don't reset the signal mask
before exec, example being the pam close_session session hook called
from runuser, which leads to svc_exec hanging, as SIGCHLD is never
delivered.

Bug: https://bugs.gentoo.org/953748

### DIFF
--- a/src/shared/misc.c
+++ b/src/shared/misc.c
@@ -270,18 +270,6 @@ signal_setup(int sig, void (*handler)(int))
 }
 
 int
-signal_setup_restart(int sig, void (*handler)(int))
-{
-	struct sigaction sa;
-
-	memset(&sa, 0, sizeof (sa));
-	sigemptyset(&sa.sa_mask);
-	sa.sa_handler = handler;
-	sa.sa_flags = SA_RESTART;
-	return sigaction(sig, &sa, NULL);
-}
-
-int
 svc_lock(const char *applet, bool ignore_lock_failure)
 {
 	int fd = openat(rc_dirfd(RC_DIR_EXCLUSIVE), applet, O_WRONLY | O_CREAT | O_NONBLOCK, 0664);
--- a/src/shared/misc.c
+++ b/src/shared/misc.c
@@ -340,8 +340,11 @@ exec_service(const char *service, const char *arg)
 		sigaction(SIGUSR1, &sa, NULL);
 		sigaction(SIGWINCH, &sa, NULL);
 
-		/* Unmask signals */
-		sigprocmask(SIG_SETMASK, &old, NULL);
+		/* Unmask all signals.
+		 * We might've been called from pam_openrc by
+		 * a process that masked signals we rely on.
+		 * Bug: https://bugs.gentoo.org/953748 */
+		sigprocmask(SIG_UNBLOCK, &full, NULL);
 
 		/* Safe to run now */
 		execl(file, file, "--lockfd", sfd, arg, (char *) NULL);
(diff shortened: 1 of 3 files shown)


navi-desu:
alternatively we can/should unblock signals in service_start()/service_stop() from pam_openrc

N-R-K:
> alternatively we can/should unblock signals in service_start()/service_stop() from pam_openrc

Yes. If the caller is blocking signals it should be the one responsible for unblocking it in the child.

N-R-K:
Okay, so looks like the signal was being masked by `runuser` and not openrc itself.

But even then, unblocking signal in a generic looking function is unexpected, there are legitimate use-cases of blocking signals while setting up the sighandler to avoid races. So I still stand by that this should be done in `pam_openrc` with a comment explaining that it's a workaround for an unusual situation.

Also, shouldn't we report this to util-linux in case this is a bug?

navi-desu:
moved the unmasking to shared/misc.c:exec_service, which is called by pam_openrc when it's time to exec the service

navi-desu:
> But even then, unblocking signal in a generic looking function is unexpected, there are legitimate use-cases of blocking signals while setting up the sighandler to avoid races.

there's also the solution of unblocking after we set our signal handlers as well, versus unblocking it all in `exec_service`

src/shared/misc.c
```diff
@@ -343,17 +331,11 @@ exec_service(const char *service, const char *arg)
...
 
 	if ((pid = fork()) == 0) {
-		/* Restore default handlers */
-		sigaction(SIGCHLD, &sa, NULL);
-		sigaction(SIGHUP, &sa, NULL);
-		sigaction(SIGINT, &sa, NULL);
-		sigaction(SIGQUIT, &sa, NULL);
-		sigaction(SIGTERM, &sa, NULL);
-		sigaction(SIGUSR1, &sa, NULL);
-		sigaction(SIGWINCH, &sa, NULL);
```
N-R-K, line 353: If a signal was ignored by the parent, it's unspecified if they remain ignored or gets reset after exec. In linux, they remain ignored, so you still need to restore the handlers back to default explicitly in those cases.

N-R-K (changes requested):

src/shared/misc.c
N-R-K: > openrc-run unconditionally sets those to `handle_signal` anyway so we shouldn't need this?

This is in `shared/misc.c` though. Is openrc-run the only other place (except `openrc_pam`) that uses this?

navi-desu: only `pam_openrc.c`, `openrc-run.c` and `openrc/rc.c` call that function

but i meant the opposite, that function always execs into a service, so, into `openrc-run`, and openrc-run will always set up their handlers for those specific signals before relying on them, at the setup step, so, it shouldn't matter if we "unset" SIG_IGN back to SIG_DFL here, i assume

N-R-K: Ah right. In that case, it's fine to remove them. Though I'd leave a comment.

Also, maybe we can move the signal unblocking to `openrc-run` too then?

navi-desu: > Also, maybe we can move the signal unblocking to openrc-run too then? Unsure, feels kinda hacky either ways.

that was, kindof the original patch, unblocking signals at the same time we install handlers for them, either way solves the bug, but i do not really know which is better.

navi-desu: this as is fixes the issue at hand and we should release soon, so, we can improve the signal handling of openrc as a whole later, merging for now

PR #843, opened 2025-04-17
shared/misc: required `ready` tag for socket notify method

navi-desu:
with this we can define `notify=socket` as an overall systemd "notify-socket" compatibility flag. in the future, `nofiy=socket` will simply enable access to the basic STATUS= command, `socket:ready` enables READY=1, `socket:fdstore` would enable fd-store compat, `socket:watchdog:<timeout>` for the watchdog and so on.

the format of the variable would end up being:
`notify="protocol[:option[,option ...]]"`

### DIFF
--- a/src/shared/misc.c
+++ b/src/shared/misc.c
@@ -548,13 +548,20 @@ struct notify notify_parse(const char *applet, const char *notify_string)
 		notify.type = NOTIFY_FD;
 		if (pipe(notify.pipe) == -1)
 			eerrorx("%s: pipe: %s", applet, strerror(errno));
-	} else if (strcmp(notify_string, "socket") == 0) {
+	} else if (strncmp(notify_string, "socket", sizeof("socket") - 1) == 0) {
 		union {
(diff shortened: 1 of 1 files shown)


src/shared/misc.c
```diff
@@ -553,8 +553,14 @@ struct notify notify_parse(const char *applet, const char *notify_string)
 			struct sockaddr header;
 			struct sockaddr_un unix;
 		} addr = { .unix = { .sun_family = AF_UNIX } };
-		int written = snprintf(addr.unix.sun_path, sizeof(addr.unix.sun_path), "%s/supervise-%s.sock", rc_svcdir(), applet);
-
+		int written;
+		const char *opts = strchr(notify_string, ':');
```
N-R-K, line 557: This is inside a `strcmp(notify_string, "socket") == 0` branch, so none of this will ever be triggered.

N-R-K (changes requested):

PR #845, opened 2025-04-24
remove the meson_runlevels.sh script

williamh:
The install_symlink function now available in meson removes the need for this installation script. This pull request reworks the meson build files to use this functionality.


runlevels/meson.build
```diff
@@ -0,0 +1,63 @@
...
+boot_runlevel_dir = runleveldir / 'BOOT'
+default_runlevel_dir = runleveldir / 'default'
+nonetwork_runlevel_dir = runleveldir / 'nonetwork'
+shutdown_runlevel_dir = runleveldir / 'shutdown'
+
+sysinit = []
+if os == 'Linux'
+  sysinit += 'cgroups devfs dmesg sysfs'.split()
+endif
+boot = 'bootmisc fsck hostname localmount loopback root swap sysctl'.split()
```
navi-desu, line 12: instead of calling `split()` in a string, we can add an array iirc, `boot += [ 'bootmisc', 'fsck', ...]`

runlevels/meson.build
```diff
@@ -0,0 +1,63 @@
...
+  boot += 'devdb swap-blk tys wscons'.split()
+endif
+default = 'local netmount'.split()
+nonetwork=['local']
+shutdown = ['savecache']
+if os == 'Linux'
+  shutdown += 'killprocs mount-ro'.split()
+endif
+
+foreach name: sysinit
```
navi-desu, line 35: maybe we could use a dict? would be less repetitive:
```meson
runlevels = { 'boot': [ 'svc1', 'svc2' ] }
foreach name, services : runlevels
    foreach service : services
        install_symlink(service, 
            install_dir: runleveldir / name,
            pointing_to: initd / service)
    endforeach
endforeach

File: meson.build [hunk 3]
```diff
@@ -213,6 +213,7 @@ subdir('etc')
 subdir('init.d')
 subdir('local.d')
 subdir('man')
+subdir('runlevels')
```
navi-desu, line 216: do we need a whole empty dir for this? considering we're symlinking init scripts, having the install logic in `init.d/meson.build` would make sense to me

navi-desu (changes requested):

runlevels/meson.build
williamh: This is resolved in the latest update.

runlevels/meson.build
williamh: I was able to do this, but I had to construct the dictionary after all of the service arrays are set up properly since dictionaries are immutable.

meson.build
williamh: We aren't creating any symlinks in the init.d directory, so I would rather have a separate runlevels directory.

N-R-K: > We aren't creating any symlinks in the init.d directory

`init.d/meson.build` creates `agetty` symlink since https://github.com/OpenRC/openrc/commit/e56171b2a1815560e03b9e9b2ef507a02c59e288.

Or did you mean the symlinks aren't going into `/etc/init.d` hence the new directory?

navi-desu: note that agetty symlinks to the runlevel dir as well, so if we stick to the new dir, we'd probably move the agetty links to this dir later asw

williamh: > note that agetty symlinks to the runlevel dir as well, so if we stick to the new dir, we'd probably move the agetty links to this dir later

Yes, the runlevels meson.build file only creates symlinks in the runlevels directory. The agetty code in the init.d meson.build does both, and I have another branch with fixes for that. I will open it as another pr once this one is merged.

N-R-K (approved):

N-R-K:
Also, there was a bit of problem with `install_symlink` (https://github.com/OpenRC/openrc/issues/847). Does that issue apply here or is this PR safe?

navi-desu:
the issue was about a symlink pointing to a broken symlink in destdir, the symlinks here should be safe since they all point to physical files, though i still have to test to be sure

PR #846, opened 2025-04-25
man: use Ic macro for rc-update commands

WhyNotHugo:
The Ar macro is used for arguments. The Ic macro is used for internal commands. The latter is suitable for subcommands of rc-update.

When rendered via man, this makes commands and variable arguments visually distinct (bold and underlined respectively by default).


navi-desu:
thanks!

PR #848, opened 2025-04-25
improve respawn handling with notify set

navi-desu:


src/shared/misc.c
```diff
@@ -585,6 +585,9 @@ bool notify_wait(const char *applet, struct notify notify)
 	for (;;) {
 		char buf[BUFSIZ];
 		ssize_t bytes = read(notify.fd, buf, BUFSIZ);
+
+		if (bytes == 0)
+			return false;
```
N-R-K, line 590: The -1 case prints error message, this doesn't. A bit inconsistent. Other than this, looks ok.

N-R-K (approved):

src/shared/misc.c
navi-desu: it should print failed to start up on the chain, as the supervisor would exit with EXIT_FAILURE

since EOF would be an usual failure case, and not an failure of the read syscall, i think it fits?

thesamesam:
Add `Bug:` tags to both commits?

navi-desu:
someone else hit a similar issue and this fixed it, i'll merge it already

Issue #847, opened 2025-04-25
Unable to install 0.62 when using -Dsysvinit=true option


Dudemanguy:
If you attempt to build and install openrc using `-Dsysvinit=true` like so,

```
meson setup build -Dprefix=/usr -Dsysvinit=true
meson compile -C build
```
when trying to install, you might get the following error.

```
meson install -C build --dry-run
ERROR: Tried to install symlink to missing file /etc/init.d/agetty
```

The culprit appears to be the `install_symlink` calls introduced in e56171b2a1815560e03b9e9b2ef507a02c59e288. I think this is actually just an upstream meson limitation though. Appears to be exactly [this issue](https://github.com/mesonbuild/meson/issues/12253). I guess some workaround for now is needed.

navi-desu:
the `agetty` target is configured and installed right above, so it should exist, i can't reproduce this with the commands as listed, as i get:

```
Installing /home/anna/src/openrc/openrc/install/init.d/agetty to /etc/init.d
```
then later
```
Installing symlink pointing to /etc/init.d/agetty to /etc/init.d/agetty.tty1
```

could you provide `install-log.txt` and `meson-log.txt` from `builddir/meson-logs`?

Dudemanguy:
I think it's a meson thing because the symlink already exists on your system but not on mine since I don't actually have openrc installed. So you don't hit that meson limitation. It can pop up while packaging for example since you install to some destdir.

Here's the two logs.
[install-log.txt](https://github.com/user-attachments/files/19911917/install-log.txt)
[meson-log.txt](https://github.com/user-attachments/files/19911928/meson-log.txt)

eli-schwartz:
https://github.com/mesonbuild/meson/pull/12266

My analysis:

> As for openrc:
> 
> ```
> Installing symlink pointing to /etc/init.d/agetty to /home/eschwartz/git/openrc/builddir/pkg/etc/init.d/agetty.tty1
> 
> ERROR: Tried to install symlink to missing file /home/eschwartz/git/openrc/builddir/pkg/etc/init.d/agetty.tty1
> ```
> 
> and:
> 
> ```
> $ ls -l builddir/pkg/etc/init.d/
> total 164
> -rwxr-xr-x 1 eschwartz eschwartz 1242 Apr 25 12:10 agetty*
> lrwxrwxrwx 1 eschwartz eschwartz   18 Apr 25 12:11 agetty.tty1 -> /etc/init.d/agetty
> ```
> 
[...]

PR #849, opened 2025-04-25
init.d/meson.build: install agetty as a relative symlink

navi-desu:
it's the standard for symlinks to exist as relative pointers.

Fixes: https://github.com/OpenRC/openrc/issues/847


eli-schwartz (approved):

N-R-K (approved):
Maybe also add that it fixes an issue with meson install in the commit message. Easier to get context while `git blame`ing instead of having to open bug links.

Dudemanguy:
Works. Thanks.

PR #850, opened 2025-04-25
conf.d/install agetty conf files via meson

Dudemanguy:
Since e56171b2a1815560e03b9e9b2ef507a02c59e288, the agetty symlinks are installed via meson option. There's no reason to not do the same for the conf.d files as well.

### DIFF
--- a/conf.d/meson.build
+++ b/conf.d/meson.build
@@ -60,3 +60,11 @@ elif os == 'NetBSD'
 endif
 
 install_data(conf_data, install_dir : conf_d_dir)
+
+if get_option('sysvinit') == true and os == 'Linux'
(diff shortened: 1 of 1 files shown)


navi-desu:
those confs don't exactly need to exist though, `conf.d/agetty` is loaded, so every service will use it, and `agetty.$tty` is only needed if you want a different config

for any `base.instance`, `conf.d/base` is loaded, then `conf.d/base.instance` is

navi-desu:
> Ah I actually didn't know it worked like that.

i really ought to improve openrc docs, specially around multiplexing (what we call the script symlinks)

i'll close this since i believe that just `agetty` is enough and less verbose, but feel free to reopen if you disagree

PR #860, opened 2025-05-05
start-stop-daemon: add longopts_help value for --notify

OddBloke:
Without this, the help output from `--notify` onwards is incorrectly off-by-one, e.g.:

```
      --notify <arg>                Display this help output
  -h, --help                        Disable color output
  -C, --nocolor                     Display software version
```

By aligning its length with that of `longopts`, this also fixes a segfault we've observed in our arm64 builds.

### DIFF
--- a/src/start-stop-daemon/start-stop-daemon.c
+++ b/src/start-stop-daemon/start-stop-daemon.c
@@ -171,6 +171,7 @@ const char * const longopts_help[] = {
 	"Print dots each second while waiting",
 	"Set process scheduler",
 	"Set process scheduler priority",
+	"Configures experimental notification behaviour",
 	longopts_help_COMMON
 };
 const char *usagestring = NULL;


OddBloke:
(I wasn't 100% sure on how to distill the manpage's description of the option into a short sentence, if there's preferred verbiage I'll happily switch to it!)

N-R-K (approved):

N-R-K:
It's very easy to forget adding the help text. I've made the same mistake before (thankfully in my case I caught it locally before it made it into master).

Ideally we should have static assert or similar in place so that mistakes like this fails to complie to begin with.

navi-desu:
> Ideally we should have static assert or similar in place so that mistakes like this fails to complie to begin with.

i have a PoC branch that ties all 3 values (short, long, help) in a single macro call, i'll try to submit it later today

Issue #859, opened 2025-05-05
openrc 0.62 breaks rc_parallel="YES"


bstaletic:
I had to revert to `rc_parallel="NO"` on my system after gentoo's upgrade to openrc 0.62.

The output of sysinit with `rc_parallel="YES"`:

```
 * Mounting /proc ... [ ok ]
 * Mounting /run ... [ ok ]
 * /run/openrc: creating directory
 * /run/lock: creating directory
 * /run/lock: correcting owner
udev                      | * ERROR: cannot start udev as sysfs would not start
udev-trigger              | * ERROR: cannot start udev-trigger as sysfs would not start
udev                      | * ERROR: cannot start udev as sysfs would not start
 * Create Static Devices Nodes in /dev ... * Mounting /sys ... * Mounting devtmpfs on /dev ... [ ok ]
 [ ok ]
 * Mounting security filesystem ... [ ok ]
 * Mounting config filesystem ... * Mounting /dev/mqueue ... [ ok ]
 [ ok ]
 * Mounting bpf filesystem ... [ ok ]
 * Mounting /dev/pts ... [ ok ]
 [ ok ]
 * Mounting /dev/shm ... [ ok ]

   OpenRC 0.62.2 is starting up Gentoo Linux (x86_64)

 * Caching service dependencies ... [ ok ]
```

The status of sysinit services after the error:

```
Runlevel: sysinit
 sysfs                                                             [  started  ]
 devfs                                                             [  started  ]
 udev                                                              [  stopped  ]
 systemd-tmpfiles-setup-dev                                        [  started  ]
 cgroups                                                           [  started  ]
 dmesg                                                             [  started  ]
 udev-trigger                                                      [  stopped  ]
```

I am not sure what else would be useful to post. Feel free to ask for more input.

thesamesam:
Please confirm if the issue is there with 0.62.2.

navi-desu:
seems to be some sort of race condition, since sysfs does start on the latter list

i'll work on this

bstaletic:
> Please confirm if the issue is there with 0.62.2

As you have noticed, I did. I too was hopeful it would resolve the problem.

> seems to be some sort of race condition, since sysfs does start on the latter list
> 
> i'll work on this

Whatever you need me to test, just ping me. I'm not afraid to mess with my system's boot process. :)

N-R-K:
> Whatever you need me to test, just ping me. I'm not afraid to mess with my system's boot process. :)

[...]

PR #879, opened 2025-06-11
librc: don't create /run/openrc itself on demand.

navi-desu:
this is a temporary fix, in a system that boots with / as rw, /run unmounted, and has rc_parallel on, openrc would create a dirfd to /run/openrc before mounting /run, thus keeping an open fd to the old /run mountpoint, and creating 'exclusive' lockfiles under it, that subsequent services can't see nor lock on.

Fixes: https://github.com/OpenRC/openrc/issues/859
Fixes: 07d6f4d5

### DIFF
--- a/src/librc/librc-depend.c
+++ b/src/librc/librc-depend.c
@@ -693,6 +693,8 @@ rc_deptree_update_needed(time_t *newest, char *file)
 
 	/* Quick test to see if anything we use has changed and we have
 	 * data in our deptree. */
+	if (mkdir(rc_svcdir(), 0755) != 0 && errno != EEXIST)
+		fprintf(stderr, "mkdir '%s': %s\n", rc_svcdir(), strerror(errno));
 
 	if (fstatat(rc_dirfd(RC_DIR_SVCDIR), "deptree", &buf, 0) == 0) {
 		mtime = buf.st_mtime;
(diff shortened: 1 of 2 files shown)


N-R-K:
I don't think I fully follow what's going on. Is it:

1. openrc creates stuff under /run/openrc but /run isn't mounted yet.
2. /run gets mounted and the older stuff can no longer be seen.

Or:

1. openrc *opens* `/run` but doesn't create anything.
2. /run is mounted
3. Now it creates stuff, but the dirfd is pointing to the old `/run`.

If it's the former then I don't think this fixes it, if it's the latter then this patch makes sense. But why call it "temporary fix"?

navi-desu:
> I don't think I fully follow what's going on. Is it:
> 
>     1. openrc creates stuff under /run/openrc but /run isn't mounted yet.
> 
>     2. /run gets mounted and the older stuff can no longer be seen.
> 
> 
> Or:
> 
>     1. openrc _opens_ `/run` but doesn't create anything.
> 
>     2. /run is mounted
> 
>     3. Now it creates stuff, but the dirfd is pointing to the old `/run`.
> 
> 
> If it's the former then I don't think this fixes it, if it's the latter then this patch makes sense. But why call it "temporary fix"?

1. openrc creates /run/openrc *and* opens a dirfd to it
2. /run gets mounted, so /run/openrc can no longer be seen by *new* processes
3. then openrc creates /run/openrc/exclusive under the dirfd for the old /run
4. new processes try to check for exclusive in the new /run, find nothing

so it's both, at the same time, and this fixes the former case because it (now, i had forgotten to) restores old behaviour of only creating /run/openrc when rc_deptree_update_needed is called, which is always after `do_sysinit`

this is also why it's temporary, in my view, i don't like this "solution" of creating the runtime dir as a side-effect of this one "check" function

PR #881, opened 2025-06-12
Clarify that cgroups v2 is the default mount mode for cgroup

zyxhere:
Change rc_cgroup_mode="unified" to "hybrid" as unified is reduntant


N-R-K (changes requested):
The commented out values themselves document the default settings.

navi-desu:
commented out values are meant to show the default value, it's not redundant since it's (kindof) documentation

also support for legacy/hybrid cgroups will probably be removed in #820 anyway

thanks for the pr anyway though

PR #884, opened 2025-06-17
*-daemon: actually filter RC_/SSD_ variables

navi-desu:
oops, inverted check

pull request of shame this one

Fixes: 7208da2


sertonix:
Would it be possible to also clean EINFO_* and SVCNAME?

(Some einfo would need to be saved somehow for later)

Issue #896, opened 2025-07-04
forever default value change leads to service stop failures (Breaking Change)


josch09:
Hi,

I have some init.d services which need more than 60 seconds (=WAIT_TIMEOUT in openrc-run.c) to stop.
And they worked fine until April and still work fine with Gentoo's 0.56 version)

But now they all fail during stop because of the active WAIT_TIMEOUT in openrc-run.c.

Easy to reproduce when I create a stop function for my /etc/init.d/douchebag  service file

stop() {
  ebegin 'wait for the WAIT_TIMEOUT'
  sleep 65
  eend $?
}

######

the output:

 * wait for the WAIT_TIMEOUT ...
 * douchebag: waiting for service (50 seconds)
 * douchebag: waiting for service (40 seconds)
 * douchebag: waiting for service (30 seconds)
 * douchebag: waiting for service (20 seconds)
 * douchebag: waiting for service (10 seconds)
 * ERROR: douchebag failed to stop

#######

After looking into the source code I was able to fix this for my own services by adding the keyword notimeout.

But this BREAKING CHANGE (from infinity to 60 seconds)  should at least be mentioned...

- E.g. ending up with a corrupt postgresql database after a reboot is not a nice thing only because of friendly fire coming from the openrc init system.
- Are the various database maintainers in the Gentoo portage tree already aware of this change ?

navi-desu:
0.62.6 reverts this change.

i'll add it back later, just, properly and with more configuration knobs that aren't just "on" and "off"

PR #897, opened 2025-07-04
openrc-run: remove timeout from start/stop

navi-desu:
4d8a5c4 added the timeout logic from svc_wait into svc_exec, with the goal of avoiding stalling the boot process should a service hang, however we failed to mention the change in NEWS, and some critical services like postgres rely on inifinite timeout for `stop()`.

this logic was also sharing the `timeout` keyword from `depend()`, which is not very clean

let's revert the timeout logic, and properly re-introduce it later, with effort put into making sure no issues happen.

Bug: https://github.com/OpenRC/openrc/issues/896

### DIFF
--- a/src/openrc-run/openrc-run.c
+++ b/src/openrc-run/openrc-run.c
@@ -364,15 +364,6 @@ svc_exec(const char *arg1, const char *arg2)
 	int slave_tty;
 	sigset_t sigchldmask;
 	sigset_t oldmask;
-	int64_t wait_timeout, warn_timeout, now;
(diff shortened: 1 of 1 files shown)


martinetd:
Thanks for the heads up on this PR in the other issue.

I think we're good but I've restarted tests with 0.62.5 + #898 (sigchld block) + this PR (as rebased in https://github.com/martinetd/openrc/commits/no-timeout-start-stop/ - not a difficult conflict fix anyway)
I'll report back if anything else comes up :)

Cheers,

PR #898, opened 2025-07-05
openrc-run: block SIGCHLD during fork() to avoid race with reaping

### LINKED ISSUE #895: child sometimes not reaped properly in openrc-run svc_exec(), signal_pipe never written to
Hi,

I've been chasing a very rare hang on boot on openrc 0.55.1 (alpine 3.21), let me start describing what happens there before updating world view to master branch.

###  on 0.55.1

Openrc start is stuck in poll() indefinitely in `svc_exec()` (after service started, before the starting link was moved to started); writing into the "writer" hand of the pipe gets it unstuck

At this point boot is stuck: openrc default never finishes, so init never starts getty or whatever it's supposed to do next.

```
armadillo:~# pstree -p
init(1)-+-avahi-daemon(3210)---avahi-daemon(3211)
        |-buttond(2404)
        |-chronyd(3364)
        |-connection-reco(2326)---sleep(7353)
        |-dbus-daemon(1414)
        |-dnsmasq(1544)
        |-klogd(1467)
        |-logger(1478)
        |-openrc(1551)-+-dummy00(1634)
        |              `-openrc(1553)
        |-sshd(3459)---sshd-session(3470)---sshd-session(3472)---sh(3473)---pstree(7535)
        |-supervise-daemo(1512)---NetworkManager(1513)-+-{gdbus}(1557)
        |                                              |-{gmain}(1555)
        |                                              `-{pool-spawner}(1556)
        |-supervise-daemo(2740)---armadillo-twin-(2744)---{AwsEventLoop 1}(3223)
        |-syslogd(1400)
        |-udevd(923)
        |-wpa_supplicant(2028)
        `-wwan-led(3292)---sleep(7533)
armadillo:~# ps auxwwf | grep [1]551
 1551 root      0:00 /sbin/openrc default
armadillo:~# ps auxwwf | grep [1]553
 1553 root      0:00 /sbin/openrc default
armadillo:~# ps auxwwf | grep [1]634
 1634 root      0:00 {dummy00} /sbin/openrc-run /etc/init.d/dummy00 --lockfd 13 start

armadillo:~# cat /proc/1634/stack
[<0>] __switch_to+0xa8/0xe0
[<0>] do_sys_poll+0x3d4/0x5dc
[<0>] __arm64_sys_ppoll+0xac/0x138
[<0>] el0_svc_common.constprop.0+0x78/0x1c0
[<0>] do_el0_svc+0x1c/0x28
[<0>] el0_svc+0x10/0x1c
[<0>] el0_sync_handler+0x108/0x120
[<0>] el0_sync+0x180/0x1c0
armadillo:~# strace -f -p 1634
strace: Process 1634 attached
ppoll([{fd=3, events=POLLIN}, {fd=5, events=POLLIN}], 2, NULL, NULL, 8) = 1 ([{fd=3, revents=POLLIN}])
<hung here>


armadillo:~# cd /proc/1634/fd
armadillo:/proc/1634/fd# ls -l
total 0
lrwx------    1 root     root            64 Jul  4 09:25 0 -> /dev/console
lrwx------    1 root     root            64 Jul  4 09:25 1 -> /dev/pts/0
l-wx------    1 root     root            64 Jul  4 09:25 13 -> /run/openrc/exclusive/dummy00
lrwx------    1 root     root            64 Jul  4 09:25 2 -> /dev/pts/0
lr-x------    1 root     root            64 Jul  4 09:25 3 -> pipe:[1635]
l-wx------    1 root     root            64 Jul  4 09:25 4 -> pipe:[1635]
lrwx------    1 root     root            64 Jul  4 09:25 5 -> /dev/ptmx
lrwx------    1 root     root            64 Jul  4 09:25 6 -> /dev/pts/7
armadillo:/proc/1634/fd# echo 0 > 4
## ^ I realize I should not have written it as a char but as a native int
## nevertheless, everything got unstuck
## also, note that there was no zombie child process for openrc-run so it was already reaped
```

### on master

https://github.cohttps://github.com/OpenRC/openrc/commit/4d8a5c47f49c775f00c508b32bb4d9a8bca9c07e introduced a timeout, so this no longer hangs indefinitely! yay!
Nevertheless, I don't see why it still couldn't happen, and we'll wait 60s (or whatever timeout) for no reason, so I think it's worth fixing this signal handling as well.

I haven't reproduced on a recent openrc yet but I'm running the same loop on alpine 3.22 (openrc 0.62.4) over the weekend, and will confirm it works as I wrote above: we get logs service timed out but boot finishes.

### What's wrong? What next?

First let me state it's very hard to reproduce, I've never seen it on other faster/slower boards, and it's only been happening at one of our customers for a reason I cannot understand, and it no longer happens when they added extra traces (like copying the /run/openrc directory in a service) which should have no impact on this...
I could finally reproduce it locally after adding a bunch of dummy services (empty start() function or just logging ebegin/eend) at boot, after ~500 reboots...

What I *think* is wrong is that, sigchld processing expects the `service_pid` to be known, but I don't think there's any guarantee the parent process will be scheduled and return from fork() before the signal handler runs if the child exits fast enough?
e.g.

- openrc-run calls fork()
- child process starts, child process terminates
- parent receives SIGCHLD and runs the handler with service_pid == -1, reaps the child
- parent continues runing from `service_pid = fork()` and the variable is set

I'm not quite sure how we're supposed to synchronize such a variable with a signal handler, but what we can do is change the meaning of the signal_pipe: if it's present when handler runs just write to it whenever sigchld happens and let the main process deal with wait()ing
e.g., something like this overcompilcated patch (I've added a few pedantic comments for opinion)
```diff
diff --git a/src/openrc-run/openrc-run.c b/src/openrc-run/openrc-run.c
index 13e5b71949c8..2e89c593d87e 100644
--- a/src/openrc-run/openrc-run.c
+++ b/src/openrc-run/openrc-run.c
@@ -105,11 +105,17 @@ static RC_STRINGLIST *deptypes_nwua;	/* need+want+use+after deps */
 static RC_STRINGLIST *deptypes_m;	/* needed deps for stopping */
 static RC_STRINGLIST *deptypes_mwua;	/* need+want+use+after deps for stopping */
 
+
+static void
+reap_all_children(void)
+{
+	while (waitpid(-1, NULL, WNOHANG) > 0);
+}
+
 static void
 handle_signal(int sig)
 {
-	int serrno = errno, status;
-	pid_t pid;
+	int serrno = errno;
 	const char *signame = NULL;
 	struct winsize ws;
 
@@ -123,11 +129,13 @@ handle_signal(int sig)
 		break;
 
 	case SIGCHLD:
-		while ((pid = waitpid(-1, &status, WNOHANG)) > 0) {
-			if (signal_pipe[1] > -1 && pid == service_pid) {
-				if (write(signal_pipe[1], &status, sizeof(status)) == -1)
-					eerror("%s: send: %s", applet, strerror(errno));
+		if (signal_pipe[1] > -1) {
+			if (write(signal_pipe[1], "", 1) < 0) {
+				eerror("%s: send: %s", applet, strerror(errno));
+				reap_all_children();
 			}
+		} else {
+			reap_all_children();
 		}
 		break;
 
@@ -351,12 +359,11 @@ openrc_sh_exec(const char *openrc_sh, const char *arg1, const char *arg2)
 static int
 svc_exec(const char *arg1, const char *arg2)
 {
-	int ret, fdout = fileno(stdout);
+	int ret, fdout = fileno(stdout), tmp;
 	struct termios tt;
 	struct winsize ws;
 	int flags = 0;
 	struct pollfd fd[2];
-	int s;
 	char buffer[BUFSIZ];
 	size_t bytes;
 	bool prefixed = false;
@@ -402,6 +409,7 @@ svc_exec(const char *arg1, const char *arg2)
 	now = tm_now();
 	wait_timeout = now + (WAIT_TIMEOUT * 1000);
 	warn_timeout = now + (WARN_TIMEOUT * 1000);
+	/* in theory we'd need a memory barrier between pipe2() and this fork...? */
 	service_pid = fork();
 	if (service_pid == -1)
 		eerrorx("%s: fork: %s", applet, strerror(errno));
@@ -449,18 +457,30 @@ svc_exec(const char *arg1, const char *arg2)
 			write_prefix(buffer, bytes, &prefixed);
 		}
 
-		/* signal_pipe receives service_pid's exit status */
+		/* signal_pipe is written to whenever SIGCHLD is received */
 		if (fd[0].revents & (POLLIN | POLLHUP)) {
-			if ((s = read(signal_pipe[0], &ret, sizeof(ret))) != sizeof(ret)) {
-				eerror("%s: receive failed: %s", applet, s < 0 ? strerror(errno) : "short read");
+			pid_t pid;
+			int status;
+			bool found = false;
+
+			if (read(signal_pipe[0], &ret, 1) < 0) {
+				eerror("%s: receive failed: %s", applet, strerror(errno));
 				ret = -1;
 				break;
 			}
-			ret = WEXITSTATUS(ret);
-			if (ret != 0 && errno == ECHILD)
-				/* killall5 -9 could cause this */
-				ret = 0;
-			break;
+
+			while ((pid = waitpid(-1, &status, WNOHANG)) > 0) {
+				if (pid == service_pid) {
+					found = true;
+					ret = WEXITSTATUS(status);
+					if (ret != 0 && errno == ECHILD)
+						/* killall5 -9 could cause this ?? */
+						ret = 0;
+				}
+			}
+			if (found) {
+				break;
+			}
 		}
 
 		if (forever)
@@ -483,8 +503,14 @@ svc_exec(const char *arg1, const char *arg2)
 	sigprocmask (SIG_BLOCK, &sigchldmask, &oldmask);
 
 	close(signal_pipe[0]);
-	close(signal_pipe[1]);
+	/* We should set signal_pipe[1] to -1 before closing it to avoid invalid write if
+	* sighandler runs between these two lines...
+	* Compiler can reorder this without a barrier :/ */
+	tmp = signal_pipe[1];
 	signal_pipe[0] = signal_pipe[1] = -1;
+	close(tmp);
+	/* try to reap children in case SIGCHLD was received between last read and now */
+	reap_all_children();
 
 	sigprocmask (SIG_SETMASK, &oldmask, NULL);
  ```

What do you think?
For now I'll test this patch over the weekend as well, if that patch works I'll open a PR with this or something similar depending on comments here.

(This is different from #589 in that we're not hanging inside the signal handler, so the problem is not that the signal handler is doing something it shouldn't it's that it's doing it too soon)

### PR DESCRIPTION
**Pull request by martinetd**
The signal handler in openrc-run expects service_pid to be set when the child
process exists, but there is no guarantee that the execution thread will have
been scheduled before the signal handler, leading to service not being detected
as started properly very occasionally.

This can be tested with the following reproducer:
```
#include <signal.h>
#include <stdio.h>
#include <string.h>
#include <sys/wait.h>
#include <unistd.h>

static pid_t service_pid;

static void
handle_signal(int sig)
{
	int status;
	pid_t pid;

	switch (sig) {
	case SIGCHLD:
		while ((pid = waitpid(-1, &status, WNOHANG)) > 0) {
			if (pid != service_pid)
				printf("Got %d != %d\n", pid, service_pid);
		}
		break;
	default:
		printf("caught unknown signal %d\n", sig);
	}
}

int
signal_setup(int sig, void (*handler)(int))
{
        struct sigaction sa;

        memset(&sa, 0, sizeof (sa));
        sigemptyset(&sa.sa_mask);
        sa.sa_handler = handler;
        return sigaction(sig, &sa, NULL);
}

int main(int argc, char **argv)
{
	signal_setup(SIGCHLD, handle_signal);

	while (1) {
#ifdef FIX
		sigset_t full, old;
		sigfillset(&full);
		sigprocmask(SIG_SETMASK, &full, &old);
#endif
		service_pid = fork();
#ifdef FIX
		sigprocmask(SIG_SETMASK, &old, NULL);
#endif

		if (service_pid < 0) {
			printf("fork fail %m\n");
			return 1;
		}
		// exit child immediately
		if (service_pid == 0)
			return 0;
		// wait for child process to stop a different way
		while (kill(service_pid, 0) >= 0);
	}

	return 0;
}
```

Fixes #895

### DIFF
--- a/src/openrc-run/openrc-run.c
+++ b/src/openrc-run/openrc-run.c
@@ -356,6 +356,7 @@ svc_exec(const char *arg1, const char *arg2)
 	struct winsize ws;
 	int flags = 0;
 	struct pollfd fd[2];
+	sigset_t full, old;
 	int s;
 	char buffer[BUFSIZ];
 	size_t bytes;
@@ -402,7 +403,11 @@ svc_exec(const char *arg1, const char *arg2)
 	now = tm_now();
 	wait_timeout = now + (WAIT_TIMEOUT * 1000);
 	warn_timeout = now + (WARN_TIMEOUT * 1000);
+	sigfillset(&full);
+	/* block SIGCHLD around the fork, so service_pid is set properly in handler */
+	sigprocmask(SIG_SETMASK, &full, &old);
 	service_pid = fork();
+	sigprocmask(SIG_SETMASK, &old, NULL);
 	if (service_pid == -1)
 		eerrorx("%s: fork: %s", applet, strerror(errno));
 	if (service_pid == 0) {
--- a/src/openrc-run/openrc-run.c
+++ b/src/openrc-run/openrc-run.c
@@ -351,7 +351,7 @@ openrc_sh_exec(const char *openrc_sh, const char *arg1, const char *arg2)
 static int
 svc_exec(const char *arg1, const char *arg2)
 {
-	int ret, fdout = fileno(stdout);
+	int ret, fdout = fileno(stdout), tmp;
 	struct termios tt;
 	struct winsize ws;
 	int flags = 0;
@@ -488,8 +488,11 @@ svc_exec(const char *arg1, const char *arg2)
 	sigprocmask (SIG_BLOCK, &sigchldmask, &oldmask);
 
 	close(signal_pipe[0]);
-	close(signal_pipe[1]);
+	/* We should set signal_pipe[1] to -1 before closing it to avoid invalid write if
+	 * sighandler runs between close() and cleaning the variable */
+	tmp = signal_pipe[1];
 	signal_pipe[0] = signal_pipe[1] = -1;
+	close(tmp);
 
 	sigprocmask (SIG_SETMASK, &oldmask, NULL);
 


navi-desu:
thanks

PR #899, opened 2025-07-07
librc-depend: clear dirfds unconditionally

navi-desu:
init.sh creates the directory, which means we never actually do it here early at boot

Fixes: 9c3f974
Fixes: https://github.com/OpenRC/openrc/pull/893


thesamesam:
9c3f974bc3a4a5848cb2184b42a3eab45282bb7c is an orphaned commit

martinetd:
Tested in the container repro I made yesterday and seems to work, thanks!

I'll also check with our real system later today just to be sure, if I don't report back that worked too.

EDIT: now confirmed, looks good to me!

PR #904, opened 2025-07-22
recognise alacritty as a color terminal

dakkar:
nothing fancy, just teaching `libeinfo` that `TERM=alacritty` means we can show colours


src/libeinfo/libeinfo.c
```diff
@@ -97,6 +97,7 @@ static bool term_is_cons25 = false;
  * Only terminals without "color" in the name need to be explicitly listed */
 static const char *const color_terms[] = {
 	"Eterm",
+        "alacritty",
```
navi-desu, line 100: use tabs instead of spaces please

dakkar: oops, sorry!

PR #908, opened 2025-08-18
init.d/net-online: Minor bug fix and minor feature addition

john-r-graham:
The code that reads interface status from the sysfs had stderr redirection in the wrong order, which means that, in the event that the interface does not yet exist, the read will display an error message that was meant to be suppressed.

Also added a single line that allows tun/tap interfaces to be properly handled. It turns out that /sys/class/net/${interface}/operstate shows "unknown" for tun/tap interfaces, not "up", which is the value that net-online depends on. So tun/tap interfaces need minor special handling.

### DIFF
--- a/init.d/net-online.in
+++ b/init.d/net-online.in
@@ -50,6 +50,7 @@ start ()
 		[ "$carrier" = 1 ] && : $((carriers += 1))
 		read operstate 2> /dev/null < /sys/class/net/$dev/operstate ||
 			operstate=
+		[ -e /sys/class/net/$dev/tun_flags ] && operstate=up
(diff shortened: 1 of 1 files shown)


navi-desu:
could you split this in two commits, one for the fix, and one for the tun/tap handling?

and name both commits accordingly

john-r-graham:
Certainly. But still *one* pull request?

navi-desu:
one pull request, yes

john-r-graham:
Done as requested.

john-r-graham:
Pushed again to correct a typo in a commit message. No code change.

navi-desu:
thanks

PR #911, opened 2025-08-30
silence out pam_openrc on login, and add rc_autostart_users to rc.conf

navi-desu:
we should also mark the service as stopped instead of failed -- at least while we're still in this experimental step

### DIFF
--- a/etc/rc.conf
+++ b/etc/rc.conf
@@ -15,6 +15,11 @@
 # set to YES.
 #rc_interactive="YES"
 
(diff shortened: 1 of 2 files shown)


etc/rc.conf
```diff
@@ -15,6 +15,11 @@
 # set to YES.
 #rc_interactive="YES"
 
+# Set to "NO" if you don't want pam_openrc autostarting user services. This
+# effetively disables the pam module, without the need of removing it from
```
thesamesam, line 19: effectively

PR #912, opened 2025-09-02
sysfs: mount tracefs explicitly if available

lotheac:
for rationale, see
https://web.git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=9ba817fb7c6afd3c86a6d4c3b822924b87ef0348


navi-desu:
thanks!

PR #918, opened 2025-09-07
sh: restore the installation of init.d/functions.sh

navi-desu:


navi-desu:
wait this is already being done, i just messed up the backporting to 0.62

bleh

PR #921, opened 2025-09-11
simplify some allocations

navi-desu:


navi-desu:
why is there 7 commits now

what

PR #933, opened 2025-09-27
Added missing options to the zsh completion definitions

nilninull:
Added missing options to the completion definition files for the rc-service, rc-status, and rc-update commands.
(Including --user and -U):

Added status to the cmd option for rc-service.


navi-desu:
thanks

Issue #937, opened 2025-10-17
nit: `rc-update`'s table format break if service's name is longer than 20 characters


lima-limon-inc:
Currently, `rc-update` displays the following output on my machine:
```
                  atd |      default
               binfmt | boot
             bootmisc | boot
              cgroups |                                 sysinit
              chronyd |      default
               cronie |      default
                 dbus |      default
                devfs |                                 sysinit
               dhcpcd |      default
                dmesg |                                 sysinit
              elogind | boot
                 fsck | boot
             hostname | boot
              hwclock | boot
              keymaps | boot
            killprocs |                        shutdown
    kmod-static-nodes |                                 sysinit
       libvirt-guests |      default
                local |      default nonetwork
           localmount | boot
             loopback | boot
              modules | boot
             mount-ro |                        shutdown
                 mtab | boot
             netmount |      default
             nftables |      default
               procfs | boot
                 root | boot
         save-keymaps | boot
    save-termencoding | boot
            savecache |                        shutdown
              seedrng | boot
                 swap | boot
               sysctl | boot
                sysfs |                                 sysinit
             sysklogd |      default
 systemd-tmpfiles-setup | boot
 systemd-tmpfiles-setup-dev |                                 sysinit
         termencoding | boot
                 udev |                                 sysinit
         udev-trigger |                                 sysinit
``` 

`systemd-tmpfiles-setup` and `systemd-tmpfiles-setup-dev` are misaligned with the rest of the services.

PR #938, opened 2025-10-17
Check for minimum padding required in rc-update

lima-limon-inc:
Before printing the service's names one by one, check the minimum padding that is going to be required for the longest service name.

Bug: https://github.com/OpenRC/openrc/issues/937

### DIFF
--- a/src/rc-update/rc-update.c
+++ b/src/rc-update/rc-update.c
@@ -191,7 +191,9 @@ show(RC_STRINGLIST *runlevels, bool verbose)
 		}
 
 		if (inone || verbose) {
-			printf(" %20s |", service->value);
+			/* For more information on why padding is 26, see issue 937: */
+			/* https://github.com/OpenRC/openrc/issues/937 */
+			printf(" %26s |", service->value);
 			TAILQ_FOREACH(runlevel, in, entries)
 			    printf (" %s", runlevel->value);
 			printf("\n");


lima-limon-inc:
With this patch, the misaligned output from `rc-update` described in the issue now looks like the following:
```
                        atd |      default
                     binfmt | boot
                   bootmisc | boot
                    cgroups |                                 sysinit
                    chronyd |      default
                     cronie |      default
                       dbus |      default
                      devfs |                                 sysinit
                     dhcpcd |      default
                      dmesg |                                 sysinit
                    elogind | boot
                       fsck | boot
                   hostname | boot
                    hwclock | boot
                    keymaps | boot
                  killprocs |                        shutdown
          kmod-static-nodes |                                 sysinit
             libvirt-guests |      default
                      local |      default nonetwork
                 localmount | boot
                   loopback | boot
                    modules | boot
                   mount-ro |                        shutdown
                       mtab | boot
                   netmount |      default
                   nftables |      default
                     procfs | boot
                       root | boot
               save-keymaps | boot
          save-termencoding | boot
                  savecache |                        shutdown
                    seedrng | boot
                       swap | boot
                     sysctl | boot
                      sysfs |                                 sysinit
                   sysklogd |      default
     systemd-tmpfiles-setup | boot
 systemd-tmpfiles-setup-dev |                                 sysinit
               termencoding | boot
                       udev |                                 sysinit
               udev-trigger |                                 sysinit
```

lima-limon-inc:
> I don't really like how the same loop is duplicated in two places now. And it's not like we can handle all edge cases anyways (e.g unicode names). 

I agree that duplicating the loop isn't ideal.

> I'd rather just increase the hardcoded padding or leave things be.

Noted, I'll push a commit with an increment in padding then. Thanks for the feedback :blush:

lima-limon-inc:
Changed here: https://github.com/OpenRC/openrc/pull/938/commits/869bb1d57307e595e5328ce3f758e90bd725e9fe

This is what the output looks like now:
```
                        atd |      default
                     binfmt | boot
                   bootmisc | boot
                    cgroups |                                 sysinit
                    chronyd |      default
                     cronie |      default
                       dbus |      default
                      devfs |                                 sysinit
                     dhcpcd |      default
                      dmesg |                                 sysinit
                    elogind | boot
                       fsck | boot
                   hostname | boot
                    hwclock | boot
                    keymaps | boot
                  killprocs |                        shutdown
          kmod-static-nodes |                                 sysinit
             libvirt-guests |      default
                      local |      default nonetwork
                 localmount | boot
                   loopback | boot
                    modules | boot
                   mount-ro |                        shutdown
                       mtab | boot
                   netmount |      default
                   nftables |      default
                     procfs | boot
                       root | boot
               save-keymaps | boot
          save-termencoding | boot
                  savecache |                        shutdown
                    seedrng | boot
                       swap | boot
                     sysctl | boot
                      sysfs |                                 sysinit
                   sysklogd |      default
     systemd-tmpfiles-setup | boot
 systemd-tmpfiles-setup-dev |                                 sysinit
               termencoding | boot
                       udev |                                 sysinit
               udev-trigger |                                 sysinit
```

navi-desu:
> I'd rather just increase the hardcoded padding or leave things be.

leaving things be is fine, though we already do have the calculation for openrc-run, moving it to our common static lib and calling it from both places would be an improvement, albeit small

doesn't need to be done now though, so, lgtm

lima-limon-inc:
Changed here: https://github.com/OpenRC/openrc/pull/938/commits/869bb1d57307e595e5328ce3f758e90bd725e9fe

This is what the output looks like now:


> > I'd rather just increase the hardcoded padding or leave things be.
> 
> leaving things be is fine, though we already do have the calculation for openrc-run, moving it to our common static lib and calling it from both places would be an improvement, albeit small
> 
> doesn't need to be done now though, so, lgtm

Wow! That was fast 

Thanks for the explanation and the context. Have a nice weekend :blush:

PR #942, opened 2025-11-03
introduce RC_PATH and dynamic path resolution

navi-desu:
a few questions still undecided:

- [x] should the system init path (so, not `--user`) include `XDG_CONFIG_DIRS`? so far i said "yes" because it simplifies logic¹
- [ ] should we go with /rc or /openrc? i prefer the former, but "rc" is more likely to cause name conflicts, so maybe it's time to start phashing it out (and i'm to blame for using it in ~/.config/rc), i do like "rc" over "openrc", though
- [x] do we overlay conf.d, or replace them? currently the code is replacing them. by overlay, i mean sourcing all found conf.d, in reverse order, so they're "merged" together, versus replacing, where only the latest found is used
- [x] do we go with `RC_PATH` or `RC_INIT_PATH`?
- [ ] systemd also adds `XDG_DATA_HOME` and `XDG_DATA_DIRS` to the system path, should we do so as well? unsure if init.d counts as "data" or "config", or if that distinction even matters here 

one goal with this patch set, is to eventually deprecate RC_USER_SERVICES, as nothing should really special-case it over doing proper "permission" checks

taking RC_PATH and RC_SVCDIR from the environment makes openrc's core logic more flexible, while removing special casing, and in doing so we redefine `rc_set_user()` as simply "use the default RC_PATH and RC_SVCDIR for user services"

this will allow us to later on remove all instances of `if (rc_yesno(getenv("RC_USER_SERVICES")) rc_set_user();`

we can't go that far yet due to runlevels and sysconfdir, but those i'm working on

i've yet to do a full non-superficial boot test with this branch, will do so over the next week

---

also the code is using strchrnul when parsing for convenience, it is not posix, but it's implemented by the bsds and musl. we could likely replace it without much difficulty, however.

---

1: systemd has casing for SYSTEMD_UNIT_PATH, where if it ends on an empty entry, aka, `:`, it'll prepend the listed unit path to the default path, instead of replacing it

by having XDG_CONFIG_DIRS be prepended by default, we avoid having to do such logic, and it makes the parsing code itself a lot more streamlined

should we not want to add XDG_CONFIG_DIRS to the system init directory path, we can instead incorporate the same special semantic as systemd

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -205,6 +205,8 @@ if get_option('pkgconfig')
     libraries : rc,
     variables: {
       'pluginsdir': pluginsdir,
+      'initdir': rc_libexecdir / 'init.d',
+      'confdir': rc_libexecdir / 'conf.d',
     }
   )
 endif


JustSoup312:
Tested on postmarketOS edge (2025-11-05) and verified working. A few services failed but I am pretty sure that is just because I hacked together the Alpine patches on top of this update and it didn't like that, but service loading worked, so that's good.

My thoughts:

> should the system init path (so, not --user) include XDG_CONFIG_DIRS

I say probably yes. If it simplifies logic, then no harm.

> should we go with /rc or /openrc

/openrc makes the most sense to me. Directory names should be descriptive of what they are used by, and as you said /rc can be taken by other service managers, which would be a conflict for no reason.

> do we overlay conf.d, or replace them

I would say overlay so a user can modify single settings of a service without duplicating the service config that might be put there by system vendors.

> do we go with RC_PATH or RC_INIT_PATH

If it is for both init.d and conf.d, then `RC_PATH`. If it is only for init.d, then `RC_INIT_PATH`.

> systemd also adds XDG_DATA_HOME and XDG_DATA_DIRS to the system path, should we do so as well

I don't see why `XDG_DATA_HOME` would be added since that is user-only stuff, but `XDG_DATA_DIRS` makes sense I guess.

> unsure if init.d counts as "data" or "config", or if that distinction even matters here

In this case, data is anything that can be architecture-specific. Configuration is anything that can not be architecture-specific. Thus, services are usually data and configs *can be* data.

navi-desu:
thanks for testing it

> /openrc makes the most sense to me. Directory names should be descriptive of what they are used by, and as you said /rc can be taken by other service managers, which would be a conflict for no reason.

we'd have to keep .config/rc as compat, then later move $libexec/rc -> $libexec/openrc, and though don't think other service managers would take '/rc', but a problem is, e.g. the 'rc' shell

/openrc looks kinda ugly to me compared to /rc, but i guess it'll have to do :(

> I don't see why `XDG_DATA_HOME` would be added since that is user-only stuff, but `XDG_DATA_DIRS` makes sense I guess.

user-wide installs that go, e.g., `meson setup --datadir $XDG_DATA_HOME` or similar, iirc that's why systemd does it
 
> > unsure if init.d counts as "data" or "config", or if that distinction even matters here
>
> In this case, data is anything that can be architecture-specific. Configuration is anything that can not be architecture-specific. Thus, services are usually data and configs _can be_ data.

i'm not sure that matches? e.g. manpages count as "data", and they're not arch-specific

i usually see data as "read-only content to be processed/displayed", and config as "read-write files that change behaviour", and executable scripts like our init scripts are weird to me bc they don't quite feel like either, bleh

JustSoup312:
> we'd have to keep .config/rc as compat, then later move $libexec/rc -> $libexec/openrc, and though don't think other service managers would take '/rc', but a problem is, e.g. the 'rc' shell

Ohhh that makes sense. Yeah, the rc shell could get mixed up.

> user-wide installs that go, e.g., meson setup --datadir $XDG_DATA_HOME or similar, iirc that's why systemd does it

I guess that makes sense, but it still feels weird. Guess no harm in including it if there is precedent.

> i'm not sure that matches? e.g. manpages count as "data", and they're not arch-specific
> i usually see data as "read-only content to be processed/displayed", and config as "read-write files that change behaviour", and
> executable scripts like our init scripts are weird to me bc they don't quite feel like either, bleh

In the context of the Linux filesystem, stuff in `/usr/share/` should be arch independent, which is where the man directory is *usually* located. `/usr/lib/` is for arch-dependent files (or executable files not in $PATH for usage by e.g. a service, but `/usr/libexec/` should really be used for that).. I should have clarified that by data I meant the stuff in `/usr/lib/` and by config I meant the stuff in `/usr/share/`.

navi-desu:
another thing to consider about '/rc' vs '/openrc', is that we also already use rc.h and librc.so

JustSoup312:
On /rc vs /openrc. I don't think it matters currently since it would involve moving other named pieces over to openrc instead of just rc if decided upon. Perhaps it makes sense to get this in first and *then* decide upon a default name without introducing scope-creep accidentally.

navi-desu:
that's... a good point, actually (i need to stop doing this thing where i stall doing things over decisions like that)

/rc vs /openrc aside, we can merge this today, i'll add the DATA dirs to the path, and we merge

JustSoup312:
No worries! I used to be the same way but pmOS taught me that if you try to get everything right instantly, it'll never get merged ;)

JustSoup312:
@navi-desu are you still working on this? If you are stuck or don't have time, I'd be fine helping fix up anything you think still needs work.

navi-desu:
oh, i am -- i've just been very busy, sorry

i have a new version almost ready, will send tomorrow after some testing

JustSoup312:
No worries! Just checking up.

navi-desu:
a) i'm not sure if we want XDG_DATA_DIRS tbh, so i've left it out for now, but the logic makes it easy to add it again eventually
b) i'm not super happy with the logic for conf.d sourcing, but if i don't come up with anything until 0.64 is released, it'll stay like this
c) ditto for "/rc" in paths and such

Issue #939, opened 2025-10-24
Incorrect warning message logic in respawn delay check


anosora233:
In the following code snippet:

https://github.com/OpenRC/openrc/blob/8d34e3f9525f0c6bb87669aec281fed0f698f09a/src/supervise-daemon/supervise-daemon.c#L1224-L1227

The warning message appears to be misleading. The condition checks if `respawn_delay` is greater than `respawn_delay_cap`, but the message suggests that the user should increase --respawn-delay to be more than --respawn-delay-cap, which is already the case.

Suggested Fix:

```c
if (respawn_delay_step > 0 && respawn_delay > respawn_delay_cap)
    ewarn("%s: Please decrease the value of --respawn-delay (%"PRId64"ms) to less "
        "than --respawn-delay-cap (%"PRId64"ms)", applet,
        respawn_delay, respawn_delay_cap);
```

PR #943, opened 2025-11-07
supervise-daemon: reverse meaning of respawn-delay value warning

JustSoup312:
Currently, the warning states that the user should increase the value of respawn-delay, even though the code checks if respawn-delay is more than the cap and period value. Thus, the correct warning is to recommend the opposite.

Fixes https://github.com/OpenRC/openrc/issues/939


navi-desu:
thanks!

Issue #949, opened 2025-11-14
Openrc/s6 integration issue


damobrisbane:
# ISSUE

I use Gentoo init/openrc and have been running docker containers some time under runlevel services but recently the functionality seems to have broken. Basic structure is:

```
# grep supervisor /etc/init.d/container.nexus
supervisor=s6
# 
# docker ps
CONTAINER ID   IMAGE                                      COMMAND                  CREATED        STATUS        PORTS     NAMES
cf611125e466   reg.ops:5000/strimzi/kafka-bridge:latest   "/opt/strimzi/bin/ka…"   8 hours ago    Up 8 hours              strimzi-bridge
dcf8e9f0bada   reg.ops:5000/traefik:xlatest               "/entrypoint.sh trae…"   8 hours ago    Up 8 hours              traefik
e7262831015d   reg.ops:5000/sonatype/nexus3               "/opt/sonatype/nexus…"   18 hours ago   Up 18 hours             nexus
# 
# /etc/init.d/container.nexus stop
container.nexus               | * Stopping container.nexus ...                                                                                                                                                                      [ ok ]
# 
# docker ps
CONTAINER ID   IMAGE                                      COMMAND                  CREATED       STATUS       PORTS     NAMES
cf611125e466   reg.ops:5000/strimzi/kafka-bridge:latest   "/opt/strimzi/bin/ka…"   8 hours ago   Up 8 hours             strimzi-bridge
dcf8e9f0bada   reg.ops:5000/traefik:xlatest               "/entrypoint.sh trae…"   8 hours ago   Up 8 hours             traefik
# 
# /etc/init.d/container.nexus start                                                                                                                                                                                                 [ ok ]
# 
# docker ps
CONTAINER ID   IMAGE                                      COMMAND                  CREATED         STATUS         PORTS     NAMES
7306942c0b47   reg.ops:5000/sonatype/nexus3               "/opt/sonatype/nexus…"   4 seconds ago   Up 3 seconds             nexus
cf611125e466   reg.ops:5000/strimzi/kafka-bridge:latest   "/opt/strimzi/bin/ka…"   8 hours ago     Up 8 hours               strimzi-bridge
dcf8e9f0bada   reg.ops:5000/traefik:xlatest               "/entrypoint.sh trae…"   8 hours ago     Up 8 hours               traefik
# 
# 
```

But as I said it broke. My rough analysis posted on gentoo forum was:

```
[...]

PR #950, opened 2025-11-18
s6 support: add compatibility with old mode

skarnet:
This change adds compatibility for people who use the old s6 support mechanism of OpenRC: it should fix the issue described in #949.

A service file with an empty `command` will use a user-provided service directory with the same name as the service file in `/var/svc.d`, if one exists, rather than attempting to create a service directory from the information given in the service file.

This PR also fixes a bug with the --notify fd:X option in `start-stop-daemon`, and also ensures that s6-svscan is run with that option, now that the bug doesn't prevent it anymore.


navi-desu:
thanks!

PR #952, opened 2025-11-19
*-daemon: always call setgroups, defaulting to count = 0

navi-desu:
services started by init don't get any supplementary groups, but anything started from a shell would inherit the groups, causing inconsistent behaviour

we can either clear all groups, or always initalize root's groups. since other init systems does not initialize anything, including us at boot, let's just always clear them unconditionally

### DIFF
--- a/src/start-stop-daemon/start-stop-daemon.c
+++ b/src/start-stop-daemon/start-stop-daemon.c
@@ -892,7 +892,7 @@ int main(int argc, char **argv)
 	/* Child process - lets go! */
 	if (pid == 0) {
 		gid_t group_buf[32], *group_list = group_buf;
-		int group_count = ARRAY_SIZE(group_buf);
+		int group_count = 0;
 		pid_t mypid = getpid();
 		close(pipefd[0]); /* Close the read end of the pipe. */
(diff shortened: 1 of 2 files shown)


navi-desu:
actually, on init, we get `Groups: 0`, while with this, we get `Groups:` (no groups)

hmmm, unsure what we *should* do here

navi-desu:
hmm, this breaks user services actually... i don't really want to check rc_is_user(), our goal is to deprecate that function

N-R-K:
Maybe we do something like `getgrouplist( changeuser ? changeuser : service_user_will_be_run_under, ...);` rather than defaulting to no groups?

navi-desu:
> Maybe we do something like `getgrouplist( changeuser ? changeuser : service_user_will_be_run_under, ...);` rather than defaulting to no groups?

i thought of always adding the groups over clearing them, though iirc the person who reported this explicitly didn't want extra groups

can non-root users set groups if those groups anyway? i think the right call might be to `getuid() == 0`

kind of feels like the simplest way to solve all those inconsistencies, would be to always run services from a dedicated runner daemon, eh

PR #961, opened 2026-01-05
 Remove restart(); add framework for reload() support

skarnet:
OpenRC will never call restart(), so scrap the definition in
openrc-run.sh and s6.sh.
 reload() caused a weird error message when used without
supervisor=s6, so add proper support for generic reload(), instantiated for s6, and with a better default error message for other backends - which can now add support for reload() down the line.

 Also clarify a comment, and add info messages on s6_reload.

### DIFF
--- a/sh/openrc-run.sh.in
+++ b/sh/openrc-run.sh.in
@@ -203,6 +203,17 @@ default_status()
 	$func
 }
 
+default_reload()
+{
+	case "$supervisor" in
+		s6) s6_reload ;;
+		?*)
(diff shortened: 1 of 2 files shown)


navi-desu:
i think i'd rather print an error like `reload not defined` (seems like we currently print `$command: not found`, i'll improve that latter), saying "only supported with supervisor=s6" sounds like it can't be used with anything else -- though i can fix it myself while merging if you'd like

skarnet:
That's why I wrote "currently", I think? But yeah, please merge and change the ewarn message to whatever feels better to you, the message is just a placeholder.

PR #962, opened 2026-01-05
man/openrc-shutdown.8: correct kexec option

jesdh:
7689106a changed the kexec option for openrc-shutdown from "-k" to "-K" but the manpage was never updated.


N-R-K (approved):

navi-desu:
thanks!

PR #964, opened 2026-01-17
rc-service: print name of command on exec failure

lanodan:
For example with noexec in the home, it changes:

    $ rc-service -U foo start
     * rc-service: Permission denied

to:

    $ rc-service -U foo start
     * rc-service: Failed executing '/home/haelwenn/.config/rc/init.d/foo': Permission denied


navi-desu:
thanks!

PR #966, opened 2026-01-29
Added experimental feature multithread start/stop services deps tree

GermanAizek:
For test feature use flag parameter `rc_paralel` to enable and disable multithreading.

nkJl9g


navi-desu:
openrc is inherently a multiprocess system, we currently spawn processes off and wait serially -- waiting in parallel won't help as all the work is being done in other processes, not to mention the now added overhead of allocating extra unnecessary thread data, and spawning and joining N threads.

i am pretty sure we do not want this, was it benchmarked at all?

PR #974, opened 2026-02-10
librc-depend: fix unnecessary copying parent memory page table

GermanAizek:
posix_spawn() is a more lightweight and modern alternative to fork()/exec(), which are used inside popen(). The main advantage of posix_spawn is that it can create a new process without completely duplicating the address space of the parent process.

References:
- https://blog.famzah.net/2018/12/19/posix_spawn-performance-benchmarks-and-usage-examples/
- https://lobste.rs/s/smbsd5/fork_road
- https://www.reddit.com/r/C_Programming/comments/1lvdhp2/fork_vs_posix_spawn/

### DIFF
--- a/src/librc/librc-depend.c
+++ b/src/librc/librc-depend.c
@@ -27,6 +27,7 @@
 #include <sys/stat.h>
 #include <time.h>
 #include <unistd.h>
+#include <spawn.h>
 
 #include "queue.h"
 #include "librc.h"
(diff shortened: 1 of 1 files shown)


navi-desu:
this is just... #975 but without the helper? as mentioned over there, popen does not fork() internally on any platform we support

GermanAizek:
> this is just... #975 but without the helper? as mentioned over there, popen does not fork() internally on any platform we support


OpenRC, which prides itself on being fast and lightweight, performs worse than systemd on older single-core systems. And the longest startup process is "Caching service dependencies". This is where the longest slowdown occurs due to outdated Popen() functions. If you are not ready to optimize even for older systems, then what is the point of supporting OpenRC if it is easier to contribute to systemd.

https://www.youtube.com/watch?v=GErsSlI_SXk

https://www.youtube.com/watch?v=NxOL4Jg-zZ0

navi-desu:
could you please not split discussion between two (basically identical) PRs?

PR #975, opened 2026-02-10
librc: continue migrate popen() to posix_spawn()

GermanAizek:
Added execute_get_output() new function has been created that reads commands and replaces two popen() calls.

posix_spawn() is a more lightweight and modern alternative to fork()/exec(), which are used inside popen(). The main advantage of posix_spawn is that it can create a new process without completely duplicating the address space of the parent process.

References:
- https://blog.famzah.net/2018/12/19/posix_spawn-performance-benchmarks-and-usage-examples/
- https://lobste.rs/s/smbsd5/fork_road
- https://www.reddit.com/r/C_Programming/comments/1lvdhp2/fork_vs_posix_spawn/

### DIFF
--- a/src/librc/librc.c
+++ b/src/librc/librc.c
@@ -31,6 +31,8 @@
 #include <sys/param.h>
 #include <sys/stat.h>
 #include <unistd.h>
+#include <spawn.h>
+#include <sys/wait.h>
 
 #include "queue.h"
 #include "librc.h"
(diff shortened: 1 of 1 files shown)


navi-desu:
> posix_spawn() is a more lightweight and modern alternative to fork()/exec(), which are used inside popen().

for any modern libc this is just... not true?

https://github.com/bminor/glibc/blob/master/libio/iopopen.c#L131
https://git.musl-libc.org/cgit/musl/tree/src/stdio/popen.c

glibc and musl uses posix_spawn internally

https://cgit.freebsd.org/src/tree/lib/libc/gen/popen.c#n116
https://github.com/openbsd/src/blob/master/lib/libc/gen/popen.c#L79

freebsd, openbsd, and others, uses vfork which has the same benefits as posix_spawn over traditional fork()

---

i don't think we gain anything by reimplementing popen when libcs already do it efficiently

navi-desu:
sorry but there's no point in changing this for modern systems, it only increases complexity

GermanAizek:
> sorry but there's no point in changing this for modern systems, it only increases complexity

OpenRC, which prides itself on being fast and lightweight, performs worse than systemd on older single-core systems. And the longest startup process is "Caching service dependencies". This is where the longest slowdown occurs due to outdated Popen() functions. If you are not ready to optimize even for older systems, then what is the point of supporting OpenRC if it is easier to contribute to systemd.

https://www.youtube.com/watch?v=GErsSlI_SXk

https://www.youtube.com/watch?v=NxOL4Jg-zZ0

navi-desu:
> OpenRC, which prides itself on being fast and lightweight, performs worse than systemd on older single-core systems. And the longest startup process is "Caching service dependencies".  

the slowdown occours because we spawn a shell and parse all init scripts, to generate a dependency tree, and because every service `stat`s all other services to check if the dependency tree is up to date, every time

this is a huge problem, which is being worked on -- being smart about dependency caching, and 0.61 (iirc) restored the behaviour of caching the dependency tree in /var/cache, a version of which, both videos you sent are not using, they're both using years old versions

> This is where the longest slowdown occurs due to outdated Popen() functions.

did you even read my first reply? glibc and musl *already* use posix_spawn, so please elaborate, what does this PR do over their implementation of popen?

> If you are not ready to optimize even for older systems, then what is the point of supporting OpenRC if it is easier to contribute to systemd.

i am ready to optimize, if you only explained how your optimizations help, and show some actual numbers

PR #986, opened 2026-03-12
openrc-init: read sysvinit commands from /run/initctl

floppym:
This PR implements support for reading sysvinit commands from /run/initctl. This allows users to boot their system using openrc-init, but still allow them to control it using the traditional halt/poweroff/reboot commands from sysvinit without replacing them with symlinks (sys-apps/openrc[sysv-utils]).

Also included is a sizable refactor of code out of the main() function. If you would prefer to review that separately I can split it out.


navi-desu:
oh neat a rebase makes the whole thing come over again huh

PR #1001, opened 2026-04-09
consolefont: allow passing extra arguments to setfont

Feasuro:
**Summary**

This change introduces support for an optional `customflags` variable in `/etc/conf.d/consolefont`, allowing users to pass additional arguments to the `setfont` command.

**Motivation**

Currently, the `consolefont` OpenRC service does not provide a way to customize the arguments passed to `setfont`. Users who need non-default behavior (for example, using the `-d` flag for double-size fonts) must modify `/etc/init.d/consolefont` directly.

Since this file is owned by the package, such changes are overwritten on upgrade, forcing users to rely on fragile workarounds such as patching via package manager hooks.

**Proposed solution**

Introduce a new optional variable:

`customflags=""`

This variable is sourced from `/etc/conf.d/consolefont` and appended to the `setfont` invocation:

`setfont $customflags $consolefont $param -C $ttydev$x >/dev/null`

**Benefits**
- Enables customization without modifying init scripts
- Preserves existing default behavior when unset
- Aligns with existing OpenRC patterns of exposing command arguments via `/etc/conf.d/*`
- Eliminates the need for downstream patching or hooks

**Backward compatibility**

No changes in behavior for existing users. If `customflags` is unset or empty, the command behaves exactly as before.

**Example usage**

To enable double-size console fonts:

`consolefont="Lat2-Terminus16"`
`customflags="-d"`

### DIFF
--- a/conf.d/consolefont
+++ b/conf.d/consolefont
@@ -16,3 +16,7 @@ consolefont="default8x16"
 # default one. Have a look in /usr/share/unimaps for a selection of map files
 # you can use.
 #unicodemap="iso01"
+
+# customflags is used to pass custom arguments to the setfont command;
+# e.g. "-d" for doubling the font size on high pixel density displays.
+#customflags=""
--- a/init.d/consolefont.in
+++ b/init.d/consolefont.in
@@ -24,6 +24,7 @@ start()
 	consolefont=${consolefont:-${CONSOLEFONT}}
 	unicodemap=${unicodemap:-${UNICODEMAP}}
 	consoletranslation=${consoletranslation:-${CONSOLETRANSLATION}}
+	customflags=${customflags:-${CUSTOMFLAGS}}
 
 	if [ -z "$consolefont" ]; then
 		ebegin "Using the default console font"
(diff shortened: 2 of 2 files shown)


rapidcow:
By merit of shell splitting I have been able to pass the `-d` flag by simply specifying the consolefont with it:

```
$ grep ^consolefont= /etc/conf.d/consolefont
consolefont="-d Tamsyn7x13r"

==> /var/log/rc.log <==
rc boot
 * Setting console font [-d Tamsyn7x13r] ...
 [ ok ]
```

That already lets me pass arbitrary flags in any way I see fit; but that's just my two cents. ;)

That aside, a few thoughts (despite not being a maintainer):

* `customflags` is way too generic (and if not, it as a variable [optionally read from the environment](https://github.com/OpenRC/openrc/blob/cb9ed066733faa2589ac8536605c090a6d60b83a/init.d/consolefont.in#L27) is!).  Better scope it with a prefix like `consolefont_args` (I like args because they seem far more common in conf.d than opt[ion]s or flags.)
* A generic pass-through variable may not be the most consistent with existing variables that deal with one argument at a time (`consoletranslation => -m/--consolemap`, `unicodemap => -u/--unicodemap`). I would consider making it a boolean like `consolefont_double="YES"` unless there is another flag other than `-d/--double` you need for **setfont**(8). (Personally, [I don't really see any...](https://man7.org/linux/man-pages/man8/setfont.8.html#OPTIONS))

PR #1007, opened 2026-05-13
Never run plugins on exit()

funderscore1:
The commits should be self-explanatory, but I'll recap anyway:

OpenRC has been running plugins in `cleanup()` on premature exit, to make sure that all plugin hooks are ran. `cleanup()` on openrc/openrc-run is an `atexit` function, as in, it will run on `exit()`. Running plugins means a `fork()`, then letting the plugins do their job, and finally `exit()`. At this point this is the second `exit()` that has been ran.

The problem is that running `exit()` more than once is undefined behavior. While glibc did not complain about it for more than 20 years (this bug was around since 2008 in rc: 4ae5aebe81c57fa3e03a48d84d1f091b6f92a027), musl gets angry at it. This will make the child process (the plugin) helpfully hang as soon as it's running `exit()`, to indicate that this should not happen. A side effect of this is OpenRC(-run) itself also hanging, as it is waiting for the plugin to exit.

However the hang only happens if a service fails. If a service succeeds, then no amount of running plugin hooks will be done in `cleanup()` (as this kind of thing is only done on failure).

This can make a system unbootable or not able to shutdown anymore if a service fails on boot/shutdown.

I've been told that this plugin system will be eventually replaced by a more robust messaging system, but for now, we can stop doing that, and instead just manually call OUT hooks *before* exit(). This is a slight change in behavior but it shouldn't really matter given the small number of OpenRC plugins.

This issue was found in Alpine edge, initially running `plymouth-openrc-plugin`, but *any* plugin will cause this bug to happen.

### DIFF
--- a/src/openrc-run/openrc-run.c
+++ b/src/openrc-run/openrc-run.c
@@ -93,7 +93,6 @@ static RC_STRINGLIST *restart_services;
 static RC_STRINGLIST *need_services;
 static RC_STRINGLIST *use_services;
 static RC_STRINGLIST *want_services;
-static RC_HOOK hook_out;
 static int exclusive_fd = -1, master_tty = -1;
 static bool in_background, deps, dry_run;
 static volatile bool sighup, skip_mark, timedout;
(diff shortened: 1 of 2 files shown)


funderscore1:
BTW, this is the downstream issue: https://gitlab.postmarketos.org/postmarketOS/pmaports/-/work_items/4536

navi-desu:
@funderscore1 i wasn't happy with the code duplication, and it didn't seem to trigger _DONE

could you test it out to see if this new version i pushed is still good?

once it's good i'll backport to 0.63.x and tag 0.63.2

PR #1010, opened 2026-05-23
fix: tools/meson_final.sh lowercase linux

roman-kiselevich:
~~DESTDIR doesn't make sense to me because rc_libexecdir is already absolute path because it depends on the get_option('prefix')~~

Just a linux lowercase fix

### DIFF
--- a/tools/meson_final.sh
+++ b/tools/meson_final.sh
@@ -6,7 +6,7 @@ set -u
 rc_libexecdir="$1"
 os="$2"
 
-if [ "${os}" != Linux ]; then
+if [ "${os}" != linux ]; then
 	install -d "${DESTDIR}/${rc_libexecdir}"/init.d
 fi
 install -m 644 "${MESON_BUILD_ROOT}/src/shared/version" "${DESTDIR}/${rc_libexecdir}"
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@ core
 .gdbinit
 
 build
+
+# IDE
+.idea


tools/meson_final.sh
```diff
@@ -6,7 +6,7 @@ set -u
 rc_libexecdir="$1"
 os="$2"
 
-if [ "${os}" != Linux ]; then
-	install -d "${DESTDIR}/${rc_libexecdir}"/init.d
+if [ "${os}" != linux ]; then
+	install -d "${rc_libexecdir}"/init.d
 fi
-install -m 644 "${MESON_BUILD_ROOT}/src/shared/version" "${DESTDIR}/${rc_libexecdir}"
+install -m 644 "${MESON_BUILD_ROOT}/src/shared/version" "${rc_libexecdir}"
```
navi-desu, line 12: changing Linux to linux is correct, that was a mistake from when we changed from a custom meson option, to meson's native os name functions

removing destdir does not seem correct, however, on irc you wrote:

> imo DESTDIR has to be gone there because rc_libexecdir is already absolute path

the "absolute" path is supposed to be installed relative to DESTDIR, as package managers when building will set DESTDIR to a temporary dir, and then package the contents of said DESTDIR to be later extracted onto /

removing DESTDIR would make the `install` command act upon the current build /, which is not what we want

navi-desu: already fixed, thanks

roman-kiselevich: thx for your reply. I removed DESTDIR editions

navi-desu:
thx

Issue #1019, opened 2026-06-06
feat: supervise-daemon: support piping both stderr and stdout to the same process


WhyNotHugo:
When using `supervise-daemon`, I'd like to pipe both stderr and stdout to the same process.

Specifying the same value for `output_logger` and `error_logger` will spawn two separate logger processes, at which point I need to log to two destinations. I want to log all of a service's output to a single destination.

WhyNotHugo:
Maybe a `--merge-stderr` flag for `supervise-daemon`, along with a `merge_stderr` for `openrc-run`?

PR #1020, opened 2026-06-06
supervise-daemon: allow merging stderr and stdout

WhyNotHugo:
Implement a --merge-stderr flag for supervise-daemon, along with a merge_stderr for openrc-run. This merges stderr into stdout, allowing a single process (via --output_logger/output_logger) to handle all output from a service.

Implements: https://github.com/OpenRC/openrc/issues/1019

### DIFF
--- a/man/openrc-run.8
+++ b/man/openrc-run.8
@@ -191,6 +191,14 @@ your service output to go to syslog.
 The same thing as
 .Pa output_logger
 but for the standard error output.
+.It Ar merge_stderr
+Set to any value to redirect stderr to stdout.
+Use this to combine all output to a single
+.Pa output_logger .
+Cannot be used with
+.Pa error_log
+or
+.Pa error_logger .
 .It Ar directory
 .Xr start-stop-daemon 8
 and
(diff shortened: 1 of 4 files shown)


navi-desu:
i think i'd more in favor of redirecting stderr to stdout if only `output_log` is set (and maybe don't if `output_log="foo" error_log=`)

after releasing 0.64 i've promised to revamp logging in openrc (so we no longer /dev/null them by default either) so i'll take a look at this then

WhyNotHugo:
> after releasing 0.64 i've promised to revamp logging in openrc (so we no longer /dev/null them by default either) so i'll take a look at this then

I'd love to be able to configure logging globally. E.g.: a default for all services which don't override it. An expected usage would be something like:

    output_logger="logger -t $SERVICE_NAME -p daemon.info"

WhyNotHugo:
> i think i'd more in favor of redirecting stderr to stdout if only output_log is set (and maybe don't if output_log="foo" error_log=)

That sounds very sensible; I can't imagine that "I want to log stdout and discard stderr" is a common use case.

PR #1026, opened 2026-06-14
RC_PATH tweaks

navi-desu:
for 2/4:
by all intents i could gather, XDG_CONFIG_DIRS aren't really supposed to be searched for system packages, no one does, and using different suffixes on it vs _HOME is technically/pedantically spec uncompliant -- i'm still a bit unsure about how RC_PATH should interact with user services, since setting an RC_PATH combined with --user could cause issues

maybe --user should also append /user to RC_PATH? have to check what systemd does there

for 4/4:
mostly done for debian, i'm not super happy with it but if there's no alternative is something we could do for them


werdahias:
Thanks, will test more in the upcoming days

PR #1028, opened 2026-06-22
openrc-run.8: Fix "retry" line in start/stop table

ulm:


navi-desu:
supervise-daemon.sh read $retry on supervise_start, so i don't quite think this is correct?

the value is used when stopping, but it's loaded to the supervisor daemon on startup (so, changing it between start and stop won't have any effect)

ulm:
> the value is used when stopping, but it's loaded to the supervisor daemon on startup (so, changing it between start and stop won't have any effect)

The description of the table says "The following table lists all the variables that are **used** by the default start and stop functions."

If it is more complicated than that, then it might profit from a more complete explanation?

navi-desu:
> The description of the table says "The following table lists all the variables that are used by the default start and stop functions."

it's meant to describe when the init script reads from the variable, this one specifically is read at start and stored until stop

i'm not sure what a better wording would be

ulm:
Feel free to close this PR then.

PR #1032, opened 2026-07-01
Add default reload function

funderscore1:
Historically, reload was meant to be implemented by the service script. However in the end it seems most service scripts just send a signal by calling `start-stop-daemon` directly, something that can easily be done as a default and would reduce duplication.

Add default reload functions for both supervise-daemon and start-stop-daemon, which just send a signal to a service. The signal to send is configured by just setting:

```sh
reloadsig=USR1  # signal to send to trigger reload
```

For service scripts that need to do other things on reload, they can define their own reload function just like before.

### DIFF
--- a/service-script-guide.md
+++ b/service-script-guide.md
@@ -103,6 +103,7 @@ All service scripts are assumed to have the following functions:
 ```
 start()
 stop()
+reload()
 status()
 ```
 
@@ -117,6 +118,7 @@ script:
 command=
 command_args=
 pidfile=
+reloadsig=
 ```
 
 Thus the 'smallest' service scripts can be half a dozen lines long
@@ -131,6 +133,7 @@ following OpenRC variables are likely all that you'll need:
   * command
   * command_args
   * pidfile
+  * reloadsig
 
 Given those three pieces of information, OpenRC will be able to start
 and stop the daemon on its own. The following is taken from an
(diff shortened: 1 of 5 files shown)


service-script-guide.md
```diff
@@ -242,7 +245,17 @@ To recap, in order of preference:
...
 Many daemons will reload their configuration files in response to a
 signal. Suppose your daemon will reload its configuration in response
-to a `SIGHUP`. It's possible to add a new "reload" command to your
+to a `SIGHUP`. You can simply add the following to your service script:
+
+```sh
+reloadsig=HUP
+```
+
+For most daemons this is sufficient, though note it only works on OpenRC TODO.
```
funderscore1, line 254: This needs to be adjusted before merging

sh/supervise-daemon.sh
```diff
@@ -82,6 +82,22 @@ supervise_stop()
...
+supervise_reload()
+{
+	if [ -n $reloadsig ]; then
+		eerror "There is nothing for ${name:-$RC_SVCNAME} to reload."
+		return 0
+	fi
+
+	local startchroot="$(service_get_value "chroot")"
+	local startpidfile="$(service_get_value "pidfile")"
+	chroot="${startchroot:-$chroot}"
```
navi-desu: supervise-daemon pidfiles are neither stored in chroot, nor used for signal, sendsig uses an internal fifopath

PR #1033, opened 2026-07-01
Further document stacked runlevels 

roselandgoose:
* Documents the order they are started in.
* Documents the limitations of `before`/`after` ordering.
* Clarifies what `rc-update -s <add|delete>` acts on.

I expect `man/rc-update.8: generalize beyond just services` to be the most controversial, since wording is hard.

### DIFF
--- a/man/rc-update.8
+++ b/man/rc-update.8
@@ -77,7 +77,7 @@ This allows inheritance of runlevels.
 .Pp
 If the
 .Fl a, -all
-option is given, we remove the service from all runlevels. This is
+option is given, we remove the service or runlevel from all runlevels. This is
 useful, for example, to clean up the dangling symlinks after a service
 is removed.
 .Sh SEE ALSO
--- a/man/rc-update.8
+++ b/man/rc-update.8
@@ -13,18 +13,18 @@
 .Os OpenRC
 .Sh NAME
 .Nm rc-update
-.Nd add and remove services to and from a runlevel
+.Nd add and remove services or runlevels, to and from runlevels
 .Sh SYNOPSIS
 .Nm
 .Op Fl s , -stack
 .Ic add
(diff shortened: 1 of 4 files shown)


man/rc-update.8
```diff
@@ -13,18 +13,18 @@
 .Os OpenRC
 .Sh NAME
 .Nm rc-update
-.Nd add and remove services to and from a runlevel
+.Nd add and remove dependencies to and from a runlevel
```
navi-desu: i think the word "dependency" is less clear for the base case, maybe `add and remove services or runlevels, to and from the target runlevels`

man/rc-update.8
```diff
@@ -72,12 +72,12 @@ system clock).
...
 If the
 .Fl s , -stack
-option is given then we either add or remove the runlevel from the runlevel.
+option is given then the dependency we add or remove must be another runlevel.
 This allows inheritance of runlevels.
 .Pp
 If the
 .Fl a, -all
-option is given, we remove the service from all runlevels. This is
+option is given, we remove the service/runlevel from all runlevels. This is
```
navi-desu: here also `service or runlevel` would be a bit better i think

navi-desu:
if no complains, i can do the changes when merging

roselandgoose:
Re:
> i think the word "dependency" is less clear for the base case, maybe `add and remove services or runlevels, to and from the target runlevels`

+1

I drafted something along those lines first, but hesitated when the command summary could no longer fit on one line.

roselandgoose:
> if no complains, i can do the changes when merging

Sure :) phrase it however you see fit! Thanks

roselandgoose:
I've re-phrased it similar to your suggestions. Feel free to adjust further!

PR #1041, opened 2026-07-20
Set default runlevel for --user to 'default'

### LINKED ISSUE #1035: Default runlevel for user services
The default runlevel for user services is currently `sysinit`, but navi mentioned a few times on IRC that this is neither intentional nor desired: it's just a side-effect of how OpenRC is implemented for user services.

Opening this issue to discuss tackling this: should a different runlevel be the default for user service? What name should it have?

I can maybe help tackle this, but the intended behaviour needs to be defined.

WhyNotHugo:
The existing default runlevel of sysinit was never intentional, and a side-effect of how system mode works.

Closes: https://github.com/OpenRC/openrc/issues/1035

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,19 @@ OpenRC NEWS
 This file will contain a list of notable changes for each release. Note
 the information in this file is in reverse order.
 
+## OpenRC 0.64
+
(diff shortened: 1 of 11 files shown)


navi-desu:
made it default to "default" even as system user, since #1061 i really think it should be the same behaviour regardless of --user

PR #1045, opened 2026-07-23
user-guide: add a note about logging out for the user scripts to be recognized

zyxhere:


navi-desu:
the logout isn't necessary though, starting a runlevel maybe be (though not anymore)

logging out and back in would just trigger pam_openrc and create default directories, but if created by hand, they should be seen regardless

zyxhere:
On Thu, 2026-07-23 at 05:52 -0700, navi wrote:
> navi-desu left a comment (OpenRC/openrc#1045)
> the logout isn't necessary though, starting a runlevel maybe be
> (though not anymore)
> logging out and back in would just trigger pam_openrc and create
> default directories, but if created by hand, they should be seen
> regardless

For me only ~/.config/rc/runlevels was created. When I created
~/.config/rc/init.d and put service files there OpenRC doesn't
recognize them at all and errors with `/etc/user/init.d/foo` doesn't
exist. Experimenting led me to logout and back in which led to OpenRC
seeing the service files in ~/.config/rc

PR #1054, opened 2026-07-28
librc: fix service startup without pty & crash regressions

pva:
This MR fixes two regressions:

- Continue service startup when openpty() fails, restoring the behavior from before 39eb8a9d.
- Avoid a crash during failure cleanup introduced by 68b05704.


navi-desu:
thx

PR #1055, opened 2026-07-29
openrc-init: keep a dummy writer on the control FIFO

pva:
The init FIFO is opened read-only and monitored with poll(). After a client writes a command and closes the FIFO, Linux continues reporting POLLHUP even after the buffered data has been consumed. Since the main loop only handles POLLIN, poll() then returns immediately on every iteration, causing PID 1 to consume all available CPU time.

This can be reproduced on a system using openrc-init as PID 1 with a harmless unknown command:

    printf invalid > /run/openrc/init.ctl
    top -p 1

Keep a dummy write descriptor open for the lifetime of openrc-init so that disconnecting a client does not leave the read end hung up. Mark the descriptor close-on-exec so it is replaced normally during reexec. A single O_RDWR descriptor would achieve the same, but POSIX leaves O_RDWR on a FIFO undefined, so use a separate write descriptor.

Regression after: afe3eca0 ("openrc-init: use poll for waiting on the fifo")

### DIFF
--- a/src/openrc-init/openrc-init.c
+++ b/src/openrc-init/openrc-init.c
@@ -285,9 +285,30 @@ static void reap_zombies(int sig RC_UNUSED)
 
 static int open_fifo(const char *path)
 {
+	int dummy_writer;
+	int reader;
+	int saved_errno;
+
 	if (mkfifo(path, 0600) == -1 && errno != EEXIST)
 		return -1;
-	return open(path, O_RDONLY | O_NONBLOCK | O_CLOEXEC);
+
+	reader = open(path, O_RDONLY | O_NONBLOCK | O_CLOEXEC);
+	if (reader == -1)
+		return -1;
+
+	/*
+	 * Keep the dummy writer open for the lifetime of the process so that the
+	 * read end does not continuously report POLLHUP when a client disconnects.
+	 */
+	dummy_writer = open(path, O_WRONLY | O_NONBLOCK | O_CLOEXEC);
+	if (dummy_writer == -1) {
+		saved_errno = errno;
+		close(reader);
+		errno = saved_errno;
+		return -1;
+	}
+
+	return reader;
 }
 
 int main(int argc, char **argv)


pva:
Note: this can be fixed with a much shorter patch, by just changing `O_RDONLY` to `O_RDWR`:
```C
static int open_fifo(const char *path)
{
	if (mkfifo(path, 0600) == -1 && errno != EEXIST)
		return -1;
	return open(path, O_RDWR | O_NONBLOCK | O_CLOEXEC);
}
```
POSIX leaves opening a FIFO with O_RDWR undefined, so this relies on platform-specific behaviour. On Linux it is documented in fifo(7), and sysvinit has kept /run/initctl open this way for decades. Since openrc-init is Linux-only anyway, maybe we should go with the shorter change. What do you think?

navi-desu:
> Note: this can be fixed with a much shorter patch, by just changing `O_RDONLY` to `O_RDWR`:
> 
> ```c
> static int open_fifo(const char *path)
> {
> 	if (mkfifo(path, 0600) == -1 && errno != EEXIST)
> 		return -1;
> 	return open(path, O_RDWR | O_NONBLOCK | O_CLOEXEC);
> }
> ```
> 
> POSIX leaves opening a FIFO with O_RDWR undefined, so this relies on platform-specific behaviour. On Linux it is documented in fifo(7), and sysvinit has kept /run/initctl open this way for decades. Since openrc-init is Linux-only anyway, maybe we should go with the shorter change. What do you think?

openrc-init is also used on gentoo's hurd port (though with patches that are not upstream yet)

if the same about O_RDWR is true for hurd, then we can go ahead with the simpler patch

src/openrc-init/openrc-init.c
```diff
@@ -283,17 +283,37 @@ static void reap_zombies(int sig RC_UNUSED)
 	errno = saved_errno;
 }
 
-static int open_fifo(const char *path)
+static int open_fifo(const char *path, int *dummy_writer)
```
navi-desu: on this version of the patch, fwiw, we're not closing it at all but relying on O_CLOEXEC, so we don't really even have to store the fd, could just let it leak

pva:
> openrc-init is also used on gentoo's hurd port (though with patches that are not upstream yet)

For GNU/Hurd, I found this [open issue concerning open(FIFO, O_RDWR)](https://www.gnu.org/software/hurd/open_issues/fifo_O_RDWR.html). Looking at the current [FIFO translator implementation](https://cgit.git.savannah.gnu.org/cgit/hurd/hurd.git/tree/trans/fifo.c), the issue does not appear to have been fixed. Since openrc-init is expected to run on GNU/Hurd, we should keep the longer version of the patch.

src/openrc-init/openrc-init.c
pva: Makes sense. I've updated patch to reflect this.

PR #1056, opened 2026-07-31
sysv-initctl: avoid busy loop without FIFO clients

pva:
sysv-initctl opened /run/initctl with O_RDONLY | O_NONBLOCK and called read() directly in a loop. Immediately after startup, no client has the FIFO open for writing, so read() returns zero instead of waiting. The daemon treated this as a short request and retried continuously, consuming CPU and flooding syslog with "read: short count".

This can be reproduced on a system using openrc-init by starting the compatibility daemon without connecting an initctl client:

    rc-service sysv-initctl start
    pid=$(pgrep -x sysv-initctl)
    strace -p "$pid" -e read
    top -p "$pid"

Before this change, strace repeatedly reports read() returning zero.

Open the read end nonblocking during setup, keep a dummy write end open, and then make the reader blocking. This allows the daemon to sleep in read() while no clients are connected. Mark both descriptors close-on-exec.

### DIFF
--- a/src/sysv-initctl/initctl.c
+++ b/src/sysv-initctl/initctl.c
@@ -14,6 +14,43 @@
 
 static bool init_halt = false;
 
+static int open_fifo(const char *path)
+{
+	int dummy_writer;
+	int flags;
+	int reader;
+	int saved_errno;
+
+	reader = open(path, O_RDONLY | O_NONBLOCK | O_CLOEXEC);
+	if (reader == -1)
+		return -1;
+
+	/*
+	 * Keep the dummy writer open for the lifetime of the process so that the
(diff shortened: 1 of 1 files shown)


navi-desu:
this is much of the same logic as openrc-init, so i pushed a commit reusing it

also there's no need to repeat the explanation every time, just refer to the old fix

pva:
@navi-desu Yea, it's a good idea to move code into a helper header. Yet, although the fixes are similar, the triggers are different. In the first case, the issue was triggered by connecting to the socket and sending a signal; here, it is triggered when no listener is present. That's why the descriptions were different.

PR #1078, opened 2026-09-17
openrc-user: add a bunch of allowed env

stacyharper:


stacyharper:
feels free to NAK the ones you don't want

navi-desu:
we probably should allow all LC_* vars, i'm unsure about LESS and PAGER (specially for system services), and SDL_* feels *very* specific... hmm

stacyharper:
Agreed, adapted a bit

navi-desu:
looks good, thanks

